PR #109, opened 2020-01-03
iochaos: add selector for injecting sidecar when using init-request config

cwen0:

In this pr, we add a selector in the config of the sidecar, because we need it when we inject sidecar at the starting of the app cluster.  For example, we want to inject sidecar to tikv pod at the starting of tidb cluster, we can apply the config of tikv sidecar  [ tikv-configmap.yaml](https://github.com/pingcap/chaos-mesh/blob/master/examples/chaosfs-configmap/tikv-configmap.yaml) and the selector need to add to the config file.

eg: 
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: chaosfs-tikv
  labels:
    app.kubernetes.io/component: webhook
data:
  chaosfs-tikv.yaml: |
    name: chaosfs-tikv
    selector: 
      labelSelectors:
    initContainers:
    - name: inject-scripts
      image: pingcap/chaos-scripts:latest
      imagePullpolicy: Always
      command: ["sh", "-c", "mkdir -p /tmp/scripts; cp -R /scripts/* /tmp/scripts/; mkdir -p /var/lib/tikv/fuse-data"]
    containers:
   .....
``` 

Then add `admission-webhook.pingcap.com/init-request=chaosfs-tikv` to the annotation of the tidb cluster namespace and start the tidb cluster. The tikv pod created next will be injected into the sidecar container.


pkg/utils/selector.go
```diff
@@ -127,6 +127,64 @@ func SelectPods(ctx context.Context, c client.Client, selector v1alpha1.Selector
...
+// TODO: support to check fieldsSelector
+func CheckPodMeetSelector(pod v1.Pod, selector v1alpha1.SelectorSpec) (bool, error) {
+	if len(selector.Pods) > 0 {
+		for ns, names := range selector.Pods {
+			if pod.Namespace != ns {
+				continue
+			}
+
+			for _, name := range names {
+				if pod.Name == name {
```
zhouqiang-cl: Did we need to mind regex such as *

pkg/webhook/config/config.go
```diff
@@ -24,6 +24,7 @@ import (
 	"sync"
 
 	"github.com/ghodss/yaml"
+	"github.com/pingcap/chaos-mesh/api/v1alpha1"
```
zhouqiang-cl: Did we can change to v1

pkg/utils/selector.go
cwen0: I think `regex` is not necessary at present.

pkg/webhook/config/config.go
cwen0: Too early 😂

pkg/utils/selector.go
```diff
@@ -127,6 +127,64 @@ func SelectPods(ctx context.Context, c client.Client, selector v1alpha1.Selector
...
+					return true, nil
```
mahjonp: why return true once pod name matches? maybe other selectors(eg.`AnnotationSelectors`) don't match

pkg/webhook/inject/inject.go
```diff
@@ -84,6 +84,24 @@ func Inject(res *v1beta1.AdmissionRequest, cli client.Client, cfg *config.Config
 		}
 	}
 
+	if injectionConfig.Selector != nil {
+		meet, err := utils.CheckPodMeetSelector(pod, *injectionConfig.Selector)
+		if err != nil {
+			log.Error(err, "Failed to check pod selector", "namespace", pod.Namespace)
+			return &v1beta1.AdmissionResponse{
+				Allowed: true,
```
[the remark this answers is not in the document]
cwen0: this response will ignore injection and just to create pod

pkg/utils/selector.go
cwen0: yep, you are right. I will fix it.

cwen0: oh, if the pod met this selector, it means that the pod is specified.

cwen0: This may be ambiguous, let me fix it

mahjonp: we'd better let the selection continues?

cwen0: yep, you are right, I will fix it.

cwen0:
@mahjonp PTAL again, thanks!

pkg/utils/selector.go
```diff
@@ -131,17 +131,24 @@ func SelectPods(ctx context.Context, c client.Client, selector v1alpha1.Selector
...
 			for _, name := range names {
 				if pod.Name == name {
-					return true, nil
+					meet = true
+					break Loop
 				}
 			}
 		}
+
+		if !meet {
```
mahjonp: Don't need `Loop` label. when leaving the inner for-block, if meet is false boolean, it has no chance to become true.

PR #117, opened 2020-01-09
selector: ignore the phase of the pod when to check if it meets the selector

cwen0:
In this pr, we add the namespace in ConfigMaps and ignore the phase of the pod when to check if it meets the selector

### DIFF
--- a/pkg/utils/selector.go
+++ b/pkg/utils/selector.go
@@ -110,7 +110,6 @@ func SelectPods(ctx context.Context, c client.Client, selector v1alpha1.Selector
 	if err != nil {
 		return nil, err
 	}
-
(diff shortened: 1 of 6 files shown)


pkg/utils/selector.go
```diff
@@ -181,8 +182,9 @@ func CheckPodMeetSelector(pod v1.Pod, selector v1alpha1.SelectorSpec) (bool, err
 	if err != nil {
 		return false, err
 	}
+
 	pods = filterByAnnotations(pods, annotationsSelector)
-	pods = filterByPhase(pods, v1.PodRunning)
+
```
mahjonp: how about adding a pod phase parameter? I don't think that removing the phase check is a good idea.

PR #140, opened 2020-01-16
cluster init script for kind 0.7.0

you06:

Current kind use `kind get kubeconfig-path --name=${clusterName}` to get kubeconfig path, this has been deprecated from v0.7.0 which is the current latest release of kind.

Others changes:

* Fix typos in example charts

* Fix a typo in demo gif and slow down the play speed

### Check List 

Tests 

 - No code


hack/kind-cluster-build.sh
```diff
@@ -132,7 +132,8 @@ done
 
 echo "start to create k8s cluster"
 kind create cluster --config ${configFile} --image kindest/node:${k8sVersion} --name=${clusterName}
-export KUBECONFIG="$(kind get kubeconfig-path --name=${clusterName})"
+kind get kubeconfig --name=${clusterName} > ${workDir}/config
```
mahjonp: use an env variable here?

you06: Seems it's not able to get the kubeconfig path directory from kind 0.7.0, so I have to dump it to a file and set env variable after that. Do you mean by use env variable directly, how can we do it?

mahjonp: sorry, I mean ${workDir}/config, it’s also used in another place.

you06: Resolved, @mahjonp PTAL again.

cwen0:
fix https://github.com/pingcap/chaos-mesh/issues/111

PR #182, opened 2020-02-03
add auto goimports into Makefile

YangKeao:

Add `goimports` command into Makefile.

Diff check has been added to CI to ensure every PR has been formatted and yaml has been regenerated.

Tests 

 - No code

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,8 +86,11 @@ manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
 # Run go fmt against code
-fmt:
+fmt: groupimports
 	$(GO) fmt ./...
+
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -86,9 +86,12 @@ manifests: controller-gen
 	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
 # Run go fmt against code
-fmt:
+fmt: groupimports
 	$(GO) fmt ./...
 
+groupimports:
```
zhouqiang-cl: Did we need to install goimport first

Yisaer:
We should also add this into the CI test.

Makefile
YangKeao: Yes! Though `goimports` is auto installed as `golang.org/x/tools` is an indirect dependency, it's still necessary to check and install.

PR #198, opened 2020-02-05
Support duration chaos for Network Chaos

Yisaer:
#171
Make `Network` chaos support duration chaos feature.

Making each kind of `Networkchaos` implement common controller interface. I have tested `netem` and `partiton` chaos locally in kind and works fine.

### Check List 

Tests 

Code changes

 - Has Go code change

Side effects

 - None

Related changes

```release-note
Support duration chaos for network chaos by omitting the scheduler and duration.
```


controllers/common/types.go
```diff
@@ -80,6 +80,7 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			r.Log.Error(err, "failed to recover chaos")
 			return ctrl.Result{Requeue: true}, nil
 		}
+		return ctrl.Result{}, nil
```
Yisaer: This is a bug during handling the recover logic in `common chaos` controller, I fixed in this request when I found it.

controllers/common/networkchaos/netem/types.go
```diff
@@ -0,0 +1,58 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package netem
+
+import (
+	"context"
+
+	"github.com/pingcap/chaos-mesh/controllers/networkchaos/netem"
```
zhouqiang-cl: group this import?

controllers/networkchaos/types.go
```diff
@@ -41,16 +48,54 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, nil
 	}
 
+	scheduler := networkchaos.GetScheduler()
+	duration, err := networkchaos.GetDuration()
+	if err != nil {
+		r.Log.Error(err, fmt.Sprintf("unable to get podchaos[%s/%s]'s duration", networkchaos.Namespace, networkchaos.Name))
+		return ctrl.Result{}, nil
+	}
+	if scheduler == nil && duration == nil {
```
zhouqiang-cl: what if duration != nil

controllers/common/networkchaos/netem/types.go
Yisaer: I execute `make groupimports` locally and there is no change. Is this something wrong?

controllers/networkchaos/types.go
Yisaer: @zhouqiang-cl `Networkchaos` controller only handle the situation when `scheduler` and `duration` are both defined or omitted. For the other situation, it would log the error and return the empty response.

More over, I also think Adding validation admission webhook for each chaos kind to check the `scheduler` and `duration` situation is necessary as I mentioned in #171

controllers/common/networkchaos/netem/types.go
zhouqiang-cl: It is wield 😯

controllers/networkchaos/types.go
zhouqiang-cl: Got

zhouqiang-cl:
@YangKeao @cwen0 PTAL

controllers/common/networkchaos/netem/types.go
at15: It seems to be default behavior https://github.com/golang/go/issues/27365#issuecomment-422720475 and https://github.com/golang/go/issues/26846#issuecomment-412629549 you need to put imports together so goimport will treat them as a group and sort within the group. Otherwise it assumes you put them into different group for some reason, e.g. third party, protobuf etc.

zhouqiang-cl:
@at15 Can you take a look at other part of this PR too

controllers/twophase/networkchaos/netem/types.go
```diff
@@ -0,0 +1,56 @@
...
+	r := &netem.Reconciler{
+		Client: c,
+		Log:    log,
+	}
+	return &TwoPhaseReconciler{
+		Reconciler: r,
+	}
+}
+
+// TwoPhaseReconciler reconcile the networkchaos
```
at15: It seems two phase pod has the [same comments](https://github.com/pingcap/chaos-mesh/blob/master/controllers/twophase/podfailure/types.go).  A more friendly comment would be `TwoPhaseReconciler implements twophase.InnerObject for netem.Reconciler so it can xxxx` It would be good to have some comment on the [InnerObject](https://github.com/pingcap/chaos-mesh/blob/master/controllers/twophase/types.go#L34) for two phase as well

controllers/networkchaos/types.go
```diff
@@ -15,15 +15,22 @@ package networkchaos
...
-	"github.com/pingcap/chaos-mesh/controllers/networkchaos/partition"
+	"github.com/pingcap/chaos-mesh/controllers/twophase"
+
+	"github.com/go-logr/logr"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/pingcap/chaos-mesh/api/v1alpha1"
+	commonNetem "github.com/pingcap/chaos-mesh/controllers/common/networkchaos/netem"
```
at15: I find the common and twwphase wrapping and renaming a bit duplicated. Is it possible to eliminate them? I am not very familiar with the code base ... it seems we have two set of interfaces under common and twophase. And the wrappers are almost same (except for the last parameter)

```go
// under common/types
// InnerCommonReconcile used in common chaos reconcile
type InnerCommonReconcile interface {
	Apply(ctx context.Context, req ctrl.Request, chaos InnerCommonObject) error
	Recover(ctx context.Context, req ctrl.Request, chaos InnerCommonObject) error
	Object() InnerCommonObject
}

// under twophase/types
type InnerReconciler interface {
	Apply(ctx context.Context, req ctrl.Request, chaos InnerObject) error
	Recover(ctx context.Context, req ctrl.Request, chaos InnerObject) error
	Object() InnerObject
}

// under common/podfailure
// CommonReconciler is reconciler for podfailure
type CommonReconciler struct {
	*podfailure.Reconciler
}

// Apply would perform common chaos for podchaos
func (r *CommonReconciler) Apply(ctx context.Context, req ctrl.Request, chaos common.InnerCommonObject) error {
	return r.Perform(ctx, req, chaos)
}

// under twophase/podfailure
// TwoPhaseReconciler reconcile the networkchaos
type TwoPhaseReconciler struct {
	*podfailure.Reconciler
}

// Apply implement InnerReconciler.Apply
func (r *TwoPhaseReconciler) Apply(ctx context.Context, req ctrl.Request, chaos twophase.InnerObject) error {
	return r.Perform(ctx, req, chaos)
}
```

Instead of having a layout where we duplicate package names under common and twophase

```
- podfailure
- netem
- common
   - podfailure
   - netem
- twophase
   - podfaillure
   - netem
```

We can define those wrappers in `podfailure` and `netem` directly with new factory func like `NewCommonReconciler`, to avoid having two extra packages for each chaos and use `netem.NewCommonReconciler` instead of `commonNetem.NewReconciler`. This removes subpackage from `common` and `twophase` packages and solves the renaming problem. 

```go
// podfailure

// NewReconciler would create new reconciler for podfailure chaos
func NewReconciler(c client.Client, log logr.Logger, req ctrl.Request) *Reconciler {
	return &Reconciler{Client: c, Log:  log}
}

func NewCommonReconciler(...) *CommonReconciler {
}

func NewTwoPhaseReconciler(...) *TwoPhaseReconciler {
}
```

btw: the wrappers should be generated

controllers/common/types.go
mahjonp: do we need `r.Update(ctx, chaos)` here?

controllers/networkchaos/types.go
```diff
@@ -41,16 +48,54 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		return r.defaultResponse(networkchaos), nil
+	}
+}
 
-		return ctrl.Result{}, nil
+func (r *Reconciler) scheduleNetworkChaos(networkchaos *v1alpha1.NetworkChaos, req ctrl.Request) (ctrl.Result, error) {
+	switch networkchaos.Spec.Action {
+	case v1alpha1.DelayAction, v1alpha1.DuplicateAction, v1alpha1.CorruptAction, v1alpha1.LossAction:
+		r := twophaseNetem.NewTwoPhaseReconciler(r.Client, r.Log.WithValues("action", "netem"), req)
+		reconciler := twophase.NewReconciler(r, r.Client, r.Log)
```
at15: nit: if you want to save two lines you can

```go
var inner twohpase.InnerReconciler
swtich {
case Delay
     // assume previous comment is applied and twophaseNetem package is gone
     inner = netem.NewTwoPhaseReconciler
case Partition:
     inner = partition.NewTwoPhaseReconciler
default:
     return r.ignoreInvalidAction
}
rc := twophase.NewReconciler(inner, ...)
return rc.Reconcile
```

controllers/networkchaos/types.go
mahjonp: Deeply agree with you.

controllers/twophase/networkchaos/netem/types.go
```diff
@@ -0,0 +1,56 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package netem
```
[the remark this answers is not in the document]
Yisaer: Nice advice, I think it should be recorded as an new issue.

zhouqiang-cl:
@at15 thank you very much❤️

Yisaer:
Thanks a lot for the code review.
Current layout is a bit of complex and I can't agree it anymore. As @at15 's advice, I would refactor the current layout to make the whole controller structure easier.

zhouqiang-cl:
@Yisaer please solve the conflict

controllers/networkchaos/types.go
Yisaer: updated.

controllers/networkchaos/types.go
Yisaer: fixed in #201

controllers/common/types.go
Yisaer: > do we need `r.Update(ctx, chaos)` here?

good catch. I adjust the common reconciler.

zhouqiang-cl:
@cwen0 @YangKeao PTAL

controllers/common/types.go
```diff
@@ -65,31 +65,30 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			r.Log.Error(err, "failed to recover chaos")
 			return ctrl.Result{Requeue: true}, nil
 		}
-		return ctrl.Result{}, nil
-	}
+	} else {
+		// Start failure action
+		r.Log.Info("Performing Action")
```
Yisaer: Adjust the common reconciler, make sure the chaos object would be udated after Performing or Recovering.

controllers/common/types.go
```diff
@@ -65,31 +65,30 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			r.Log.Error(err, "failed to recover chaos")
 			return ctrl.Result{Requeue: true}, nil
 		}
```
cwen0: `return return ctrl.Result{}, nil`

controllers/common/types.go
```diff
@@ -65,31 +65,30 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 
-	status := chaos.GetStatus()
+		err = r.Apply(ctx, req, chaos)
+		if err != nil {
+			r.Log.Error(err, "failed to apply chaos action")
 
-	err = r.Apply(ctx, req, chaos)
-	if err != nil {
-		r.Log.Error(err, "failed to apply chaos action")
+			updateError := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
cwen0: If `apply action` is failed, we should probably not update the chaos object and recover it. @YangKeao what do you think? Maybe we need a separate pr to fix this issue.

controllers/common/types.go
Yisaer: updated.

controllers/common/types.go
```diff
@@ -63,33 +63,32 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		err = r.Recover(ctx, req, chaos)
 		if err != nil {
 			r.Log.Error(err, "failed to recover chaos")
-			return ctrl.Result{Requeue: true}, nil
+			return ctrl.Result{}, nil
```
mahjonp: emm, so we won't retry if the recovery fails? I think it will cause more unrecovered chaos actions.

Yisaer: I agreed. What's your opinion? @cwen0

mahjonp: And could we set a field like `Status`, let user be aware of it when observer chaos result?

cwen0: yes, `requeue` is better and if `recover action` was failed, we also can block this delete action. @mahjonp There is already a status field in chaos object, but we have not used it well.

zhouqiang-cl: so, this pr's handle is ok?

Yisaer: updated.

PR #210, opened 2020-02-09
Support prometheus metrics in chaos-daemon

yeya24:

This PR adds some basic prometheus metrics to chaos-daemon:

1. process metrics
2. go metrics
3. grpc metrics

Since we need to expose `/metrics` http endpoint, so add a `http-port` to chaos-daemon.

Also, refactor a little bit of code in chaos-daemon to make it more readable. This seems make this PR much larger, but I cannot come up with a way to split it. So feel free to review and comment on this PR!

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
The previous `port` parameter in chaos-daemon is renamed to `grpc-port`. 
Add a new `http-port` in chaos-daemon to specify the port for Prometheus metrics.
```

### DIFF
--- a/cmd/chaos-daemon/main.go
+++ b/cmd/chaos-daemon/main.go
@@ -16,7 +16,8 @@ package main
 import (
 	"flag"
 	"os"
-	"strconv"
+
+	"github.com/prometheus/client_golang/prometheus"
 
 	"github.com/pingcap/chaos-mesh/pkg/chaosdaemon"
 	"github.com/pingcap/chaos-mesh/pkg/version"
@@ -25,18 +26,18 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 )
 
-var log = ctrl.Log.WithName("chaos-daemon")
-
 var (
-	printVersion     bool
-	rawPort          string
-	containerRuntime string
+	log  = ctrl.Log.WithName("chaos-daemon")
+	conf = &chaosdaemon.Config{Host: "0.0.0.0"}
+
+	printVersion bool
 )
 
 func init() {
 	flag.BoolVar(&printVersion, "version", false, "print version information and exit")
-	flag.StringVar(&rawPort, "port", "", "the port which server listens on")
-	flag.StringVar(&containerRuntime, "runtime", "docker", "current container runtime")
+	flag.IntVar(&conf.GRPCPort, "grpc-port", 8080, "the port which grpc server listens on")
+	flag.IntVar(&conf.HTTPPort, "http-port", 8081, "the port which http server listens on")
+	flag.StringVar(&conf.Runtime, "runtime", "docker", "current container runtime")
 
 	flag.Parse()
 }
@@ -50,21 +51,13 @@ func main() {
 
 	ctrl.SetLogger(zap.Logger(true))
 
-	if rawPort == "" {
-		rawPort = os.Getenv("PORT")
-	}
+	reg := prometheus.NewRegistry()
+	reg.MustRegister(
+		prometheus.NewGoCollector(),
+		prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{}),
+	)
 
-	if rawPort == "" {
-		rawPort = "8080"
-	}
-
-	port, err := strconv.Atoi(rawPort)
-	if err != nil {
-		log.Error(err, "Error while parsing PORT environment variable", "port", rawPort)
-		os.Exit(1)
-	}
-
-	if err := chaosdaemon.StartServer("0.0.0.0", port, containerRuntime); err != nil {
+	if err := chaosdaemon.StartServer(conf, reg); err != nil {
 		log.Error(err, "failed to start chaos-daemon server")
 		os.Exit(1)
 	}
--- a/cmd/chaos-daemon/main.go
+++ b/cmd/chaos-daemon/main.go
@@ -35,8 +35,8 @@ var (
 
 func init() {
 	flag.BoolVar(&printVersion, "version", false, "print version information and exit")
-	flag.IntVar(&conf.GRPCPort, "grpc-port", 8080, "the port which grpc server listens on")
-	flag.IntVar(&conf.HTTPPort, "http-port", 8081, "the port which http server listens on")
+	flag.IntVar(&conf.GRPCPort, "grpc-port", 31767, "the port which grpc server listens on")
+	flag.IntVar(&conf.HTTPPort, "http-port", 31766, "the port which http server listens on")
 	flag.StringVar(&conf.Runtime, "runtime", "docker", "current container runtime")
 
 	flag.Parse()
--- a/helm/chaos-mesh/templates/chaos-daemon-deployment.yaml
+++ b/helm/chaos-mesh/templates/chaos-daemon-deployment.yaml
@@ -33,6 +33,10 @@ spec:
             - /usr/local/bin/chaos-daemon
             - --runtime
             - {{ .Values.chaosDaemon.runtime }}
+            - --http-port
+            - !!str {{ .Values.chaosDaemon.httpPort }}
+            - --grpc-port
+            - !!str {{ .Values.chaosDaemon.grpcPort }}
           env:
             - name: PORT
               value: !!str {{ .Values.chaosDaemon.port }}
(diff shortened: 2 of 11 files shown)


yeya24:
Already tested in my cluster
image

cmd/chaos-daemon/main.go
```diff
@@ -25,18 +26,18 @@ import (
...
+	conf = &chaosdaemon.Config{Host: "0.0.0.0"}
+
+	printVersion bool
 )
 
 func init() {
 	flag.BoolVar(&printVersion, "version", false, "print version information and exit")
-	flag.StringVar(&rawPort, "port", "", "the port which server listens on")
-	flag.StringVar(&containerRuntime, "runtime", "docker", "current container runtime")
+	flag.IntVar(&conf.GRPCPort, "grpc-port", 8080, "the port which grpc server listens on")
```
zhouqiang-cl: this two port is too common, did we need to change @cwen0 @YangKeao

helm/chaos-mesh/templates/chaos-daemon-deployment.yaml
```diff
@@ -48,8 +52,11 @@ spec:
               mountPath: /run/containerd/containerd.sock
               {{- end }}
           ports:
-            - containerPort: {{ .Values.chaosDaemon.port }}
-              hostPort: {{ .Values.chaosDaemon.port }}
+            - name: grpc
+              containerPort: {{ .Values.chaosDaemon.grpcPort }}
+              hostPort: {{ .Values.chaosDaemon.grpcPort }}
+            - name: http
+              containerPort: {{ .Values.chaosDaemon.httpPort }}
```
zhouqiang-cl: Did we also need hostPort here

yeya24: Yes, Prometheus needs to scrape the http port for metrics

pkg/chaosdaemon/server.go
```diff
@@ -30,45 +35,96 @@ var log = ctrl.Log.WithName("chaos-daemon-server")
...
 	}
 
-	s := grpc.NewServer(grpc.UnaryInterceptor(utils.TimeoutServerInterceptor))
-	chaosDaemonServer, err := newServer(containerRuntime)
-	if err != nil {
-		log.Error(err, "failed to create server")
-		return err
+	grpcMetrics := grpc_prometheus.NewServerMetrics()
+	grpcMetrics.EnableHandlingTimeHistogram(
+		grpc_prometheus.WithHistogramBuckets([]float64{0.001, 0.01, 0.1, 0.3, 0.6, 1, 3, 6, 9, 20}),
```
yeya24: I will reduce the buckets. Seems like the latency cannot come to 20s, it won't be that slow

helm/chaos-mesh/templates/chaos-daemon-deployment.yaml
zhouqiang-cl: ok

pkg/chaosdaemon/server.go
```diff
@@ -30,45 +35,96 @@ var log = ctrl.Log.WithName("chaos-daemon-server")
...
+	)
+	reg.MustRegister(grpcMetrics)
+
+	grpcOpts := []grpc.ServerOption{
+		grpc_middleware.WithUnaryServerChain(
+			grpcMetrics.UnaryServerInterceptor(),
+			utils.TimeoutServerInterceptor,
```
mahjonp: The order seems matter, putting `TimeoutServerInterceptor` before `grpcMetrics.UnaryServerInterceptor` may be better, what's your opinion?

zhouqiang-cl:
After this pr，I think we can start add some metrics to grafana @cwen0 @YangKeao @lucklove

yeya24:
> After this pr，I think we can start add some metrics to grafana @cwen0 @YangKeao @lucklove

Actually, I think we need a Prometheus setup in helm charts for chaos-mesh to monitor the components. We can open an issue to track this.

zhouqiang-cl:
> > After this pr，I think we can start add some metrics to grafana @cwen0 @YangKeao @lucklove
> 
> Actually, I think we need a Prometheus setup first in helm charts for chaos-mesh to monitor the components. We can open an issue to track this.

ah, already open an issue https://github.com/pingcap/chaos-mesh/issues/213

pkg/chaosdaemon/server.go
yeya24: Yes, that's better. Updated.

pkg/chaosdaemon/server.go
yeya24: Updated

cmd/chaos-daemon/main.go
```diff
@@ -50,21 +51,13 @@ func main() {
 
 	ctrl.SetLogger(zap.Logger(true))
 
-	if rawPort == "" {
-		rawPort = os.Getenv("PORT")
```
cwen0: This environment variable is set in the helm script. If you want to remove this environment variable, please also update the helm script https://github.com/pingcap/chaos-mesh/blob/master/helm/chaos-mesh/templates/chaos-daemon-deployment.yaml#L41

cmd/chaos-daemon/main.go
cwen0: I agree to change the default port

helm/chaos-mesh/templates/chaos-daemon-deployment.yaml
```diff
@@ -37,6 +37,10 @@ spec:
             - /usr/local/bin/chaos-daemon
             - --runtime
             - {{ .Values.chaosDaemon.runtime }}
+            - --http-port
+            - !!str {{ .Values.chaosDaemon.httpPort }}
+            - --grpc-port
+            - !!str {{ .Values.chaosDaemon.grpcPort }}
           env:
             - name: PORT
```
cwen0: this `env` is useless, we can remove it

pkg/chaosdaemon/http.go
```diff
@@ -0,0 +1,24 @@
+package chaosdaemon
```
cwen0: add the license

cmd/chaos-daemon/main.go
yeya24: Yes, but change to what? Any recommendations? But since we can set a large port (3xxxx) in the helm chart so I think it is ok to not change it.

Yisaer:
@yeya24  Hi, could you point out the `Breaking backward compatibility` change in Release Note?

yeya24:
> @yeya24 Hi, could you point out the `Breaking backward compatibility` change in Release Note?

Sure, but I cannot find a changelog or release file. Could you please tell me what file should I change?

Yisaer:
@yeya24 Currently, point it In the request description, this would help us when we release new version.

**Does this PR introduce a user-facing change?:**

pkg/chaosdaemon/http.go
yeya24: Done

cmd/chaos-daemon/main.go
yeya24: Done

yeya24:
> @yeya24 Currently, point it In the request description, this would help us when we release new version.
> 
> **Does this PR introduce a user-facing change?:**

Done. Please check it

cmd/chaos-daemon/main.go
mahjonp: Maybe use the same value with values.yaml?

yeya24: That sounds good. I will set to that value.

zhouqiang-cl:
@Yisaer PTAL

cmd/chaos-daemon/main.go
yeya24: Actually we don't need to set http port, it is ok to be a common port since we don't expose it as a host port

helm/chaos-mesh/values.yaml
```diff
@@ -48,7 +48,8 @@ controllerManager:
 chaosDaemon:
   image: pingcap/chaos-daemon:latest
   imagePullPolicy: Always
-  port: 31767
+  grpcPort: 31767
+  httpPort: 31766
```
yeya24: Maybe we should remove the httpPort config. I think we don't need to expose this config and it is ok to hard code it in the yaml.

Issue #95, opened 2019-12-31
support time chaos


zhouqiang-cl:
RT

zhouqiang-cl:
The possible option is going to split into two stage

1. use bpf to support all kinds of time chaos
2. use vDSO to speed up C/Rust program(which support dynamic link) time chaos

PR #222, opened 2020-02-11
support timechaos

YangKeao:

Add TimeChaos: #95

These changes are applied to chaos-daemon

- Add hostPID and hostIPC attribute to pod
- Add SYS_PTRACE capacity to container
- Don't mount /proc to /mnt/proc
- Add time chaos request handler

These changes are applied to build procedure

- Build all go binary in docker image
- Auto pass HTTP_PROXY and HTTPS_PROXY environment variables

### Check List

Tests 

 - Unit test
 - Manual test

I use a simple go file to test:

```go
package main

import (
        "fmt"
        "time"
)

func main() {
        ticker := time.NewTicker(2 * time.Second)

        for {
                select {
                case <-ticker.C:
                        now := time.Now()
                        fmt.Println(now)
                }
        }
}
```

The tested yaml is:

```
kind: TimeChaos
metadata:
  name: time-shift
  namespace: chaos-testing
spec:
  mode: all
  selector:
    namespaces:
      - default
    labelSelectors:
      "app": "timer"
  timeOffset:
    sec: 100000
    nsec: 100000
  duration: "15s"
  scheduler:
    cron: "@every 20s"
```

Code changes

 - Has Go code change


Makefile
```diff
@@ -101,15 +104,18 @@ endif
...
-	$(GO) vet ./...
+	$(GOENV) CGO_ENABLED=1 go vet ./...
 
 tidy:
 	@echo "go mod tidy"
 	GO111MODULE=on go mod tidy
 	git diff --quiet go.mod go.sum
 
 image:
+	mkdir -p images/chaos-daemon/src
```
YangKeao: Should we use a standalone Dockerfile at root to build all binaries? This can also forbid the leak of developer's information (such as path in log :sweat_smile: )

Copy source file is quite ugly :(

pkg/fakeclockgettime/fake_clock_gettime.c
```diff
@@ -0,0 +1,25 @@
+#include <time.h>
```
YangKeao: This file was just used to show the logic of these byte code. (We cannot use `dlopen` to open a so image now, so we cannot compile these codes and just use them.)

zhouqiang-cl:
> * Build binary in docker image (as we use cgo, the build environment should be same with production environment)

Did we need target test binary also using cgo

pkg/time/time.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/pingcap/chaos-mesh/pkg/ptrace"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+var log = ctrl.Log.WithName("time")
+
+// TODO: auto generate these codes
+var fakeImage = []byte{
+	0xb8, 0xe4, 0x00, 0x00, 0x00, //mov    $0xe4,%eax
```
YangKeao: These codes are "linked" manually. Maybe we can write a simple linker later (which seems little complicate)

YangKeao:
@zhouqiang-cl The test hasn't failed. It seems that `chaosdaemon` doesn't have any test :cry:

zhouqiang-cl:
> The test hasn't failed. It seems that `chaosdaemon` doesn't have any test 😢

😝

Makefile
```diff
@@ -37,7 +37,7 @@ endif
 
 all: yaml build image
 
-build: chaosdaemon manager chaosfs dashboard dashboard-server-frontend
+build: manager chaosfs dashboard dashboard-server-frontend
```
zhouqiang-cl: why move out chaosdaemon

Makefile
zhouqiang-cl: Agree!

Makefile
YangKeao: chaosdaemon is built in `image` target now.

pkg/fakeclockgettime/fake_clock_gettime.c
zhouqiang-cl: Did the byte code has compatibility issues during Macos and linux

YangKeao: Yes. These byte codes can only run on linux 64-bit x86 platform.

As Mac OS programs use a different way to get time, we will not support time chaos on Mac OS. But if needed, we can support more cpu architecture in the future.

zhouqiang-cl: LGTM. then maybe we should add a `//TODO` in code

Makefile
zhouqiang-cl: Will you use a standalone Dockerfile in this pr?

pkg/fakeclockgettime/fake_clock_gettime.c
zhouqiang-cl: Soon. we need to support ARM. so I think later we should support other cpu architecture too.

lucklove:
"I use a simple go file to test", In the test it's better not to import C

YangKeao:
@lucklove OK, I have tested without `import "C"`.

Makefile
```diff
@@ -2,7 +2,8 @@
 LDFLAGS = $(if $(DEBUGGER),,-s -w) $(shell ./hack/version.sh)
 
 # SET DOCKER_REGISTRY to change the docker registry
-DOCKER_REGISTRY := $(if $(DOCKER_REGISTRY),$(DOCKER_REGISTRY),localhost:5000)
+DOCKER_REGISTRY_PREFIX := $(if $(DOCKER_REGISTRY),$(DOCKER_REGISTRY)/,)
```
YangKeao: Maybe in develop environment, we want to directly build image on pingcap/chaos-mesh, so that we will not need to modify config in helm to test.

Makefile
YangKeao: Yes! I have fixed this by build all binary in docker.

A Dockerfile and .dockerignore has been written in / to build binaries.

pkg/fakeclockgettime/fake_clock_gettime.c
YangKeao: OK. I add a `//TODO` right before `var fakeImage = []byte{`

zhouqiang-cl:
Did we need to support duration chaos in this pr

pkg/time/time.go
```diff
@@ -0,0 +1,117 @@
...
+	for _, e := range *program.Entries {
+		e := e
+
+		image, err := program.ReadSlice(e.StartAddress, uint64(len(fakeImage)))
+		if err != nil {
+			continue
+		}
+
+		if bytes.Equal(*image, fakeImage) {
+			fakeEntry = &e
```
lucklove: unreachable

Makefile
```diff
@@ -37,7 +38,7 @@ endif
 
 all: yaml build image
 
-build: chaosdaemon manager chaosfs dashboard dashboard-server-frontend
+build: dashboard-server-frontend
```
mahjonp: miss binary here?

zhouqiang-cl:
I think later we should also write TimeChaos Document @YangKeao

Makefile
YangKeao: No. Building binary on host is useless. All binary will be built in image.

For debug use (such as check grammer...), you can `make binary` manually.

pkg/time/time.go
YangKeao: Fixed

Makefile
YangKeao: Aha, it's the responsibility for privileged people to know what he is doing with `make docker-push`. 

Even without this Makefile, you can push to product env direcly by typing every `docker push` command. If you don't want this, set a `DOCKER_REGISTRY` variable.

Makefile
mahjonp: emm, I thought our CI run `make build`, but I'm wrong. That's ok.

pkg/time/time.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = program.WriteUint64ToAddr(fakeAddr+52, uint64(deltaNsec))
+	if err != nil {
+		return err
+	}
+
+	originAddr, err := program.FindSymbolInEntry("clock_gettime", vdsoEntry)
```
cwen0: Must `vdsoEntry` include `clock_gettime`?

api/v1alpha1/timechaos_types.go
```diff
@@ -0,0 +1,177 @@
...
+
+	// Value is required when the mode is set to `FixedPodMode` / `FixedPercentPodMod` / `RandomMaxPercentPodMod`.
+	// If `FixedPodMode`, provide an integer of pods to do chaos action.
+	// If `FixedPercentPodMod`, provide a number from 0-100 to specify the max % of pods the server can do chaos action.
+	// If `RandomMaxPercentPodMod`,  provide a number from 0-100 to specify the % of pods to do chaos action
+	// +optional
+	Value string `json:"value"`
+
+	// Selector is used to select pods that are used to inject chaos action.
+	Selector SelectorSpec `json:"selector"`
```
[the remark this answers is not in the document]
YangKeao: At present we didn't handle this situation, only the main process of container will be injected.

This could be fixed in the future by iterating in it's pid namespace and find their pid in ancestor's namespace. However, both two steps are not easy enough to contain in this PR.

doc/time_chaos.md
```diff
@@ -0,0 +1,45 @@
...
+* **mode** defines the mode to select pods.
+* **selector** specifies the target pods for chaos injection.
+* **timeOffset** specifies the offset of time. `sec` means the offset of seconds and `nsec` means the offset of nanoseconds.
+* **duration** defines the duration for each chaos experiment. In the sample file above, the network partition lasts for 10 seconds.
+* **scheduler** defines the scheduler rules for the running time of the chaos experiment. For more rule information, see <https://godoc.org/github.com/robfig/cron>.
+
+## Limitation
+
+* Time modification will only be injected into the main process of container
+* The process should get time through vdso `clock_gettime`. Time chaos has no effect on pure syscall `clock_gettime`
```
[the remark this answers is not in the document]
YangKeao: Yes. I should remove the first sentence.

pkg/ptrace/program.go
```diff
@@ -0,0 +1,444 @@
...
+}
+
+// Trace ptrace all threads of a process
+func Trace(pid int) (*TracedProgram, error) {
+	tidMap := make(map[int]bool)
+	for {
+		threads, err := ioutil.ReadDir(fmt.Sprintf("/proc/%d/task", pid))
+		if err != nil {
+			log.Error(err, "read failed", "pid", pid)
+			return constructPartialProgram(pid, tidMap), err
```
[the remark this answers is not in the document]
YangKeao: Yes. a `nil` is enough here.

pkg/ptrace/program.go
```diff
@@ -0,0 +1,444 @@
...
+		tids:       tids,
+		Entries:    nil,
+		backupRegs: nil,
+		backupCode: nil,
+	}
+}
+
+// Trace ptrace all threads of a process
+func Trace(pid int) (*TracedProgram, error) {
+	tidMap := make(map[int]bool)
```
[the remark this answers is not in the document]
YangKeao: Yes. I think `tid` is a common enough short name for `Thread ID`. It's used to avoid duplicate `ptrace` here.

pkg/ptrace/program.go
```diff
@@ -0,0 +1,444 @@
...
+			log.Error(err, "read failed", "pid", pid)
+			return constructPartialProgram(pid, tidMap), err
+		}
+
+		if len(threads) == len(tidMap) {
+			break
+		}
+
+		for _, thread := range threads {
+			tid64, err := strconv.ParseInt(thread.Name(), 10, 32)
```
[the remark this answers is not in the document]
YangKeao: But `ParseInt` will always return a `int64`. I turn it into a `int` right after checking error. However, I cannot rebind a variable in go.

Do you have suggestion on how to name it?

pkg/ptrace/ptrace.go
```diff
@@ -0,0 +1,455 @@
...
+	return nil
+}
+
+// PtraceWriteSlice uses ptrace rather than process_vm_write to write a buffer into addr
+func (p *TracedProgram) PtraceWriteSlice(addr uint64, buffer []byte) error {
+	wroteSize := 0
+
+	for wroteSize+ptrSize < len(buffer) {
+		addr := uintptr(addr + uint64(wroteSize))
+		data := buffer[wroteSize : wroteSize+ptrSize]
```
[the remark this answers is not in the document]
YangKeao: `PTRACE_POKEDATA` writes only one word with one request.

pkg/time/time.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	if fakeEntry == nil {
+		fakeEntry, err = program.MmapSlice(fakeImage)
+		if err != nil {
+			return err
+		}
+	}
+	fakeAddr := fakeEntry.StartAddress
+
+	err = program.WriteUint64ToAddr(fakeAddr+44, uint64(deltaSec))
```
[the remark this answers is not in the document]
YangKeao: As commented in `fakeImage`, `TV_SEC_DELTA` starts from 44 of `fakeImage` and `TV_NSEC_DELTA` starts from 52 of `fakeImage`.

As `fakeImage` is hard coded here, these offsets are also hard coded. The offsets in instructions (`0x1a` and `0x13`) and these two offsets (44 and 52) are all calculated manually.

pkg/time/time.go
YangKeao: Literally yes. Since LINUX_2.6 (when vdso is used in linux), it contains `clock_gettime`.

api/v1alpha1/timechaos_types.go
```diff
@@ -0,0 +1,177 @@
...
+)
+
+// +kubebuilder:object:root=true
+
+// TimeChaos is the Schema for the timechaos API
+type TimeChaos struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec defines the behavior of a pod chaos experiment
```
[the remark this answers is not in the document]
YangKeao: Fixed

api/v1alpha1/timechaos_types.go
```diff
@@ -0,0 +1,177 @@
...
+	Spec TimeChaosSpec `json:"spec"`
+
+	// +optional
+	// Most recently observed status of the chaos experiment about pods
```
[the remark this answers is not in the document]
YangKeao: Fixed

controllers/timechaos/types.go
```diff
@@ -0,0 +1,258 @@
+// Copyright 2019 PingCAP, Inc.
```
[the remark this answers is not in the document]
YangKeao: Fixed

controllers/timechaos/types.go
```diff
@@ -0,0 +1,258 @@
...
+
+const timeChaosMsg = "time is shifted with %v"
+
+// Reconciler is time-chaos reconciler
+type Reconciler struct {
+	client.Client
+	Log logr.Logger
+}
+
+// Reconcile reconcile a request from controller
```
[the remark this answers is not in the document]
YangKeao: Fixed

doc/time_chaos.md
```diff
@@ -0,0 +1,45 @@
...
+```
+
+> For more sample files, see [examples](../examples). You can edit them as needed. 
+
+Description:
+
+* **mode** defines the mode to select pods.
+* **selector** specifies the target pods for chaos injection.
+* **timeOffset** specifies the offset of time. `sec` means the offset of seconds and `nsec` means the offset of nanoseconds.
+* **duration** defines the duration for each chaos experiment. In the sample file above, the network partition lasts for 10 seconds.
```
ethercflow: s/network partition/time chaos/g ?

pkg/ptrace/ptrace.go
```diff
@@ -0,0 +1,483 @@
...
+func waitPid(pid int) error {
+	ret := waitpid(pid)
+	if ret == pid {
+		return nil
+	}
+
+	return errors.Errorf(waitPidErrorMessage, ret)
+}
+
+func constructPartialProgram(pid int, tidMap map[int]bool) *TracedProgram {
```
YangKeao: I should use better way (such as defer and check ret value) to handle error here.

pkg/ptrace/ptrace.go
```diff
@@ -0,0 +1,483 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	return regs.Rax, p.Restore()
+}
+
+// Mmap runs mmap syscall
+func (p *TracedProgram) Mmap(length uint64, fd uint64) (uint64, error) {
+	return p.Syscall(9, 0, length, 7, 0x22, fd, 0)
```
cwen0: why use `0x22`? Can we use the const `syscall.PR_MCE_KILL_GET` install of it?

pkg/ptrace/ptrace.go
```diff
@@ -0,0 +1,483 @@
...
+
+			_, ok := tidMap[tid]
+			if ok {
+				continue
+			}
+
+			err = syscall.PtraceAttach(tid)
+			if err != nil {
+				log.Error(err, "attach failed", "tid", tid)
+				return constructPartialProgram(pid, tidMap), errors.WithStack(err)
```
ethercflow: In my mind that `ptrace attach` will only stop the corresponding thread, so if the subsequent threads are ignored and still running, is there a risk of panic? It may be safer to send a `SIGSTOP` signal to the process pid before attaching it.

YangKeao: I looped over /proc/[pid]/tasks to attach to all threads so that all threads will stop.

Actually I want to check every thread's current instruction to make sure I am not injecting while it's inside `clock_gettime` function, but this feature hasn't been implemented :cry:

pkg/time/time.go
```diff
@@ -0,0 +1,122 @@
...
+	}()
+	if err != nil {
+		return err
+	}
+
+	var vdsoEntry *mapreader.Entry
+	for _, e := range *program.Entries {
+		e := e
+		if e.Path == "[vdso]" {
+			vdsoEntry = &e
```
ethercflow: add `break`? BTW reverse iter faster

pkg/ptrace/ptrace.go
YangKeao: Oh! I should replace it with `syscall.MAP_ANON|syscall.MAP_PRIVATE`

pkg/ptrace/ptrace.go
```diff
@@ -0,0 +1,484 @@
...
+	if err != nil {
+		return 0, err
+	}
+
+	return regs.Rax, p.Restore()
+}
+
+// Mmap runs mmap syscall
+func (p *TracedProgram) Mmap(length uint64, fd uint64) (uint64, error) {
+	return p.Syscall(9, 0, length, syscall.PROT_READ|syscall.PROT_WRITE|syscall.PROT_EXEC, syscall.MAP_ANON|syscall.MAP_PRIVATE, fd, 0)
```
cwen0: why is the number `9`?

YangKeao: It is replaced with `syscall.SYS_MMAP` now :beers:

cwen0:
@YangKeao Fix CI, please

YangKeao:
@cwen0 @lucklove  PTAL

ethercflow:
LGTM

PR #228, opened 2020-02-12
Expose chaos controller metrics in helm chart

yeya24:

This is a prerequisite of adding Prometheus to chaos-mesh. Exposing 8080 in chaos controller-manager enables Prometheus to get metrics from controller-manager.

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
Change controller-manager default metrics port to 10080
```

### DIFF
--- a/helm/chaos-mesh/templates/controller-manager-deployment.yaml
+++ b/helm/chaos-mesh/templates/controller-manager-deployment.yaml
@@ -56,7 +56,7 @@ spec:
           - name: webhook
             containerPort: 9443 # Customize containerPort
           - name: http
-            containerPort: 8080
+            containerPort: 10080
       volumes:
         - name: webhook-certs
           secret:
--- a/helm/chaos-mesh/templates/controller-manager-deployment.yaml
+++ b/helm/chaos-mesh/templates/controller-manager-deployment.yaml
@@ -53,8 +53,10 @@ spec:
             mountPath: /etc/webhook/certs
             readOnly: true
         ports:
-          - name: webhook-api
+          - name: webhook
             containerPort: 9443 # Customize containerPort
+          - name: http
+            containerPort: 8080
       volumes:
         - name: webhook-certs
           secret:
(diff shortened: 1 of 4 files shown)


helm/chaos-mesh/templates/controller-manager-service.yaml
```diff
@@ -11,12 +11,12 @@ metadata:
 spec:
   type: {{ .Values.controllerManager.service.type }}
   ports:
-#    - port: {{ .Values.controllerManager.service.port }}
-#      targetPort: http
-#      protocol: TCP
-#      name: http
+    - port: 8080
```
yeya24: I remove the `port` var here since I think it is useless. We can just use a hardcode 8080 here

helm/chaos-mesh/templates/controller-manager-deployment.yaml
```diff
@@ -53,8 +53,10 @@ spec:
             mountPath: /etc/webhook/certs
             readOnly: true
         ports:
-          - name: webhook-api
+          - name: webhook
             containerPort: 9443 # Customize containerPort
+          - name: http
+            containerPort: 8080
```
cwen0: Can we switch to an infrequently used port? eg: 20180

yeya24: May I ask any difference between 8080 and 20180? Since we are not using hostport here and any access to controller-manager is via Service. So I think any port is okay. I am not sure if I miss something.

cwen0: I think `8080` is used for common service and this port may be used by chaos-dashboard or the other services. At here, we just need to export our metric.

yeya24: Metric is also an HTTP endpoint. And the default value is 8080 as well https://github.com/pingcap/chaos-mesh/blob/09afa9d7074b88bdb143e6f8bcbdd9e7b69c4b3c/cmd/controller-manager/main.go#L67. Do you think we need to change it?

cwen0: I think we can change it.

yeya24: OK. I will update them in this PR

yeya24: Updated. Please take a look

PR #236, opened 2020-02-14
Correct minikube restriction: sch_netem is loaded by default in v1.6.

LaumiH:
Signed-off-by: Laura Henning <laura-marie.henning@stud.h-da.de>

The restrictions for minikube are not up to date: netem related chaos can be run in default minikube vm in versions >= 1.6 now

I did a PR to minikube to make a tool called pumba work, which also uses tc netem. I noticed the kernel module was missing, and minikube guys accepted my PR to add it :)

This is just a change to the README of this project, so I ran no additional tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,9 +169,9 @@ You can try Chaos Mesh on your local K8s environment deployed using `kind` or `m
 
 There are some known restrictions for Chaos Operator deployed on `minikube` clusters:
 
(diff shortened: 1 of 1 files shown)


YangKeao:
Wow! Thank you🎉

zhouqiang-cl:
@LaumiH thank you very much

PR #244, opened 2020-02-15
Make pod/container-kill controlled by twophase controller in scheduler mode

Yisaer:

This requests make container/pod-kill chaos only to maintain their own logic without where the calling from in order to support trigger chaos 
And I also let the pod/container-kill chaos controlled by twophase controller by hacking the recover duration as 0 second when duration is nil.

```release-note
NONE
```


zhouqiang-cl:
> And I also think the logic in specific chaos action controller should only contain performing the chaos action without caring where the calling from.

I think so. Now we add a new chaos, then need to add two types of chaos, we should make it more simple

cwen0:
@Yisaer If you remove these checks, the scheduler of chaos object won't work.

cwen0:
> @Yisaer If you remove these checks, the scheduler of chaos object won't work.

It is a bug here, Can you fix it?

Yisaer:
> @Yisaer If you remove these checks, the scheduler of chaos object won't work.

@cwen0 Thanks for reminding me. I previously had a misunderstanding for `container-kill` and `pod-kill` reconciler.
It seems we can't use twophase controller to directly control the `pod/container`-killl action yet. However, I think there still exists the way to achieve it. I would try to figure it out.

Yisaer:
ref https://github.com/pingcap/chaos-mesh/issues/246

Yisaer:
@zhouqiang-cl @cwen0  PTAL

controllers/twophase/types.go
```diff
@@ -83,11 +83,16 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 	}
 
 	scheduler := chaos.GetScheduler()
-	if duration == nil || scheduler == nil {
-		r.Log.Info("scheduler and duration should be defined currently")
+	if scheduler == nil {
+		r.Log.Info("scheduler should be defined currently")
```
cwen0: For `container-kill` and `pod-kill`, this is a normal log, for the other action, this is an error.

Yisaer: Not actually, the twophase controller let the `duration` become 0 second when it is omitted which would take effect nearly 10 seconds as default resync interval.
And further more, we can validate the chaos Spec in their own chaos kind controller or validation webhook when created or updated.

cwen0: Sorry, I misunderstand this check. I think we need to return `ctrl.Result{RequeueAfter: recover-duration}` when to set recover time.

cwen0:
@YangKeao PTAL

PR #251, opened 2020-02-17
Support pprof

### LINKED ISSUE #238: Enable pprof in all conponents of Chaos Mesh
## Feature Request

Sometimes we need to locate some Chaos Mesh issue, such as performance, memory leaks, etc. I think `pprof` can help us. So I recommend enabling `pprof` in all components of chaos-mesh

yeya24:

Fixes #238

For chaos-damon, just registering pprof handler to the http sever.
But for controller-manager, because https://github.com/kubernetes-sigs/kubebuilder/issues/1225, it would be not that clear to use the same metrics port in pprof. So I add a new pprof port, which can be disabled by setting "0"

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


cmd/controller-manager/main.go
```diff
@@ -65,6 +68,7 @@ func init() {
 
 func parseFlags() {
 	flag.StringVar(&metricsAddr, "metrics-addr", ":10080", "The address the metric endpoint binds to.")
+	flag.StringVar(&pprofAddr, "pprof-addr", "0", "The address the pprof endpoint binds to.")
```
zhouqiang-cl: I think we should enable pprof default

helm/chaos-mesh/values.yaml
```diff
@@ -24,6 +24,9 @@ controllerManager:
   service:
     type: ClusterIP
 
+  # enableProfiling is a flag to enable pprof in controller-manager
+  enableProfiling: false
```
Yisaer: I tend to make `enableProfiling` as a global switch to control to pprof both for `manager` and daemon`.

yeya24: Thanks for your review! Updated.

PR #273, opened 2020-02-21
support kernel chaos

ethercflow:

Support kernel fault injection, please read [inject_example.txt](https://github.com/iovisor/bcc/blob/master/tools/inject_example.txt) to learn more.

1. Add a new chaos type;
2. Send request to bpfki instead of chaos-daemon

### Check List 

Tests 

 - No code

Code changes

 - Has Go code change

Side effects

 - None

Related changes

```release-note
NONE
```


api/v1alpha1/kernelchaos_types.go
```diff
@@ -0,0 +1,185 @@
...
+	return in.Value
+}
+
+type FailKernRequest struct {
+	FailType int32 `json:"failtype"`
+	Headers []string `json:"headers"`
+	Callchain []Frame `json:"callchain"`
+	Probability uint32 `json:"probability"`
+	Times uint32 `json:"times"`
+};
```
you06: May we remove the semicolon?

api/v1alpha1/kernelchaos_types.go
```diff
@@ -0,0 +1,185 @@
...
+
+type Frame struct {
+	Funcname string `json:"funcname"`
+	Parameters string `json:"parameters"`
+	Predicate string `json:"predicate"`
+};
```
you06: ditto

Yisaer:
@ethercflow  Can this request split into several requests? I think we can start from the API definition.

api/v1alpha1/kernelchaos_types.go
```diff
@@ -0,0 +1,185 @@
...
+	// If `FixedPercentPodMod`, provide a number from 0-100 to specify the max % of pods the server can do chaos action.
+	// If `RandomMaxPercentPodMod`,  provide a number from 0-100 to specify the % of pods to do chaos action
+	// +optional
+	Value string `json:"value"`
+
+	// Selector is used to select pods that are used to inject chaos action.
+	Selector SelectorSpec `json:"selector"`
+
+	// FailKernRequest defines the request of kernel injection
+	FailKernRequest FailKernRequest `json:"failKernRequest"`
```
AngleNet: s/FailKernRequest/FailKernelRequest?

pkg/utils/grpc.go
```diff
@@ -32,12 +32,7 @@ const DefaultRPCTimeout = 60 * time.Second
 // RPCTimeout specifies timeout of RPC between controller and chaos-operator
 var RPCTimeout = DefaultRPCTimeout
 
-func CreateGrpcConnection(ctx context.Context, c client.Client, pod *v1.Pod) (*grpc.ClientConn, error) {
-	port := os.Getenv("CHAOS_DAEMON_PORT")
-	if port == "" {
-		port = "8080"
-	}
-
+func CreateGrpcConnection(ctx context.Context, c client.Client, pod *v1.Pod, port string) (*grpc.ClientConn, error) {
```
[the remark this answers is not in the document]
AngleNet: Adding port to the signature seems to be redudant ?

yeya24: I think it's fine. But each time, get the port from env variables like `CreateGrpcConnection(ctx, r.Client, pod, os.Getenv("CHAOS_DAEMON_PORT"))` is not good. It is better to use variable keep the port (daemon/bpfki). But it is fine in this PR. Maybe we can open an issue about it and change it in the future.

controllers/kernelchaos/types.go
```diff
@@ -0,0 +1,316 @@
...
+	"github.com/pingcap/chaos-mesh/controllers/twophase"
+	"github.com/pingcap/chaos-mesh/pkg/utils"
+
+	pb "github.com/pingcap/chaos-mesh/pkg/chaosdaemon/pb"
+	pb_ "github.com/pingcap/chaos-mesh/pkg/chaoskernel/pb"
+)
+
+const kernelChaosMsg = "kernel is injected with %v"
+
+// Reconciler is KernelChoas reconciler
```
yeya24: s/KernelChoas/KernelChaos/g

controllers/kernelchaos/types.go
```diff
@@ -0,0 +1,316 @@
...
+			Message:   fmt.Sprintf(kernelChaosMsg, kernelchaos.Spec.FailKernRequest),
+		}
+
+		kernelchaos.Status.Experiment.Pods = append(kernelchaos.Status.Experiment.Pods, ps)
+	}
+
+	return nil
+}
+
+// Recover means the reonciler recover the chaos action
```
yeya24: This should be `Recover means the reconciler recovers the chaos action`

helm/chaos-mesh/values.yaml
```diff
@@ -139,3 +139,10 @@ preJobs:
 
 postJobs:
   podAnnotations: {}
+
+bpfki:
+  image: pingcap/chaos-kernel:latest
+  imagePullPolicy: Always
+  grpcPort: 50051
+
+  podAnnotations: {}
```
yeya24: Seems bpfki is deployed in the same pod with chaos-daemon, so what about moving this part to be part of `chaos-daemon` in `values.yaml`? What do you think?

And I am not that familar with bpfki, can I ask is there any case that we don't want to deploy bpfki, only deploy chaos-daemon?  If that, I think it is better to add an `enable` option

zhouqiang-cl: bpfki have performance issues on every container(not only the inject one)

pkg/chaoskernel/pb/bpfki.proto
```diff
@@ -0,0 +1,55 @@
...
+}
+
+service BPFKIService {
+  rpc SetTimeVal(BumpTimeRequest) returns (StatusResponse) {}
+  rpc RecoverTimeVal(BumpTimeRequest) returns (StatusResponse) {}
+  rpc SetTimeSpec(BumpTimeRequest) returns (StatusResponse) {}
+  rpc RecoverTimeSpec(BumpTimeRequest) returns (StatusResponse) {}
+  rpc FailMMOrBIO(FailKernRequest) returns (StatusResponse) {}
+  rpc RecoverMMOrBIO(FailKernRequest) returns (StatusResponse) {}
+  rpc FailSyscall(FailSyscallRequest) returns (StatusResponse) {}
```
cwen0: Where does use `FailSyscall`?

doc/kernel_chaos.md
```diff
@@ -0,0 +1,86 @@
...
+> Cannot allocate memory
+> Invalid argument
+
+> For more sample files, see [examples](https://github.com/chaos-mesh/bpfki/tree/develop/examples). You can edit them as needed. 
+
+Description:
+
+* **mode** defines the mode to select pods.
+* **selector** specifies the target pods for chaos injection.
+* **failkernRequest** defines the specified injection mode (kmalloc,bio,etc) with a call chain and an optional set of predicates.
```
cwen0: add more description about failKernRequest

cwen0: What is the difference between `kmalloc` and `bio`?

ethercflow: `bio` is a concept of block layer, so it's used to inject io request error. 
`kmalloc` is the mostly caller of `slab`.

cwen0:
@ethercflow fix conflict?

api/v1alpha1/kernelchaos_types.go
```diff
@@ -0,0 +1,215 @@
...
+	// Probability indicates the fails with probability.
+	// If you want 1%, please set this field with 1.
+	Probability uint32 `json:"probability,omitempty"`
+
+	// Times indicates the max times of fails.
+	Times uint32 `json:"times,omitempty"`
+}
+
+// Frame defines the function signature and predicate in function's body
+type Frame struct {
```
cwen0: add some comments for these

controllers/kernelchaos_controller.go
```diff
@@ -0,0 +1,53 @@
...
+
+	reconciler := kernelchaos.Reconciler{
+		Client: r.Client,
+		Log:    logger,
+	}
+
+	return reconciler.Reconcile(req)
+}
+
+// SetupWithManager setups a kernel chaos reconciler on controller-manager
```
fewdan: “setup” is not a verb.   Maybe we should s/setups/sets up/g ?

Issue #274, opened 2020-02-21
Support validate chaos obj by admission Webhook


Yisaer:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

Add admission webhook to support validate the chaos obj.
For example, `pod-failure`/`network`/`io` chaos 's `schedule` and `duration` should be omitted or defined at the same time.
`container-kill` and `pod-kill` shoudn't defined scheduler property.

cwen0:
@Yisaer I don't understand what you mean, Do you mean that `container-kill` / `pod-kill` should not support scheduler fields?

Yisaer:
> @Yisaer I don't understand what you mean, Do you mean that `container-kill` / `pod-kill` should not support scheduler fields?

My fault, I mean duration propetry

PR #279, opened 2020-02-25
Add validate admission webhook for chaos api

Yisaer:
#274 

Adding new validation admission webhook to validate chaos api property when they are created or updated. This request only consider the `scheduler` and `duration` validation for now.

Code changes

 - Has Go code change

```release-note
NONE
```


YangKeao:
Please upgrade your controller-gen with `go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.5`

api/webhook/validate.go
```diff
@@ -0,0 +1,43 @@
...
+
+// ChaosValidator used to handle the validation admission request
+type ChaosValidator struct {
+}
+
+// Handle handle the requests from validation admission requests
+func (v *ChaosValidator) Handle(ctx context.Context, req admission.Request) admission.Response {
+	name := req.Name
+	namespace := req.Namespace
+	kind := req.Kind
```
yeya24: Is it better to directly use `kind := req.Kind.Kind` here?

pkg/webhook/validation/validate.go
```diff
@@ -0,0 +1,136 @@
...
+		if podChaos.Spec.Scheduler == nil {
+			return false, invalidConfigurationMsg, nil
+		}
+		return true, "", nil
+	case v1alpha1.ContainerKillAction:
+		// We choose to ignore the Duration property event user define it
+		if podChaos.Spec.Scheduler == nil {
+			return false, invalidConfigurationMsg, nil
+		}
+		return true, "", nil
```
yeya24: Maybe in the case of `container kill`, we need to check if `podchaos.Spec.ContainerName == ""` ? I am not sure about this. Because the logic here seems quite odd to me. https://github.com/pingcap/chaos-mesh/blob/2dc58db9600f9088eacfe00743ee8f4fbe1c72f4/controllers/podchaos/containerkill/types.go#L75 

In my opinion, this field could be empty if the pod only has 1 container, then uses that container's name is fine. But if it has multiple containers, then the `containerName` field is required.

Yisaer: It's ok to me about your suggestion. However, in this request,  I suggest we only consider the `scheduler` and `duration` property validation. After that, we could enrich the validation logic for each kind of chaos API in parallel.

api/webhook/validate.go
Yisaer: updated.

pkg/webhook/validation/validate.go
yeya24: That sounds good

pkg/webhook/validation/validate.go
```diff
@@ -0,0 +1,136 @@
...
+
+	return &admissionv1beta1.AdmissionResponse{
+		Allowed: permitted,
+		Result: &metav1.Status{
+			Message: msg,
+		},
+	}
+}
+
+func validatePodchaos(rawObj []byte) (bool, string, error) {
```
YangKeao: Should we moidfy the development guide to show that we can register validate hook here?

Yisaer: It's ok to me.

api/webhook/validate.go
```diff
@@ -0,0 +1,43 @@
...
+// ChaosValidator used to handle the validation admission request
+type ChaosValidator struct {
+}
+
+// Handle handle the requests from validation admission requests
+func (v *ChaosValidator) Handle(ctx context.Context, req admission.Request) admission.Response {
+	name := req.Name
+	namespace := req.Namespace
+	kind := req.Kind.Kind
+	validatelog.Info(fmt.Sprintf("receive validation req for obj[%s/%s/%s]", kind, namespace, name))
```
cwen0: s/receive/Receive/g

pkg/webhook/validation/validate.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/pingcap/chaos-mesh/api/v1alpha1"
+)
+
+const (
+	invalidConfigurationMsg = "invalid configuration"
+)
+
+var log = ctrl.Log.WithName("validate-webhook")
+
+// ValidateChaos handle the validation for chaos api
```
cwen0: s/handle/handles/g

pkg/webhook/validation/validate.go
cwen0: Can we implement different chaos validation functions in the corresponding chaos definition?

Yisaer: > Can we implement different chaos validation functions in the corresponding chaos definition?

I didn't get your point. Can you provide more detail?

Yisaer:
@YangKeao @cwen0  PTAL

helm/chaos-mesh/values.yaml
```diff
@@ -18,8 +18,8 @@ controllerManager:
 
   replicaCount: 1
 
-  image: pingcap/chaos-mesh:latest
-  imagePullPolicy: Never
+  image: yisa/chaos:latest
```
yeya24: Update this?

helm/chaos-mesh/templates/pre-delete-job.yaml
```diff
@@ -36,4 +36,4 @@ spec:
             - |
               set -e
               kubectl delete secrets -n {{ .Release.Namespace }} {{ template "chaos-mesh.certs" . }} || true
-{{- end }}
+{{- end }}
```
yeya24: Add a new line here

helm/chaos-mesh/values.yaml
Yisaer: updated and fixed.

pkg/webhook/validation/validate.go
```diff
@@ -0,0 +1,136 @@
...
+	case v1alpha1.PodFailureAction:
+		if podChaos.Spec.Duration != nil && podChaos.Spec.Scheduler != nil {
+			return true, "", nil
+		} else if podChaos.Spec.Duration == nil && podChaos.Spec.Scheduler == nil {
+			return true, "", nil
+		} else {
+			return false, invalidConfigurationMsg, nil
+		}
+	case v1alpha1.PodKillAction:
+		// We choose to ignore the Duration property event user define it
```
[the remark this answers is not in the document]
Yisaer: fixed, it should be `even` here.

pkg/webhook/validation/validate.go
cwen0: eg: in `api/v1alpha1/iochaos_types.go` 
```
func (i *IOChaos) Validate{
xxxxx
}
```

cwen0: we can define an `interface` with `Validate ` function for all Chaos  and require all Chaos struct to implement this interface . Here we just call this method and don't care about the specific type of chaos and the specific implementation

zhouqiang-cl: @Yisaer PTAL again

Yisaer: updated.

Yisaer:
@cwen0 @yeya24  @YangKeao  PTAL

helm/chaos-mesh/templates/post-install-webhook-job.yaml
```diff
@@ -56,6 +56,23 @@ spec:
...
+                    service:
+                      name: {{ template "chaos-mesh.svc" . }}
+                      namespace: {{ .Release.Namespace }}
+                      path: "/validate-v1alpha1-chaos"
+                    caBundle: "${CA_BUNDLE}"
+                  rules:
+                    - operations: [ "CREATE","UPDATE"]
+                      apiGroups: ["pingcap.com"]
+                      apiVersions: ["v1alpha1"]
+                      resources: ["podchaos","iochaos","timechaos","networkchaos"]
```
Gallardot: `post-install-webhook-job` is only update the `caBundle`. I don't think we need update the `rules` fields too.

Yisaer: good catch

PR #280, opened 2020-02-25
Update controller-gen version to fix CI

cwen0:

fix CI


YangKeao:
This will only fix those PR which doesn't make any change to yaml. User with lower version of `controller-gen` will still procude "wrong" yaml files.

How can we alert developer to upgrade `controller-gen`?

cwen0:
> This will only fix those PR which doesn't make any change to yaml. User with lower version of `controller-gen` will still procude "wrong" yaml files.
> 
> How can we alert developer to upgrade `controller-gen`?

How about adding `go get xxxx@v0.2.5` before running `make yaml`?

Issue #178, opened 2020-02-02
UCP: Use the namespace of chaos object as the namespace selector if the namespaces selector was unset


mahjonp:
## Description

At present, the organization of chaos crd and the system under test(SUT) is not very friendly. 

Refer to documents of chaos-mesh, crds are placed in the [chaos-testing](https://github.com/pingcap/chaos-mesh/blob/4613209daea617519adb00d6ea596b270874b906/examples/pod-failure-example.yaml#L5) namespace, There are currently 2 problems:
* There is currently no intuitive way to see which chaos rules exist on one SUT if chaos rule was not placed in the same namespace with SUT(Components of system under test are usually organized in same namespace).
* When the SUT is destroyed, the corresponding chaos crd may still exist under the chaos-testing namespace, and it will take effect again the next time when another system with the same namespace name is created.

Current design enable one chaos rule take effect on different namespaces, but that's not necessary and it's better to bind a chaos rule with only one namespace IMO.

## Score 

* 300

## Mentor(s)

* @fewdan 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Golang
* Kubernetes

## Learning Materials

* [Development guide of chaos-mesh](https://github.com/pingcap/chaos-mesh/wiki/Development-Overview)

YangKeao:
So, the requested feature is to automatically validate through webhook or something to ensure that chaos rules are placed in the same namespace with its objects? @mahjonp

mahjonp:
> So, the requested feature is to automatically validate through webhook or something to ensure that chaos rules are placed in the same namespace with its objects? @mahjonp

Yeah, that's what I mean.

cwen0:
> So, the requested feature is to automatically validate through webhook or something to ensure that chaos rules are placed in the same namespace with its objects? @mahjonp 

I think this requirement is a bit strict because it is possible that the chaos objects involves multiple namespaces.

cwen0:
[...]

Issue #291, opened 2020-02-28
We should remove the dependence on concrete chaos struct from validation pkg


cwen0:
## Feature Request

In [validation](https://github.com/pingcap/chaos-mesh/tree/master/pkg/webhook/validation) pkg, the current implementation needs to use the concrete chaos struct ([L65-94](https://github.com/pingcap/chaos-mesh/blob/master/pkg/webhook/validation/validate.go#L65-L94)), which is very unfriendly for developing new chaos type.  

**Describe the feature you'd like:**

I think we only need to use the concrete chaos struct in [API](https://github.com/pingcap/chaos-mesh/tree/master/api/v1alpha1)  pkg and [controller](https://github.com/pingcap/chaos-mesh/tree/master/controllers) pkg. 

We can define a function in `API` pkg for `validation pkg`,  like: 

```
ParseChaosForValidation(data []bytes) Validation {
.... 
}
```

Gallardot:
@cwen0 please assigned this to me.I will try this after the current work dones.I think it's the related work.

cwen0:
fixed in #338

PR #300, opened 2020-02-29
Use the namespace of chaos object as the default namespace selector

Gallardot:
#178 

- Use the namespace of chaos object as the namespace selector if the namespaces selector was unset.
- Change the method name to make the semantics clearer
- Use a more generic `rest.Config`.  Development environment is more friendly.
- Updated the Unit test.
- Updated the docs.

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

Side effects

 - None

Related changes

 - Update the documentation

```release-note
NONE
```


zhouqiang-cl:
@Yisaer @cwen0 @yeya24 @YangKeao PTAL

controllers/iochaos/fs/types.go
```diff
@@ -84,7 +84,7 @@ func (r *Reconciler) Apply(ctx context.Context, req ctrl.Request, chaos reconcil
 		return err
 	}
 
-	pods, err := utils.SelectAndGeneratePods(ctx, r.Client, &iochaos.Spec)
+	pods, err := utils.SelectAndFilterPods(ctx, r.Client, &iochaos.Spec, iochaos.ObjectMeta)
```
yeya24: Is it okay to just pass namespace? Passing objectMeta seems odd.

pkg/utils/selector.go
```diff
@@ -207,8 +217,8 @@ func CheckPodMeetSelector(pod v1.Pod, selector v1alpha1.SelectorSpec) (bool, err
 	return false, nil
 }
 
-// GeneratePods generate pods according to mode from pod list
-func GeneratePods(pods []v1.Pod, mode v1alpha1.PodMode, value string) ([]v1.Pod, error) {
+// FilterPodsByMode filter pods according to mode from pod list
```
yeya24: I think `FilterPodsByMode filters pods by mode` is better. What do you think?

pkg/utils/selector.go
```diff
@@ -40,12 +42,13 @@ type SelectSpec interface {
 	GetValue() string
 }
 
-func SelectAndGeneratePods(ctx context.Context, c client.Client, spec SelectSpec) ([]v1.Pod, error) {
+// SelectAndFilterPods returns the list of pods that filtered by selector and filtered by the PodMode
```
yeya24: that filtered by selector and PodMode

controllers/iochaos/fs/types.go
Yisaer: I agreed.

pkg/utils/selector.go
```diff
@@ -106,10 +109,17 @@ func SelectPods(ctx context.Context, c client.Client, selector v1alpha1.Selector
 
 	pods = append(pods, podList.Items...)
 
-	namespaceSelector, err := parseSelector(strings.Join(selector.Namespaces, ","))
+	namespaceSelector, err := parseSelector(objMeta.GetNamespace())
```
Yisaer: It seems this change would only let namespaceSelector using the chaos object own namespace instead of the selected namespaces?

Yisaer:
First, I agreed that chaos object would fulfill the namespace selector with its own namespace when the namespace selector is no set. 
I suggest not to change the logic in controller and default the chaos object by mutation admission webhook.
@Gallardot

pkg/utils/selector.go
Gallardot: In fact, the full code for namespaceSelector is shown below
```golang
	namespaceSelector, err := parseSelector(objMeta.GetNamespace())
	if err != nil {
		return nil, err
	}
	if len(selector.Namespaces) != 0 {
		namespaceSelector, err = parseSelector(strings.Join(selector.Namespaces, ","))
		if err != nil {
			return nil, err
		}
	}
```
[code](https://github.com/Gallardot/chaos-mesh/blob/b320467cc44f8f61d1774e8ab4f5a8e7052be94d/pkg/utils/selector.go#L112-L121)

1. using the chaos object own namespace as default.
2. If `selector.Namespaces` is set, instead of the default.

Gallardot:
> I suggest not to change the logic in controller and default the chaos object by mutation admission webhook.

@Yisaer I don't get your point. This logic is for the controller, How to do this by mutation admission webhook?

pkg/utils/selector.go
Gallardot: Thanks. I agree with you. I will fix it.

controllers/iochaos/fs/types.go
Gallardot: I actually also just pass `namespace` before.But I think 'SelectAndFilterPods' is a very important and widely used function.Almost all chaos objects will invoke this function. And We may extend its logic by this [issue](https://github.com/pingcap/chaos-mesh/issues/193) . The best choice is pass the `InnerObject`.  But it will takes more time to refactor.

Yisaer:
@Gallardot 
The mutation admission webhook would watch the `create` and `update` event for chaos obj api, if the obj's namespace selector is not set, it would update its spec to set the namespace selector as same as its own namespace property.

Gallardot:
> The mutation admission webhook would watch the `create` and `update` event for chaos obj api, if the obj's namespace selector is not set, it would update its spec to set the namespace selector as same as its own namespace property.

@Yisaer  I get your point. May be like this [issue](https://github.com/pingcap/chaos-mesh/issues/292) . Use MutatingWebhook to implement the default value feature.

Gallardot:
> The mutation admission webhook would watch the `create` and `update` event for chaos obj api, if the obj's namespace selector is not set, it would update its spec to set the namespace selector as same as its own namespace property.

@cwen0  What is your opinion? If use mutation admission webhook, we can also solve this [issue](https://github.com/pingcap/chaos-mesh/issues/292) .

cwen0:
@Gallardot I didn't think of using mutation admission-webhook to solve this problem before. I'm really sorry, but I agree with @Yisaer.

Gallardot:
@Yisaer  @cwen0 

It's OK. I will use MutatingWebhook to implement the default value feature.

zhouqiang-cl:
@Gallardot can we review this pr now

Gallardot:
> @Gallardot can we review this pr now

@zhouqiang-cl Sorry. It's still WIP.

zhouqiang-cl:
> @zhouqiang-cl Sorry. It's still WIP.

You are welcome

Gallardot:
@Yisaer @cwen0 @yeya24 @YangKeao PTAL

zhouqiang-cl:
@cwen0 After this pr is merged. I think we should also update readme and examples

Gallardot:
> @cwen0 After this pr is merged. I think we should also update readme and examples

@zhouqiang-cl  I think so. Current doc [dev_hello_world.md](https://github.com/pingcap/chaos-mesh/blob/master/doc/dev_hello_world.md#develop-a-new-chaos) is not developer friendly about how to develop a new chaos using the kubebuilder.

Gallardot:
@cwen0  
[controller-tools](https://github.com/kubernetes-sigs/controller-tools/releases) v0.2.6 is released today.It causes the current jenkins job failed.

Why not we specify a version?
for example
```
controller-gen:
	$(GO) get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.5
```

What's your opinion？

YangKeao:
Don't worry. This CI error will be fixed in #330 (by locking controller-tools version)

Gallardot:
> Don't worry. This CI error will be fixed in #330 (by locking controller-tools version)

Good job！

api/v1alpha1/common_webhook.go
```diff
@@ -0,0 +1,21 @@
+// Copyright 2019 PingCAP, Inc.
```
cwen0: s/2019/2020/g

api/v1alpha1/common_webhook_test.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2019 PingCAP, Inc.
```
cwen0: ditto

api/v1alpha1/iochaos_webhook.go
```diff
@@ -0,0 +1,41 @@
+// Copyright 2019 PingCAP, Inc.
```
cwen0: ditto

api/v1alpha1/kernelchaos_webhook.go
```diff
@@ -0,0 +1,43 @@
+/*
```
cwen0: unity the LICENSE head

cmd/controller-manager/main.go
```diff
@@ -115,6 +115,10 @@ func main() {
 		setupLog.Error(err, "unable to create controller", "controller", "PodChaos")
 		os.Exit(1)
 	}
+	if err = (&chaosmeshv1alpha1.PodChaos{}).SetupWebhookWithManager(mgr); err != nil {
```
cwen0: Can we just use a webhook for all Chaos object?  like: https://github.com/pingcap/chaos-mesh/blob/master/pkg/webhook/validation/validate.go

YangKeao: But this way seems better. Should we modify the validate webhook into this form?

cwen0: @YangKeao Your proposal is better.

Gallardot: > @cwen0 But this way seems better (though it has so many files and duplicate codes). Should we modify the validate webhook into this form by implementing `admission.Validator` for them?

@YangKeao agree with you.  On this issue #291 , I want to refactoring the validate webhook by implementing `admission.Validator`

So many files and duplicate codes is autogenerated by kubebuilder. I think we can just follow kubebuilder‘s rules.

cwen0: @Gallardot LGTM

api/v1alpha1/common_webhook.go
Gallardot: All done.

PR #306, opened 2020-03-03
use local file instead of downloading from github and ensure pod status 

cwen0:

fix install chaos mesh failed caused by the access to github

* use local file 
* ensure pod status  
* fix sort invalid option

### Check List 

Tests 

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -767,9 +767,6 @@ install_chaos_mesh() {
 }
 
 version_lt() {
-    echo $1
-    echo $2
-
     vercomp $1 $2
     if [ $? == 2 ];  then
         return 0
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -445,11 +465,145 @@ EOF
...
+        kind load docker-image ${volume_provisioner_image} > /dev/null 2>&1 || true
+    fi
+
+    cat <<EOF >"${config_file}"
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
+metadata:
+  name: "local-storage"
+provisioner: "kubernetes.io/no-provisioner"
+volumeBindingMode: "WaitForFirstConsumer"
```
Yisaer: Why we install local-pv here ?

cwen0: this is just an available option for developer

PR #323, opened 2020-03-08
Use the `pause` image to inject pod-failure chaos

### LINKED ISSUE #189: UCP : Use a real image for fake image
## Description

Currently, in pod-failure, we use a non-existing image (fake image) to update the current container image. In this way, we can make the pod unusable. 

But at the same time, there is a problem with this approach.  The restart time will accumulate to infinity. So we should use other methods. For example,  We can use `k8s.gcr.io/pause` or an empty image as a fake image.

## Score

50

## Mentor(s)

* @YangKeao 

Contact the mentors: #tidb-challenge-program channel in TiDB Community Slack Workspace

## Recommended Skills

* Golang
* Kubernetes

## Learning Materials

AngleNet:
Fix #189 
The `pause` image will always fail to start. 
### Check List 

Tests 

Code changes

 - Has Go code change

Side effects

 - No

Related changes

```release-note
Use the `pause` image to inject pod-failure chaos
```

### DIFF
--- a/controllers/podchaos/podfailure/types.go
+++ b/controllers/podchaos/podfailure/types.go
@@ -38,8 +38,9 @@ import (
 )
 
 const (
-	// fakeImage is a not-existing image.
-	fakeImage = "pingcap.com/fake-chaos-mesh:latest"
+
+	// Always fails a container
+	pauseImage = "gcr.io/google-containers/pause:latest"
 
 	podFailureActionMsg = "pause pod duration %s"
 )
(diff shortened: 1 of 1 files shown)


zhouqiang-cl:
@YangKeao seems CI fail https://internal.pingcap.net/idc-jenkins/blue/organizations/jenkins/chaos-mesh-test/detail/chaos-mesh-test/1028/pipeline

PR #326, opened 2020-03-09
Edit chaos-daemon DaemonSet to allow Tolerations

tfulcrand:

In some Kubernetes configuration chaos-daemon DaemonSet default behaviour is not allowed to create pod on each kubernetes node.

Being able to use tolerations for chaos-daemon in chart values.yaml.

```release-note
Allow users to configure tolerations for chaos-daemon.
```

### DIFF
--- a/helm/chaos-mesh/values.yaml
+++ b/helm/chaos-mesh/values.yaml
@@ -70,6 +70,8 @@ chaosDaemon:
   # runtime: containerd
   # socketPath: /run/containerd/containerd.sock
 
+  tolerations: []
+
 dashboard:
   create: false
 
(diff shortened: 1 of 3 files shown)


helm/chaos-mesh/values.yaml
```diff
@@ -70,6 +70,8 @@ chaosDaemon:
   # runtime: containerd
   # socketPath: /run/containerd/containerd.sock
 
+  tolerations: []
```
cwen0:
```
      - key: node-role.kubernetes.io/master
        effect: NoSchedule
``` 

How about setting a default value?

tfulcrand: We have never observed any problems with master nodes (chaos-daemon pod created well on them) but yes we can add a default value.

YangKeao: @cwen0 I prefer not to set default value actually. 

Because we have no place to tell/warn user that our chaos-daemon won't follow your guide to place pods.

Issue #145, opened 2020-01-17
UCP: Add memory chaos


zhouqiang-cl:
## Description

Sometimes we should simulate the situation we can not alloc memory. So we need a kind of "chaos" to simulate this situation.

## Score

1679.5

## Mentor(s)

@ethercflow 

## Recommended Skills

* Golang
* Kubernetes

## Learning Materials

* [Development guide of chaos-mesh](https://github.com/pingcap/chaos-mesh/wiki/Development-Overview)

AngleNet:
I'm very interested in this chaos. Any implementation hint or potential solution about this ?

zhouqiang-cl:
@AngleNet sorry for the late update. https://github.com/pingcap/chaos-mesh/pull/273 this pr can do this by bpf

zhouqiang-cl:
@AngleNet My fault. the bpf inject from kernel and works on slab/alloc/etc. but it not simulate lack of memory
If simulate lack of memory. I think you can inject to the pod and eat some memory of the pod

zhouqiang-cl:
And stress-ng may be very helpful to do the memory chaos

AngleNet:
Thank you. I will check that. I'll try to do it.

zhouqiang-cl:
@AngleNet Maybe this simple stress is also very good https://github.com/vishh/stress/blob/master/main.go

AngleNet:
Sorry for the delay. I already checked that. The idea is to eat memory out to get memory allocation failures.

AngleNet:
I think the memory chaos should try to do the following jobs:
1. eats a fixed size of memory of the entire pod to inject the chaos. 
2. free the chaos meory to recover it.
If I'm wrong, please corret me.  @zhouqiang-cl @cwen0

zhouqiang-cl:
> I think the memory chaos should try to do the following jobs:
> 
> 1. eats a fixed size of the memory of the entire pod to inject the chaos.
> 2. free the chaos memory to recover it.
> 
> If I'm wrong, please correct me. @zhouqiang-cl @cwen0

Yes. exactly

AngleNet:
OK. I'm working on it along with #189

zhouqiang-cl:
You can pick it such as run `/pick-up-challenge`

Issue #144, opened 2020-01-17
UCP: Add CPU chaos


zhouqiang-cl:
## Description

In an actual production environment, the CPU may be busy. So there needs to be a way to simulate a busy CPU situation.  Sometimes we need to look at the performance of the server's performance when the CPU is busy. 

So we need to add a new chaos type, namely "CPU chaos". Maybe we can you stress-ng to simulate CPU is busy.

## Score

1679.5

## Mentor(s)

@ethercflow 

## Recommended Skills

* Golang
* stress-ng

## Learning Materials

* [Development guide of chaos-mesh](https://github.com/pingcap/chaos-mesh/wiki/Development-Overview)

huangwei2013:
Found what chaos-cpu/memory support 

> memory:
    --vm 
    --vm-keep
    --vm-bytes
cpu:
    --cpu
    --cpu-load


What if expect more function of stress-ng？ Is there someway to specify more stress-ng options？

cwen0:
@huangwei2013 You can define plenty of stressors by `StressngStresssors`  field, https://github.com/pingcap/chaos-mesh/blob/master/api/v1alpha1/stresschaos_types.go#L63-L70

huangwei2013:
@cwen0  I have read the code and understand this solution. Just to confirm is there some “cheap way” to make use of more stress-ng options

cwen0:
@huangwei2013 Sorry, at present, we only support use this way to make use of more stress-ng options.

huangwei2013:
@cwen0  what about make extend as this format：

>   options: ["--cpu 2", "--timeout 600", "--hdd 1"]

Of course it is no  strict enough but more powerful

cwen0:
@AngleNet How do you think?

AngleNet:
@cwen0 Sorry for the late reply.  I think defining another options for each stressor does not have big advantages over defining one `StressngStressors`. We should extend stressor spec with more powerful features via our own definition of stressor not from `stress-ng`.

PR #332, opened 2020-03-11
 Add a new `StressChaos` to generate plenty of stresses

AngleNet:
Add a new `StressChaos` to generate plenty of stresses over a collection of pods. We can burn cpus #144 , eat memory #145 and so on.
Inject stresses into container's `cgroup`
### Check List 

Tests 

 - Manul tests

Code changes

Side effects

 - NO

Related changes

```release-note
Add a new `StressChaos` to generate plenty of stresses
```


zhouqiang-cl:
well done

AngleNet:
I'm very pleased to say that it's ready to be reviewed now ! Please help to take a look. @zhouqiang-cl @ethercflow :heart:

AngleNet:
Still missing unit tests for it since I'm not familiar with the current testing framework. Maybe I could add them later in another PR ?

cwen0:
@AngleNet Good job!

api/v1alpha1/stresschaos_webhook.go
```diff
@@ -0,0 +1,39 @@
...
+
+package v1alpha1
+
+import (
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+// log is for logging in this package.
+var stressChaosLog = ctrl.Log.WithName("podchaos-resource")
```
zhouqiang-cl: s/podchaos-resource/stresschaos?

cmd/chaos-stress/main.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+var (
+	addr         string
+	printVersion bool
+	log          = ctrl.Log.WithName("chaos-stress")
+)
+
+func initFlag() {
+	flag.StringVar(&addr, "addr", ":65533", "RPC server address")
```
zhouqiang-cl: Do we need to add chaos-stress to chaos-daemon.  @cwen0 @YangKeao @Yisaer what about your opinion

api/v1alpha1/stresschaos_webhook.go
AngleNet: Misspelled, will fix it

cmd/chaos-stress/main.go
YangKeao: No. `chaos-stress` is a sidecar container, but not a daemon.

YangKeao:
Maybe we could integrate this function into `chaos-daemon` but not a sidecar container.

We could find related cgroup path in `/proc/[pid]/cgroup`, and then use `cgexec` to start a process in a cgroup. But unfortunately alpine doesn't have `libcg` pkg and doesn't provide `cgexec` command. You could use this go [package](https://github.com/containerd/cgroups) as an alternative.

Without sidecar, it will be much easier to use :rocket:

AngleNet:
Make sense ! Will migrate it into chaos-daemon

cwen0:
@AngleNet Hi, Currently, we are preparing Chaos Mesh-for CNCF Sandbox entry. https://github.com/cncf/toc/pull/367. At present, we are preparing the slide which requires some contributor-related information.  Would you mind tell us what company are you from?

AngleNet:
Help ! How could I get a pod's cgroup path ? Is there any one know how to do this over docker and containerd runtime ? @YangKeao @zhouqiang-cl

YangKeao:
@AngleNet I have a draft roadmap, but not 100% sure it can work.

* `GetPidFromContainerID` in `/pkg/chaosdaemon/util.go` can help you get pid from a container id in chaos-daemon. (You can check other functions in chaos-daemon)

* Get cgroup path in `/proc/[pid]/cgroup`.

Then you can find the real cgroup path in `/sys/fs/cgroup `. Notice that the `/sys` in chaos-daemon is different from host `/sys`. You can mount it directly.

AngleNet:
Thank you very much !  ! Seems mounts to /sys is ok. I will try it in that way.

AngleNet:
I finally migrate it to chaos-daemon now ! Please help to take a look.

Gallardot:
@AngleNet Good job! 

By the way, We have added a new validator for chaos object.  For example 
https://github.com/pingcap/chaos-mesh/blob/master/api/v1alpha1/timechaos_webhook.go#L47-L95

`StressChaos` should also implementing the `ChaosValidator`

AngleNet:
OK. Will add it for `StressChaos`. Thank you for point it out.

Gallardot:
@AngleNet We should add `StressChaos ` to `webhook.CRDS`.

https://github.com/pingcap/chaos-mesh/blob/1c6a077335d19b9c412b4f6cda0236283882e0a2/helm/chaos-mesh/values.yaml#L171-L176

pkg/chaosdaemon/stress_server.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/containerd/cgroups"
+	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/pkg/errors"
+
+	pb "github.com/pingcap/chaos-mesh/pkg/chaosdaemon/pb"
+)
+
+var (
+	stressorLocker = new(sync.Mutex)
+	podStressors   = make(map[string]*exec.Cmd)
```
YangKeao: Save pid list here is not safe. If `chaos-daemon` restarts, we can never stop the chaos action. Why not use `-t` argument for `stress-ng` and do nothing when `Recover`? (Though this is also little dangerous.)

api/v1alpha1/stresschaos_types.go
```diff
@@ -0,0 +1,180 @@
...
+	// Selector is used to select pods that are used to inject chaos action.
+	Selector SelectorSpec `json:"selector"`
+
+	// Stressors defines plenty of stressors to be generated. It's simply a few
+	// arguments for stress-ng. For fully supported stressors, see stress-ng(1)
+	Stressors string `json:"stressors"`
+
+	// Duration represents the duration of the chaos action
+	// +optional
+	Duration *string `json:"duration"`
```
YangKeao: Please add `omitempty` for `Duration` and `Scheduler`

YangKeao:
Awesome PR. The effect is very significant! 🍻

YangKeao:
Use `stressor` (`stress-ng`'s argument) as a field is a little difficult to understand. I think at least we should turn it into a `map[string]string` form.

pkg/chaosdaemon/stress_server.go
AngleNet: Yes. Seems to be better to wrapps all of arguments of `stress-ng` with our own type of argument as you said below. I eases usiblity and isolates the underlying stressing generator. However, the power of `stress-ng` may be limited by our own type. I think it's a tradeoff.  What do you think ?

api/v1alpha1/stresschaos_types.go
AngleNet: Thank you.  Will add it.

AngleNet:
> Use `stressor` (`stress-ng`'s argument) as a field is a little difficult to understand. I think at least we should turn it into a `map[string]string` form.

Did you mean we turn `--cpu 8 -t 5s` to `{ "--cpu": '8', "-t": '5s'}` ?

pkg/chaosdaemon/stress_server.go
AngleNet: Yes. It's safer than now. We could convert user specified `duration` into `-t`.

AngleNet: Maybe we could wrap arguments of `stress-ng` with our own arguments ?

YangKeao:
@AngleNet I prefer `{"cpu":"8"}`. But this will fail if the user specifies the arguments abbreviation. @cwen0 What's your opinion?

pkg/chaosdaemon/stress_server.go
AngleNet: Howabout we save pid here and keep it in `Status`?  When the chaos daemon restarts we could always cancel it. I think it's more reliable.

AngleNet:
@YangKeao Maybe we should keep it that way to make it easy for users to learn and edit the chaos ? However it binds our `StressChaos` onto `stress-ng`. That is awful for changing to a new stressing framework in futher. But I think `stress-ng` is enough.

pkg/chaosdaemon/stress_server.go
YangKeao: Great 👍

YangKeao:
@AngleNet Yes! It seems to be a bad idea to convert arguments to map (because they are both hard to understand without a document). 

Do we need to expose the implementation detail to provide customizability? What is the difference between calculating FFT, log, and matrix? Should the user care how the tons of memory was allocated? Through `calloc` or `brk`?

However, there are still some valuable options. Not only memory, cpu, but there are also vm stresses, interrupts, and page faults. 

It's always a dilemma to provide a high customizable or a friendly interface. Should we provide the users with several limited options so that they can easily learn how to use right after reading the example, or should we provide them a strongly customizable field (such as arguments for `stress-ng`, which can use the full power of it), but expect users to learn how to use an external tool (stress-ng).

For example, a highly simplified version can be:

```
stress:
- memory
- cpu
- page faults
```

Then the stress chaos will allocate tons of memory, squeeze out the CPU and generate a lot of page faults. Is it too limited?

@cwen0 I am quite indecisive actually. What's your opinion?

pkg/chaosdaemon/stress_server.go
```diff
@@ -0,0 +1,223 @@
...
+			log.Error(err, "fail to exit stressors", "pid", cmd.Process.Pid)
+			return nil, err
+		}
+	}
+	return &empty.Empty{}, nil
+}
+
+func findValidCgroup(path cgroups.Path, target string) (string, error) {
+	for _, subsys := range cgroupSubsys {
+		if p, _ := path(cgroups.Name(subsys)); strings.Contains(p, target) {
```
cwen0: Why ignore the error here?

cwen0:
> @AngleNet Yes! It seems to be a bad idea to convert arguments to map (because they are both hard to understand without a document).
> 
> Do we need to expose the implementation detail to provide customizability? What is the difference between calculating FFT, log, and matrix? Should the user care how the tons of memory was allocated? Through `calloc` or `brk`? (There may be different behaviors. I don't know 😢 )
> 
> However, there are still some valuable options. Not only memory, cpu, but there are also vm stresses, interrupts, and page faults.
> 
> It's always a dilemma to provide a high customizable or a friendly interface. Should we provide the users with several limited options so that they can easily learn how to use right after reading the example, or should we provide them a strongly customizable field (such as arguments for `stress-ng`, which can use the full power of it), but expect users to learn how to use an external tool (stress-ng).
> 
> For example, a highly simplified version can be:
> 
> ```
> stress:
> - memory
> - cpu
> - page faults
> ```
> 
> Then the stress chaos will allocate tons of memory, squeeze out the CPU and generate a lot of page faults. Is it too limited?
> 
> @cwen0 I am quite indecisive actually. What's your opinion?

I agree with this opinion from @YangKeao，we can't require users to learn `stress-ng` before using stressChaos.

pkg/chaosdaemon/stress_server.go
AngleNet: Nice catch ! If we delete the container without delete the chaos, I think the stressing instance is still there. If we delete the pod whithout delete the chaos, I'm not sure about it. I'll check it. Did you mean we should eliminate the stressing instance once the container or pod is destroyed ? @cwen0

api/v1alpha1/stresschaos_types.go
```diff
@@ -0,0 +1,180 @@
...
+	// If `RandomMaxPercentPodMod`,  provide a number from 0-100 to specify the % of pods to do chaos action
+	// +optional
+	Value string `json:"value"`
+
+	// Selector is used to select pods that are used to inject chaos action.
+	Selector SelectorSpec `json:"selector"`
+
+	// Stressors defines plenty of stressors to be generated. It's simply a few
+	// arguments for stress-ng. For fully supported stressors, see stress-ng(1)
+	Stressors string `json:"stressors"`
```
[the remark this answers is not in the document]
AngleNet: Yes ! I will refine it according to our discussion abount how we expose these stuff to users and then add an example for it.

api/v1alpha1/stresschaos_types.go
```diff
@@ -0,0 +1,180 @@
...
+
+// GetValue is a getter for Value (for implementing SelectSpec)
+func (in *StressChaosSpec) GetValue() string {
+	return in.Value
+}
+
+// StressChaosStatus defines the observed state of StressChaos
+type StressChaosStatus struct {
+	ChaosStatus `json:",inline"`
+	Target      string `json:"uuid"`
```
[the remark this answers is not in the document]
AngleNet: You are right. I think it's a bug here. I will fix it.

api/v1alpha1/stresschaos_webhook.go
```diff
@@ -0,0 +1,103 @@
...
+	if strings.TrimSpace(in.Spec.Stressors) == "" {
+		errList = append(errList, field.Invalid(
+			specField.Child("stressors"), in.Spec.Stressors,
+			"stressors should always be set"))
+	}
+	if len(errList) > 0 {
+		return fmt.Errorf(errList.ToAggregate().Error())
+	}
+	return nil
+
```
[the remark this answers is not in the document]
AngleNet: OK

controllers/stresschaos/types.go
```diff
@@ -0,0 +1,239 @@
...
+	daemonClient, err := utils.NewChaosDaemonClient(ctx, r.Client,
+		pod, os.Getenv("CHAOS_DAEMON_PORT"))
+	if err != nil {
+		return err
+	}
+	defer daemonClient.Close()
+	if len(pod.Status.ContainerStatuses) == 0 {
+		return fmt.Errorf("%s %s can't get the state of container", pod.Namespace, pod.Name)
+	}
+	target := pod.Status.ContainerStatuses[0].ContainerID
```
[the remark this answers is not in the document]
AngleNet: I think it only supports to inject stressing chaos into pods for now. Plans to support container in later PRs. Is that an option ? @cwen0

doc/stress_chaos.md
```diff
@@ -0,0 +1,41 @@
...
+
+This document helps you to build stress chaos experiments.
+
+Stress chaos is a chaos to generate plenty of stresses over a collection of pods. A sidecar will be injected along with the target pod during creating. It's the sidecar which generates stresses or cancels them. For now, we use `stress-ng` as the stress generator for the chaos.
+
+## Usage 
+
+### Define a `StressChaos`
+
+An example `yaml` of `StressChaos` which burns 1 CPU for 30 seconds in every 2 minutes is listed below. You can specify all of `stress-ng`  stressors via the `stressors` field, see `stress-ng(1)` for supported stressors. The others `spec` fields are common to all of chaos.
```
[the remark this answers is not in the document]
AngleNet: OK

pkg/chaosdaemon/stress_server.go
```diff
@@ -0,0 +1,223 @@
...
+
+	// Possible cgroup subsystems
+	cgroupSubsys = []string{"cpu", "memory", "systemd", "net_cls",
+		"net_prio", "freezer", "blkio", "perf_event", "devices",
+		"cpuset", "cpuacct", "pids", "hugetlb"}
+)
+
+func (s *daemonServer) ExecStressors(ctx context.Context,
+	req *pb.StressRequest) (*empty.Empty, error) {
+	log.Info("executing stressors", "request", req)
```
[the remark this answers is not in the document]
AngleNet: OK

pkg/chaosdaemon/stress_server.go
```diff
@@ -0,0 +1,223 @@
...
+	pid, err := s.crClient.GetPidFromContainerID(ctx, req.Target)
+	if err != nil {
+		return nil, err
+	}
+	path := PidPath(int(pid))
+	id, err := s.crClient.StripContainerIDProtocolPrefix(ctx, req.Target)
```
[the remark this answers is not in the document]
AngleNet: Yes

pkg/chaosdaemon/stress_server.go
AngleNet: I think we only need to find one cgroup path here. The `subsys` tried could not exist. So I think it's safe to ignore the error.

AngleNet:
@YangKeao @cwen0 Great ! I will define our own stressors based on `stress-ng`. Any suggestions on  the stressor definitions ?

pkg/chaosdaemon/stress_server.go
cwen0: If the stress action takes no effect on other pods when the container or pod was deleted, it is unnecessary to eliminate the stressing instance.

controllers/stresschaos/types.go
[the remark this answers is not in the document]
cwen0: It is ok

pkg/chaosdaemon/stress_server.go
cwen0: Got

pkg/chaosdaemon/stress_server.go
YangKeao: @AngleNet As I have tested on minikube, when the pod is removed, the `stress-ng` process is also removed. But I am not clear about this mechanism.

Yisaer:
IMO, I think it's ok to expose configuration for `stress-ng` to users for customization while the configuration still could be omitted with a default value to provide a `out-of-box` user experience.
FYI @AngleNet

AngleNet:
Thank you for the suggestion. I will consider it as an option

AngleNet:
Just mark it active ~~

zhouqiang-cl:
😄

pkg/chaosdaemon/stress_server.go
Yisaer: I suggest not to ignore error here for any reason.

AngleNet: @Yisaer I don't get it. Could you explain this more specificially ?

cwen0:
@AngleNet Is there any progress on this PR?

AngleNet:
@zhouqiang-cl It's finally ready for review again. Sorry for the late submit. Hope it will catch the release. Could you please help to review this?  Sorry again !

pkg/chaosdaemon/stress_server.go
```diff
@@ -0,0 +1,221 @@
...
+func (s *daemonServer) CancelStressors(ctx context.Context,
+	req *pb.StressRequest) (*empty.Empty, error) {
+	pid, err := strconv.Atoi(req.Target)
+	if err != nil {
+		return nil, err
+	}
+	log.Info("Canceling stressors", "request", req)
+	process, err := os.FindProcess(pid)
+	if err == nil {
+		process.Kill()
```
cwen0, line 102: handle this error

api/v1alpha1/stresschaos_webhook.go
```diff
@@ -0,0 +1,188 @@
...
+	return ValidatePodMode(in.Spec.Value, in.Spec.Mode, spec.Child("value"))
+}
+
+// ValidateScheduler validates whether scheduler is well defined
+func (in *StressChaos) ValidateScheduler(root *field.Path) field.ErrorList {
+	panic("implement me")
+}
+
+// Validate validates the scheduler and duration
+func (in *StressChaosSpec) Validate(parent *field.Path, errs field.ErrorList) field.ErrorList {
```
Gallardot, line 104:
```suggestion
func (in *StressChaosSpec) Validate(parent *field.Path) field.ErrorList {
```

Why not keep the same style?

api/v1alpha1/stresschaos_types.go
```diff
@@ -0,0 +1,263 @@
...
+
+// VMStressor defines how to stress memory out
+type VMStressor struct {
+	Stressor `json:",inline"`
+
+	// Bytes specifies N bytes consumed per vm worker, default is the total available memory.
+	// One can specify the size as % of total available memory or in units of B, KB/KiB,
+	// MB/MiB, GB/GiB, TB/TiB.
+	// +optional
+	Bytes string `json:"bytes,omitempty"`
```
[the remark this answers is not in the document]
AngleNet: I think it's better to comply to SI standard. https://en.wikipedia.org/wiki/Binary_prefix

api/v1alpha1/stresschaos_webhook_test.go
```diff
@@ -0,0 +1,207 @@
...
+			vm := VMStressor{}
+			incorrectBytes := []string{"-1", "-1%", "101%", "x%", "-1Kb"}
+			for _, b := range incorrectBytes {
+				vm.Bytes = b
+				Expect(vm.tryParseBytes()).Should(HaveOccurred())
+			}
+			correctBytes := []string{"", "1%", "100KB", "100B"}
+			for _, b := range correctBytes {
+				vm.Bytes = b
+				Expect(vm.tryParseBytes()).ShouldNot(HaveOccurred())
```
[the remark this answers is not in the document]
AngleNet: I think it's already been checked.

api/v1alpha1/stresschaos_webhook.go
```diff
@@ -0,0 +1,188 @@
...
+	once := false
+	if in.VMStressor != nil {
+		errs = in.VMStressor.Validate(current, errs)
+		once = true
+	}
+	if in.CPUStressor != nil {
+		errs = in.CPUStressor.Validate(current, errs)
+		once = true
+	}
+	if !once {
```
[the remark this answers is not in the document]
AngleNet: I don't get it. I only want to fire an error when no stressors specified.

api/v1alpha1/stresschaos_webhook.go
AngleNet: Fixed

api/v1alpha1/stresschaos_webhook.go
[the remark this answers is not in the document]
cwen0: Yes, If the `errs` from `in.VMStressor.Validate(current, errs)` is not nil, but this `err
s` maybe overwrite by the `in.CPUStressor.Validate(current, errs)`, in this case, we will lose the errors from `in.VMStressor.Validate(current, errs)`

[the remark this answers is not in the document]
cwen0: Sorry, I misunderstood the logic of this place.

pkg/chaosdaemon/util.go
```diff
@@ -51,6 +51,7 @@ const (
 type ContainerRuntimeInfoClient interface {
 	GetPidFromContainerID(ctx context.Context, containerID string) (uint32, error)
 	ContainerKillByContainerID(ctx context.Context, containerID string) error
+	FormatContainerID(ctx context.Context, containerID string) (string, error)
```
yeya24, line 54: I don't really think this method should be part of the interface.

controllers/stresschaos/types.go
```diff
@@ -0,0 +1,254 @@
...
+
+	stresschaos.Status.Experiment.Pods = []v1alpha1.PodStatus{}
+
+	for _, pod := range pods {
+		ps := v1alpha1.PodStatus{
+			Namespace: pod.Namespace,
+			Name:      pod.Name,
+			HostIP:    pod.Status.HostIP,
+			PodIP:     pod.Status.PodIP,
+			Message:   fmt.Sprintf(stressChaosMsg, stresschaos.Spec.Stressors),
```
yeya24, line 110: `stresschaos.Spec.Stressors` is a pointer, so the result looks like below. Please use a string function here.

``` yaml
  status:
    experiment:
      phase: Running
      podChaos:
      - action: ""
        hostIP: 172.17.0.4
        message: 'stressors: &{<nil> 0xc0001786e0}'
        name: etcd-7db9d6f6f9-lgr6k
        namespace: chaos-testing
        podIP: 10.244.1.8
      startTime: "2020-04-24T20:34:53Z"
```

yeya24:
Seems this doesn't support `apply` semantics? 

For example, if first applying a memory stress with 10MB bytes. Then update the stress yaml to 100MB, and apply again. Seems that 100MB stress is not executed until the nextStart. 

But actually what we need is to replace the current execution to 100MB.

api/v1alpha1/stresschaos_types.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+// StressInstance is an instance generates stresses
+type StressInstance struct {
+	// UID is the instance identifier
+	// +optional
+	UID string `json:"uid"`
+	// StartTime specifies when the instance starts in unix timestamp
+	// +optional
+	StartTime int64 `json:"startTime"`
```
yeya24, line 125: Is it better to use a specific format, instead of just timestamp?

api/v1alpha1/stresschaos_types.go
```diff
@@ -0,0 +1,275 @@
...
+func (in *StressChaosSpec) GetValue() string {
+	return in.Value
+}
+
+// StressChaosStatus defines the observed state of StressChaos
+type StressChaosStatus struct {
+	ChaosStatus `json:",inline"`
+	// Instances always specifies stressing instances
+	// +optional
+	Instances map[types.UID]StressInstance `json:"instances,omitempty"`
```
yeya24, line 115: I hope we can use a formatted string as the key, such as `namespace/podname`, instead of just the uid of the pod. That should be more readable. The current is like below, `91cb8ac8-a123-4568-a180-80734d3533fd` doesn't mean anything to user

``` yaml
    instances:
      91cb8ac8-a123-4568-a180-80734d3533fd:
        startTime: 1587763257000
        uid: "7346"
```

pkg/chaosdaemon/stress_server.go
```diff
@@ -0,0 +1,241 @@
...
+func errorPath(err error) cgroups.Path {
+	return func(_ cgroups.Name) (string, error) {
+		return "", err
+	}
+}
+
+func existingPath(paths map[string]string, pid int, suffix string) cgroups.Path {
+	// localize the paths based on the root mount dest for nested cgroups
+	for n, p := range paths {
+		dest, err := getCgroupDestination(pid, string(n))
```
yeya24, line 159: Why not just read `/proc/pid/mountinfo` out of the loop, then just check `n` inside the loop? WDYT @AngleNet ?

doc/stress_chaos.md
```diff
@@ -0,0 +1,69 @@
...
+```bash
+# If you do not have a namespace for your application, you can create one such as tidb-cluster-demo
+kubectl create ns tidb-cluster-demo
+# Create your pods in the target namespace such as tidb-cluster-demo
+kubectl apply -f your-pods.yaml
+# Inject a stress chaos
+kubectl apply -f your-stress-chaos.yaml
+# Your pod's cpu will burn for 30s
+```
+
```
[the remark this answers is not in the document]
AngleNet: @yeya24 Since the injected stressing process is not in the same namespace with the container, we could not see it directly. I tested it locally via starting a busy process while stress cpu. Its cpu drops dramatically.

AngleNet:
> Seems this doesn't support `apply` semantics?
> 
> For example, if first applying a memory stress with 10MB bytes. Then update the stress yaml to 100MB, and apply again. Seems that 100MB stress is not executed until the nextStart.
> 
> But actually what we need is to replace the current execution to 100MB.

I think it did not support `apply`.

pkg/chaosdaemon/stress_server.go
AngleNet: I think that also works. I just steal the code from `github.com/containerd/cgroups` package since the official version could not find mountinfo correctly. Hope it will fixed and we could remove the entire stealling pieces and use the official one.

pkg/chaosdaemon/stress_server.go
Yisaer: The error may contains the situation which we haven't considered, it is dangerous to ignore error unless for a strong reason. I didn't see it here.

pkg/chaosdaemon/stress_server.go
```diff
@@ -0,0 +1,241 @@
...
+	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/pkg/errors"
+	"github.com/shirou/gopsutil/process"
+
+	pb "github.com/pingcap/chaos-mesh/pkg/chaosdaemon/pb"
+)
+
+var (
+	stressorLocker = new(sync.Mutex)
+	podStressors   = make(map[string]*exec.Cmd)
```
Yisaer, line 39: Are these variables used now? If not, plz remove.

pkg/chaosdaemon/stress_server.go
AngleNet: I think if we never find a match due to whatever happened, we finally give an error. From that point of view, I think that is ok.

Yisaer: Though we would finally get an error, but we would miss the root error which we ignored here. I don't think this is a good reason to ignore error here.

Yisaer:
> > Seems this doesn't support `apply` semantics?
> > For example, if first applying a memory stress with 10MB bytes. Then update the stress yaml to 100MB, and apply again. Seems that 100MB stress is not executed until the nextStart.
> > But actually what we need is to replace the current execution to 100MB.
> 
> I think it did not support `apply`.

When we update the exited stress object from 10MB memory stress to 100MB stress, would the previous stress process be replaced to the new sress process. Or there would be 2 stress process ?One for 10MB, another for 100MB.

AngleNet:
It will be replaced to the newest config.

AngleNet:
> Almost LGTM

I think if we never find a valid cgroup inside `findValidCgroup` due to whatever happened, it finally gives us an error. I think that's ok.

pkg/chaosdaemon/util.go
yeya24: WDYT? @AngleNet I think this function shouldn't be in part of the interface.

yeya24, line 54: This is just a helper function, which shouldn't be part of the interface. Also it is too easy to be a function...

yeya24:
Have you ever tested this feature in contained environment, such as `Kind`? Seems it doesn't work, both CPU stress and Memory stress.

```
➜  examples git:(feature/stress) ✗ kubectl top pods -l app=etcd
NAME                    CPU(cores)   MEMORY(bytes)
etcd-7db9d6f6f9-2tt6m   6m           4Mi

➜  examples git:(feature/stress) ✗ cat burn-memory.yaml
kind: StressChaos
metadata:
  name: burn-memory
  namespace: chaos-testing
spec:
  mode: one
  selector:
    labelSelectors:
      app: etcd
  stressors:
    memory:
      workers: 1
      size: 100MB
  duration: "20m"
  scheduler:
    cron: "@every 30m"

➜  examples git:(feature/stress) ✗ klo chaos-daemon-ht9st
Chaos-daemon Version: version.Info{GitVersion:"v0.0.0-master+$Format:%h$", GitCommit:"d893703a7602ebbf8a5bc3dba0aab0de3b7e8cea", GitTreeState:"dirty", BuildDate:"2020-04-24T21:35:38Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
2020-04-25T14:24:04.132Z	INFO	chaos-daemon-server	Starting http endpoint	{"address": "0.0.0.0:31766"}
2020-04-25T14:24:04.134Z	INFO	chaos-daemon-server	Starting grpc endpoint	{"address": "0.0.0.0:31767", "runtime": "containerd"}
2020-04-25T14:31:13.881Z	INFO	chaos-daemon-server	Executing stressors	{"request": "scope:POD target:\"containerd://236bb108ef4070579b843562b3999d4dea188cce86a17233b3fb7bf9d322a4aa\" stressors:\" --vm 1 --vm-keep --vm-bytes 100000000\" "}
```

AngleNet:
What did you mean it didn’t work? I see the stressors are injected

yeya24:
Yes, it is injected. But the memory/CPU is not increasing at all

cwen0:
> Yes, it is injected. But the memory/CPU is not increasing at all. They remain the same. Is it the expected behavior? I guess no?
> 
> Or maybe the measuring method is not correct. `kubectl top pods` cannot get the correct result. So is there any other ways to validate there is a CPU/memory load. How did you do that? It would be helpful if you could provide some guides in the document.

I also run stressChaos in my environment, this is work for me, but we can't find this `stress` process in containers, I guess that we should run this `stress-ng`  in the target namespace.

api/v1alpha1/stresschaos_webhook.go
```diff
@@ -0,0 +1,182 @@
...
+	if len(in.StressngStressors) == 0 && in.Stressors == nil {
+		errs = append(errs, field.Invalid(current, in, "missing stressors"))
+	} else if in.Stressors != nil {
+		errs = in.Stressors.Validate(current, errs)
+	}
+	return append(errs, ValidateScheduler(in.Duration, in.Scheduler, current)...)
+}
+
+// Validate validates whether the Stressors are all well defined
+func (in *Stressors) Validate(parent *field.Path, errs field.ErrorList) field.ErrorList {
```
Gallardot, line 115:
```suggestion
func (in *Stressors) Validate(parent *field.Path) field.ErrorList {
```

keep the same style

api/v1alpha1/stresschaos_webhook.go
```diff
@@ -0,0 +1,182 @@
...
+		once = true
+	}
+	if !once {
+		errs = append(errs, field.Invalid(current, in, "missing stressors"))
+	}
+	return errs
+}
+
+// Validate validates whether the Stressor is well defined
+func (in *Stressor) Validate(parent *field.Path, errs field.ErrorList) field.ErrorList {
```
Gallardot, line 133:
```suggestion
func (in *Stressor) Validate(parent *field.Path) field.ErrorList {
```

api/v1alpha1/stresschaos_webhook.go
```diff
@@ -0,0 +1,182 @@
...
+	if in.Workers <= 0 {
+		errs = append(errs, field.Invalid(parent, in, "workers should always be positive"))
+	}
+	return errs
+}
+
+// Validate validates whether the MemoryStressor is well defined
+func (in *MemoryStressor) Validate(parent *field.Path, errs field.ErrorList) field.ErrorList {
```
Gallardot, line 141:
```suggestion
func (in *MemoryStressor) Validate(parent *field.Path) field.ErrorList {
```

api/v1alpha1/stresschaos_webhook.go
```diff
@@ -0,0 +1,182 @@
...
+		if err != nil {
+			return err
+		}
+		in.Size = fmt.Sprintf("%db", size)
+	}
+	return nil
+}
+
+// Validate validates whether the CPUStressor is well defined
+func (in *CPUStressor) Validate(parent *field.Path, errs field.ErrorList) field.ErrorList {
```
Gallardot, line 175:
```suggestion
func (in *CPUStressor) Validate(parent *field.Path) field.ErrorList {
```

pkg/chaosdaemon/util.go
AngleNet: Thank you for the review. Yes. It’s simple. However, I think it might be runtime related. So I put it into the interface. WDYT?

yeya24: It is runtime related, but it doesn't mean this should be a method in the interface, actually this function is too simple to be a function... Why do we need that? What's the problem of just using `containerID[len(dockerProtocolPrefix):]`? I think this is fine, we can totally remove the prefix check below
```
	if len(containerID) < len(dockerProtocolPrefix) {
		return "", fmt.Errorf("container id %s is not a docker container id", containerID)
	}
	if containerID[0:len(dockerProtocolPrefix)] != dockerProtocolPrefix {
		return "", fmt.Errorf("expected %s but got %s", dockerProtocolPrefix, containerID[0:len(dockerProtocolPrefix)])
	}
```

cwen0: `FormatContainerID` also is used to verify the ContianerID.

api/v1alpha1/stresschaos_types.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+// Normalize the stressors to comply with stress-ng
+func (in *Stressors) Normalize() string {
+	stressors := ""
+	if in.MemoryStressor != nil {
+		stressors += fmt.Sprintf(" --vm %d --vm-keep", in.MemoryStressor.Workers)
+		if len(in.MemoryStressor.Size) != 0 {
+			if in.MemoryStressor.Size[len(in.MemoryStressor.Size)-1] != '%' {
+				size, _ := units.FromHumanSize(in.MemoryStressor.Size)
```
cwen0, line 218: Handle this error

pkg/chaosdaemon/stress_server.go
Yisaer: Almost LGTM. plz handle the error instead of ignoring the error.

PR #338, opened 2020-03-14
Refactoring validator

Gallardot:
#291

use `webhook.Validator` to refactoring validator
unity the form of webhook
add Unit test for validator

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

Side effects

 - 

Related changes

 - 

```release-note
NONE
```


cwen0:
@Gallardot Good job!

YangKeao:
It seems that I cannot delete a chaos resource. Do you know what the problem is ?

```
Error from server: admission webhook "vtimechaos.kb.io" denied the request: there is no content to decode
```

YangKeao:
It seems related with this [PR](https://github.com/kubernetes/kubernetes/pull/76346) on kubernetes, which means Validator will break compatibility with old version of kubernetes.

The former error occurs on kubernetes 1.14.8. As I have tested, 1.17.0 works well.

cwen0:
> It seems related with this [PR](https://github.com/kubernetes/kubernetes/pull/76346) on kubernetes, which means Validator will break compatibility with old version of kubernetes. There is also a related [issue](https://github.com/kubernetes-sigs/controller-runtime/issues/845) for controller-runtime.
> 
> The former error occurs on kubernetes 1.14.8. As I have tested, 1.17.0 works well.

About this issue, Can we disable validator by default? When the controller-runtime fix this https://github.com/kubernetes-sigs/controller-runtime/issues/845, we enable validator.

Gallardot:
> It seems related with this [PR](https://github.com/kubernetes/kubernetes/pull/76346) on kubernetes, which means Validator will break compatibility with old version of kubernetes. There is also a related [issue](https://github.com/kubernetes-sigs/controller-runtime/issues/845) for controller-runtime.
> 
> The former error occurs on kubernetes 1.14.8. As I have tested, 1.17.0 works well.

Thank you. I will test it on kubernetes 1.14.8 later. Maybe we can ignore the `DELETE` operations. I will have a try.

cwen0:
@Gallardot Hi, Currently, we are preparing Chaos Mesh-for CNCF Sandbox entry. https://github.com/cncf/toc/pull/367. At present, we are preparing the slide which requires some contributor-related information.  Would you mind tell us what organization are you from?

Gallardot:
@YangKeao @cwen0 

I reproduce this error on kubernetes 1.14.8. Currently, we do nothing for `DELETE` operations. Maybe we can ignore the `DELETE ` operations on the `ValidatingWebhookConfiguration`. It is Ok on kubernetes 1.14.8. Please take a look.

api/v1alpha1/common_webhook.go
```diff
@@ -13,9 +13,33 @@
...
 // DefaultNamespace set the namespace of chaos object as the default namespace selector if namespaces not set
 func (in *SelectorSpec) DefaultNamespace(namespace string) {
 	if len(in.Namespaces) == 0 {
 		in.Namespaces = []string{namespace}
 	}
 }
+
+// +kubebuilder:object:generate=false
+
+// ChaosValidator describe the interface should be implemented in chaos
```
cwen0: s/describe/describes/g

api/v1alpha1/common_webhook.go
```diff
@@ -13,9 +13,33 @@
...
+type ChaosValidator interface {
+	webhook.Validator
+	// Validate validate chaos object
```
cwen0: s/validate/validates/g

api/v1alpha1/common_webhook.go
```diff
@@ -13,9 +13,33 @@
...
+	Validate() error
+	// ValidateScheduler validate the scheduler and duration
```
cwen0: ditto, and please fix the same issue at the following code.

Gallardot: Thank you. All done.Please take a look.

PR #345, opened 2020-03-17
Extend `mod` to support `exact` pod to inject in case of misinjection

### LINKED ISSUE #179: UCP: Add protect policy to prevent the injected system from being unintentionally damaged
## Description

For example, after you inject ‘pod-kill’, the target pods are killed but they cannot return to the ‘Running’ state in time due to some reasons. Yet the pod-killing action continues. So all the pods might be under unhealthy status after a while and the overall system is somehow disrupted.

Add protection mechanism to avoid similar situations for pod-kill or other chaos crds: 
- block injection when the number of Running pod is below at a certain percentage rate.
- extend `selector`, e.g use regex to filter target pods.

Or we can extend `mode` methods, add `uniqueOne`.

## Score 

* 936.1

## Mentor(s)

* @fewdan 

Contact the mentors: **#tidb-challenge-program** channel in [TiDB Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace

## Recommended Skills

* Golang
* Kubernetes

## Learning Materials

AngleNet:
Fix #179 
Extend `mod` to support `exact` pod to inject in case of misinjection
### Check List 

Tests 

Code changes

 - Has Go code change

Side effects

 - No

Related changes

```release-note
NONE
```


YangKeao:
If I don't misunderstand your intention, `Pods` field in `SelectorSpec` can have the same effect.

AngleNet:
Yeah. I think the `exact` mod can prevent users from unintentional chaos injection. Or I misunderstand #179 . What do you think?  @mahjonp @zhouqiang-cl

fewdan:
My understanding is that we need to "protect" certain "pods" from being injected into chaos, rather than injecting chaos into specific "pods". So in my opinion, "exact mod" is not necessary.

AngleNet:
> My understanding is that we need to "protect" certain "pods" from being injected into chaos, rather than injecting chaos into specific "pods". So in my opinion, "exact mod" is not necessary.

Did you mean we should add something like 'blacklist`? Every pods that matches them should never been injected by the chaos.

fewdan:
> > My understanding is that we need to "protect" certain "pods" from being injected into chaos, rather than injecting chaos into specific "pods". So in my opinion, "exact mod" is not necessary.
> 
> Did you mean we should add something like `blacklist`? Every pods that matches them should never been injected by the chaos. Or any better ideas for how to do the `protection` ?

Yes, I think so. What do you think? @YangKeao

YangKeao:
@fewdan 

As pod name is not stable (with random postfix), blacklist of pod name isn't more useful than matchExpressions mechanism, which we also haven't implemented yet.

MatchExpressions will provide ability to exclude set of pods according to label.

@mahjonp What's your opinion?

AngleNet:
> @fewdan 
> 
> As pod name is not stable (with random postfix), blacklist of pod name isn't more useful than matchExpressions mechanism, which we also haven't implemented yet.
> 
> MatchExpressions will provide ability to exclude set of pods according to label.
> 
> @mahjonp What's your opinion?

Seems to be a better way !

fewdan:
In simple terms, does this mean we need a mechanism to protect pods based on "labels"?

cwen0:
> In simple terms, does this mean we need a mechanism to protect pods based on "labels"?

@AngleNet @YangKeao @fewdan  this mechanism is necessary, in addition, we also need to avoid  situations like described in the issue  #179

>For example, after you inject ‘pod-kill’, the target pods are killed but they cannot return to the ‘Running’ state in time due to some reasons. Yet the pod-killing action continues. So all the pods might be under unhealthy status after a while and the overall system is somehow disrupted.

AngleNet:
It turns out that we already have a `LabelSeletors` which used to find the **injected pods**. Now, we need another barely new label selector to filter out the **uninjected pods** ? I'm sorrry. I don't get it. Could you please give me more details ?

AngleNet:
Maybe we should replace the ‘LabelSelector’ with a matchExpression label selector?

cwen0:
In my opinion, we can think that the pods that meet the selector are all the pods that may be injected. Suppose we currently need to select a pod to perform the pod-kill action. At the same time, we find that many of the pods that meet the selection are unavailable. With state, if the kill pod is executed and the current application cluster is completely unavailable, I think we should give up this kill behavior at this time. The purpose of the #179 issue is to provide a rule of protection measures to protect the maximum scope of the cluster to be destroyed, to avoid chaos experiments of scheduler type due to multiple schedules causing the entire cluster to crash.  In addition, `matchExpression` can be used to protect the specified pods.  

To achieve this propose, we can define a field for each chaos object: 

```
type Protection struct {
          MatchExpression xxx // used to protect the specified pods.  Defining this field in labelSelector maybe better. 
          MaxInjectedPercent int 
}
``` 

If we set `MaxInjectedPercent` to 50% and the numbers of pods met the selectors is 3,  we should avoid more than two unhealthily pods.

Such as,  we use this chaos experiment 

```
xxxx
Kind: PodChaos 
Spec: 
   action: pod-kill
   selector: 
      "app": "etcd"
   scheduler
     cron: "@every 2m"
```

Suppose we have three etcd pods, first, we apply this chaos experiment, then one etcd pod will be killed and the etcd pod killed does not recover normally due to some reasons, After 2m, the scheduler will run this experiment again, at this time, the other etcd pod will be killed, in this situation, the etcd cluster may be crashed, because the etcd cluster only works if more than half of the nodes are normal.
If we set `MaxInjectedPercent` to 50  and the first pod killed did not return to normal.,  we can prevent the second pod from being killed.

AngleNet:
Great ! One more detail to discuss. How should we think a pod is unhealthy? I think we could see a pod as unhealthy whenever a pod's `Status` is **not** `Running`. When the ratio of healthy pods are below the `MaxInjectedPercent`, we could deny the injected chaos.

YangKeao:
For pod, it's quite simple. I think we should use `livenessProbe` and `readinessProbe`. But for a cluster, it's more difficult to discuss whether it's healthy.

cwen0:
> For pod, it's quite simple. I think we should use `livenessProbe` and `readinessProbe`. But for a cluster, it's more difficult to discuss whether it's healthy.

From my point of view, we don't need to consider the health of the cluster,  it should be considered by the user, and the user should set the protection rules. We only need to run according to the protection rules, and for the protection rules here, we only prevent that chaos experiments continue to run and make the cluster crash. To judge whether the pod is healthy, we could see a pod as unhealthy whenever a pod's Status is not `Running` or the chaos action was injected into this pod but not recovered.  

`livenessProbe` and `readinessProbe` are defined by the application pod, for Chaos Mesh we can't require that all application pod define `livenessProbe` and `readinessProbe`.

YangKeao:
@cwen0  If they don't set corresponding probe, I think the default value of `livenessProbe` and `readinessProbe` are "Success" as described [here](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes)

AngleNet:
Did you mean we should define `livenessProbe`, `readinessProbe` and `startupProbe` along with `MaxInjectedPercent` inside `Protection`? If we do this, is it too heavy for user to specify too many probes for a chaos ?

AngleNet:
> @cwen0 If they don't set corresponding probe, I think the default value of `livenessProbe` and `readinessProbe` are "Success" as described [here](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes)

Great ! I think we have agreed on the details as followings :
1. Add a `Protection` field to the all of the chaos types which defines a `MaxInjectedPercent` to deny the chaos action once the ratio of healthy pods is below it.
2. Take the `PodCondition` of type `Ready` as a indicator of whether a pod is healthy
3. Let the user define their protection rules for whether the application cluster is healthy or not

Feel free to correct me if I'm wrong about this.

AngleNet:
I will close this PR since it is filed in a new PR #363

PR #360, opened 2020-03-22
Support helm3 in `make install`

yujunz:

Errors running `make install` when helm version 3 is installed

Add support for helm 3

### Check List

Tests

 - Manual test

`make install`

```release-note
NONE
```


Makefile
```diff
@@ -98,7 +100,7 @@ run: generate fmt vet manifests
 # Install CRDs into a cluster
 install: manifests
 	kubectl apply -f manifests/crd.yaml
-	helm install helm/chaos-mesh --name=chaos-mesh --namespace=chaos-testing
+	bash -c '[[ `helm version --client --short` == "Client: v2"* ]] && helm install helm/chaos-mesh --name=chaos-mesh --namespace=chaos-testing || helm install chaos-mesh helm/chaos-mesh --namespace=chaos-testing;'
```
cwen0: I see you have defined `HELM_INSTALL`, why don't you use it at here?

yujunz: Cleaned up. I was trying different way for conditional script in Makefile.

Issue #292, opened 2020-02-28
UCP: Support to set default value of chaos object by AdmissionWebhook


cwen0:
## Description

When users apply a network chaos, some error may be reported: 

```
The NetworkChaos "network-delay-example" is invalid: []: Invalid value: map[string]interface {}{"spec":map[string]interface {}{"action":"delay", "delay":map[string]interface {}{"latency":"10ms"}, "duration":"30s", "mode":"one", "scheduler":map[string]interface {}{"cron":"@every 60s"}, "selector":map[string]interface {}{"labelSelectors":map[string]interface {}{"app":"web-show"}, "namespaces":[]interface {}{"default"}}}, "apiVersion":"pingcap.com/v1alpha1", "kind":"NetworkChaos", "metadata":map[string]interface {}{"namespace":"chaos-testing", "creationTimestamp":"2020-02-28T01:35:05Z", "annotations":map[string]interface {}{"kubectl.kubernetes.io/last-applied-configuration":"{\"apiVersion\":\"pingcap.com/v1alpha1\",\"kind\":\"NetworkChaos\",\"metadata\":{\"annotations\":{},\"name\":\"network-delay-example\",\"namespace\":\"chaos-testing\"},\"spec\":{\"action\":\"delay\",\"delay\":{\"latency\":\"10ms\"},\"duration\":\"30s\",\"mode\":\"one\",\"scheduler\":{\"cron\":\"@every 60s\"},\"selector\":{\"labelSelectors\":{\"app\":\"web-show\"},\"namespaces\":[\"default\"]}}}\n"}, "generation":1, "uid":"8bc122fd-59ca-11ea-a548-0242ac110002", "name":"network-delay-example"}}: validation failure list:
spec.delay.correlation in body is required
spec.delay.jitter in body is required
```

I think To avoid this, we may need `spec.delay.correlation` and `spec.delay.jitter` to be allowed to be unset in NetworkChaos and use a default value.

## Score 

* 604.6

## Mentor (s)

@cwen0 

## Recommended Skills

* Golang
* Kubernetes

## Learning Materials

* [Development guide of chaos-mesh](https://github.com/pingcap/chaos-mesh/wiki/Development-Overview)

Gallardot:
+1
As the project evolves, we will have more and more CRDS and more and more fields.

[...]

PR #361, opened 2020-03-22
Set default value for correlation and jitter of NetworkChaos

Gallardot:
UCP #292 
#292 
### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

Side effects

 - 

Related changes

 - 

```release-note
NONE
```

### DIFF
--- a/api/v1alpha1/networkchaos_webhook.go
+++ b/api/v1alpha1/networkchaos_webhook.go
@@ -23,6 +23,14 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
+const (
+	// DefaultJitter defines default value for jitter
+	DefaultJitter = "0ms"
+
+	// DefaultCorrelation defines default value for correlation
+	DefaultCorrelation = "0"
+)
+
 // log is for logging in this package.
 var networkchaoslog = logf.Log.WithName("networkchaos-resource")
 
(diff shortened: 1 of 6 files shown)


zhouqiang-cl:
Did we need to update doc about network

api/v1alpha1/networkchaos_webhook.go
```diff
@@ -44,6 +52,20 @@ func (in *NetworkChaos) Default() {
 	in.Spec.Selector.DefaultNamespace(in.GetNamespace())
 	// the target's namespace selector
 	in.Spec.Target.TargetSelector.DefaultNamespace(in.GetNamespace())
+
+	in.defaultDelay()
```
AngleNet: Shall we add it to `NetworkChaosSpec` to keep same style as other's setting default ?

cwen0: I agree with this opinion @Gallardot

Yisaer:
I think it is necessary to add unit test to verify the validating and default webhook.  This request is ok to me. I suggest to create issue to mention us for the unit test later.

api/v1alpha1/networkchaos_webhook.go
Gallardot: Good suggestion!All done!

Gallardot:
> Did we need to update doc about network?

@zhouqiang-cl I have updated the doc.

AngleNet:
LGTM

Gallardot:
> I think it is necessary to add unit test to verify the validating and default webhook. This request is ok to me. I suggest to create issue to mention us for the unit test later.

@Yisaer I have added some unit test for the validating and default webhook before.  Maybe you means e2E test for the webhook?

PR #363, opened 2020-03-22
Add `Protector` for all of the chaos to prevent unintentional injection

### LINKED ISSUE #179: UCP: Add protect policy to prevent the injected system from being unintentionally damaged

AngleNet:
Fix #179 
See #345 
### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

Related changes

```release-note
NONE
```


AngleNet:
@cwen0 Help me to take a look at

api/v1alpha1/common_types.go
```diff
@@ -14,10 +14,35 @@
...
+// entire application cluster. That may be not our intention. We should protect
+// our application cluster from that disaster via a protector to make the
+// application cluster always in available.
+type ProtectorSpec struct {
+
+	// MaxInjectPercent specifies the ratio of maximum pods involved in the
+	// chaos to be injected later.  A value of less than or equal to 0, or
+	// greater than 100, we see it as illegal. It takes no effects for the
+	// protection.
+	MaxInjectPercent int `json:"maxInjectPercent,omitempty"`
```
Yisaer: I suggest replace the `MaxInjectPercent` to the `MinUnInfectedCount`.  The uninfected replica counts would be flexible during each round of syncing considering the `pod-kill` case or the external factor while the `MinUnInfectedCount` would always guarantee the expected health instances.

AngleNet: I think use `MinUnInfectedCount` is another point of view. Maybe we can support both at the same time ? @Yisaer

AngleNet: About ** considering the pod-kill case**, I did not fully understand this.I guess that in each round of a `pod-kill` action the selected pods will be decreased, however the `MaxInjectPercent` should calculated according to the number of initial selected pods. The implementation did that exactly right now. For example, we set `MaxInjectPercent` to `30`. If we selected 10 pods in the first reconciling round of the `pod-kill`, the max injected pods should always less than or equal to `10*0.3` in the later rounds. @Yisaer

Yisaer: Things would be much complex if we use radio here. And I also don't recommend recording the max injected pod counts in the first reconciling, considering not only the scaling (especially elastic auto-scaling) is common to see in the Kubernetes, (If the total pods which controlled by deployment have 10 replicas at initial, then we recorded 3 as max injected pods, and after that, the replicas of the deployment become 2, and the chaos-mesh kill all the pods finally), but also this design would make chaos-mesh controller more and more stateful. (I didn't see the benefit to use radio here for solving #179 comparing the fixed uninfected pods strategy)

AngleNet: Agreed ! I missed that services will be scaled during chaos injection. A stateless service is more preferred.

cwen0:
@AngleNet Is there any progress on this pr?

AngleNet:
@cwen0 Sorry for the late reply. I think it's almost there. However, I do not have time working on it lately. Is there kind guys to take over this?

Yisaer:
I can take this after June 1th ( after TiDB 4.0 release 😂 )

PR #370, opened 2020-03-23
Pause experiment

### LINKED ISSUE #294: UCP: Add a field in chaos object to allow to pause the current chaos experiment
## Description

Sometimes, we want to stop the current chaos experiment to debug the application and the chaos experiment will be restart later. So we don't want to delete this chaos experiment.

In order to meet the above scenario, we want to add an option to each chaos object to temporarily pause the current experiment. we can directly use `kubectl patch` method to modify this field to start or stop the chaos experiment. 

If the chaos action has been applied successfully,  when we pause the chaos experiment, we should cancel the current chaos action. 

## Score 

* 936

## Mentor(s)
* @cwen0 

## Recommended Skills

* Golang
* Kubernetes 
* CRD

## Learning Materials

* [CRD](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/custom-resources/)

oraluben:
UCP #294 
Close #294 

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


doc/pause.md
```diff
@@ -0,0 +1,112 @@
...
+chaos-controller-manager-7f67fbcfdc-nm6jl   1/1     Running             0          8s
+chaos-daemon-fbpsz                          1/1     Running             0          5m56s
+chaos-daemon-z7p5f                          1/1     Running             0          5m26s
+chaos-daemon-zhz2k                          0/1     ContainerCreating   0          6s
+```
+
+Pause the running chaos:
+```shell script
+$ kubectl patch podchaos pod-kill-example --namespace chaos-testing --type merge --patch 'status:
+  paused: true'
```
Gallardot: Why not defined the paused field on spec? I think define the paused field on status is not friendly for user.

oraluben: That's a point I'm not sure about. Firstly I have the field defined in spec, then I think it's a "status" thing and is not supposed to be in the "specification" of chaos, so here it is.
But on the other hand, I don't think user should patch on status, for it's better to be automatically generated/modified.

Gallardot: I think `spec` fields are for controller and user, `status` fields are for controller only.

@cwen0 @YangKeao What's your opinion.

Yisaer: According to `Deployment`'s spec, I tend to leave paused into spec.

https://github.com/kubernetes/api/blob/b5bd82427fa87d8b6fdf2c0b4cc2a2115c0c6de9/apps/v1/types.go#L303

controllers/common/types.go
```diff
@@ -69,6 +70,20 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+
+			err = r.Recover(ctx, req, chaos)
+			if err != nil {
+				r.Log.Error(err, "failed to pause chaos")
+				return ctrl.Result{Requeue: true}, err
+			}
+			status.Experiment.Phase = v1alpha1.ExperimentPhasePaused
+		} else {
+			r.Log.Error(err, "failed to pause chaos, the common chaos is not running.", "status", status.Experiment.Phase)
+			return ctrl.Result{}, errors.New("common chaos is not running")
```
Yisaer, line 85: I think this message would be misleading. If we paused the chaos when it is not running, we just skip the running process and I think it's also a successful puased action.

I also suggest send an event here to indicate the chaos is paused.

doc/pause.md
YangKeao: I prefer to set `Paused` field in spec .

controllers/common/types.go
oraluben: Thanks for your review, that makes more sense.
And I'm not sure if I get you about the "send an event" part. I have sent events in controller like https://github.com/pingcap/chaos-mesh/pull/370/files#diff-a52c75ede938739cb8d8e45acdfef8acR58-R62, are you referring to also send events in `{twophase,common}.Reconciler`?

controllers/twophase/twophase_suite_test.go
```diff
@@ -94,10 +94,16 @@ func (in *fakeTwoPhaseChaos) GetStatus() *v1alpha1.ChaosStatus {
 	return &in.Status
 }
 
+// IsDeleted returns whether this resource has been deleted
 func (in *fakeTwoPhaseChaos) IsDeleted() bool {
 	return in.Deleted
 }
 
+// IsPaused returns whether this resource is paused
```
AngleNet, line 102: `whether this resource is paused` to `whether this resource has been paused` ? To keep same doc style.

doc/pause.md
AngleNet: @oraluben Great ! I think you have already fixed this up. However, the doc should be fixed same either.

cwen0: @oraluben Please update this document.

oraluben:
Everything should be fine except a e2e test, and I'm having some trouble writing one because of my poor network connection (to run the test). I think I can resolve this in the next two days.

oraluben:
@Gallardot @AngleNet @cwen0 @Yisaer PTAL

test/e2e/chaos/pod.go
```diff
@@ -184,6 +186,86 @@ var _ = ginkgo.Describe("[chaos-mesh] Basic", func() {
...
+
+		err = wait.Poll(5*time.Second, 5*time.Minute, func() (done bool, err error) {
+			chaos := &v1alpha1.PodChaos{}
+			err = cli.Get(ctx, chaosKey, chaos)
+			if chaos.Status.Experiment.Phase == chaosmeshv1alpha1.ExperimentPhasePaused {
+				return true, nil
+			}
+			return false, err
+		})
+		framework.ExpectNoError(err, "Check paused chaos failed")
```
Yisaer, line 245: Almost lgtm, but the e2e case should be revised. I think we could record the pods uid after we create the nginx Deployments. Then, we apply the pod-kill chaos and checking wether is there any pod get killed by comparing the pods uid. Next, we patch the pause into pod-kill chaos. After that, we record the pods uid again and expect a timeout error from checking the pods whether haven't been deleted for 5 mins.

controllers/twophase/types.go
```diff
@@ -105,6 +105,22 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		}
 
 		status.Experiment.Phase = v1alpha1.ExperimentPhaseFinished
+	} else if chaos.IsPaused() {
+		if status.Experiment.Phase == v1alpha1.ExperimentPhaseRunning {
```
fewdan, line 109: What happens when "chaos" is not "running"? For example, the current "chaos" has been "finished" or "failed".

fewdan: Rest LGTM

oraluben: > https://github.com/pingcap/chaos-mesh/pull/370#discussion_r401341634

@Yisaer 's idea is that we could pause a chaos that's not running. I think that makes some sense for the scenario in which we don't want to delete this chaos but to rerun later.

test/e2e/chaos/pod.go
oraluben: @Yisaer PTAL at the new e2e test

doc/pause.md
```diff
@@ -0,0 +1,149 @@
...
+      hostIP: 172.17.0.2
+      message: delete pod
+      name: chaos-daemon-g4s25
+      namespace: chaos-testing
+      podIP: ""
+    startTime: "2020-04-10T06:44:54Z"
+  phase: ""
+NAME                                        READY   STATUS              RESTARTS   AGE
+chaos-controller-manager-7f67fbcfdc-n2dht   1/1     Running             0          47s
+chaos-daemon-2tfcw                          0/1     ContainerCreating   0          2s
```
cwen0, line 63: `ContainerCreating` may confuse users.

test/e2e/chaos/pod.go
```diff
@@ -214,3 +330,27 @@ func waitDeploymentReady(name, namespace string, cli kubernetes.Interface) error
...
+				found = true
+				break
+			}
+		}
+		if !found {
+			return true
+		}
+	}
+	return false
+}
```
Yisaer, line 356: The logic here seems complex. We can change them into `string` and sort the string array and compare.

test/e2e/chaos/pod.go
```diff
@@ -214,3 +330,27 @@ func waitDeploymentReady(name, namespace string, cli kubernetes.Interface) error
...
 	})
 }
+
+func getUIDs(pods []corev1.Pod) []types.UID {
+	var ids []types.UID
+	for _, pod := range pods {
+		ids = append(ids, pod.UID)
+	}
+	return ids
+}
```
Yisaer, line 340: we could return string array here.

test/e2e/chaos/pod.go
```diff
@@ -214,3 +328,12 @@ func waitDeploymentReady(name, namespace string, cli kubernetes.Interface) error
 		return true, nil
 	})
 }
+
+func getUIDs3Pods(pods []corev1.Pod) [3]string {
```
Yisaer, line 332: why we limit 3 here. I think this is not necessary.

oraluben: I was trying to return an array here but it looks like there's no way to return a not-fixed-length array in a go function. Is there some better design for this kind of scenario?

test/e2e/chaos/pod.go
```diff
@@ -184,6 +186,118 @@ var _ = ginkgo.Describe("[chaos-mesh] Basic", func() {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "nginx-kill",
+				Namespace: ns,
+			},
+			Spec: v1alpha1.PodChaosSpec{
+				Selector: v1alpha1.SelectorSpec{
+					Namespaces:     []string{ns},
+					LabelSelectors: map[string]string{"app": "nginx"},
+				},
+				Action: v1alpha1.PodKillAction,
```
cwen0, line 219: I think `pod-failure` is more available for this case, we can check the status of all pods to judge wheather the `pause`  is work.

Yisaer: we would finally cover all kinds of chaos in every case in e2e, "schedule / duration / pause", current e2e test is ok.

test/e2e/chaos/pod.go
Yisaer: I think you could build a function which receives 2 pod array and compare whether they have same uid and return the boolean result.

Yisaer:
@oraluben  plz solve the conflict, and we can run e2e test in our ci now.

test/e2e/chaos/pod.go
oraluben: I see

Issue #362, opened 2020-03-22
Support incremental build


yujunz:
## Feature Request

**Is your feature request related to a problem? Please describe:**
No

**Describe the feature you'd like:**
binary seems always rebuilt regardless code changes. 

```
make image-chaos-mesh
make image-chaos-daemon
```

`pingcap/binary` will be built twice although it is not changed at all between the two steps. This is slowing down the development process.

**Describe alternatives you've considered:**
Options for solution of incremental build
* mount build cache to container
* looks for alternative toolchain other than make

YangKeao:
Sorry. I didn't reproduce this problem. Docker will cache the build result, doesn't it? Can you provide your docker version?

```
docker build -t pingcap/binary  .

Sending build context to Docker daemon  83.99MB
Step 1/18 : FROM golang:alpine3.10 AS build_base
 ---> dd50f8b1f673
Step 2/18 : ARG HTTPS_PROXY
 ---> Using cache
 ---> ed05e4f428f0
Step 3/18 : ARG HTTP_PROXY
 ---> Using cache
 ---> 4d7a51c92635
Step 4/18 : RUN apk add --no-cache gcc g++ make bash git
 ---> Using cache
 ---> b8b43c5ddfe7
Step 5/18 : ENV GO111MODULE=on
 ---> Using cache
 ---> 76f8023e6424
Step 6/18 : RUN mkdir /src
 ---> Using cache
 ---> 1fcf57f78a5e
Step 7/18 : WORKDIR /src
 ---> Using cache
 ---> bf4900ff1bb1
Step 8/18 : COPY go.mod .
 ---> Using cache
 ---> 7561d5e369fc
Step 9/18 : COPY go.sum .
 ---> Using cache
 ---> 722ca97f3341
Step 10/18 : RUN go mod download
 ---> Using cache
 ---> b0f51f67a398
Step 11/18 : RUN go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.2.5
 ---> Using cache
 ---> bb42256913bd
Step 12/18 : RUN go get golang.org/x/tools/cmd/goimports
 ---> Using cache
 ---> 26ddc7dc965e
Step 13/18 : FROM build_base AS binary_builder
 ---> 26ddc7dc965e
Step 14/18 : ARG HTTPS_PROXY
 ---> Using cache
 ---> 35bb41474e77
Step 15/18 : ARG HTTP_PROXY
 ---> Using cache
[...]

PR #374, opened 2020-03-24
Build binary without vet and fmt

YangKeao:

* Add some folders into `.dockerignore`

* Remove `fmt` and `vet` as dependencies for binaries.

Related with #362 , this will accelerate binary image building.

### Check List

Tests 

 - Manual test

Code changes

`make test` will also run `fmt` and `vet`, so this won't break CI.

```release-note
NONE
```


Makefile
```diff
@@ -71,22 +71,22 @@ else
 endif
 
 # Build chaos-daemon binary
-chaosdaemon: generate fmt vet
+chaosdaemon: generate
```
cwen0: Can we add a single command `make check` for users to check code? 

```
check: fmt vet lint goimportxxxx
```

YangKeao: 🍻 I agree with you. Though adding `fmt` to `check` is a little wired, I cannot figure out a better name for it.

PR #375, opened 2020-03-24
automatically calculate duration

YangKeao:

Calculate time chaos offset automatically.

### Check List 

Tests 

 - Manual test

Code changes

 - Has Go code change

Side effects

Related changes

Should use new form of `timeOffset` field.


controllers/timechaos/types.go
```diff
@@ -285,11 +286,19 @@ func (r *Reconciler) applyContainer(ctx context.Context, client chaosdaemon.Chao
 		return err
 	}
 
-	r.Log.Info("setting time shift", "mask", mask, "sec", chaos.Spec.TimeOffset.Sec, "nsec", chaos.Spec.TimeOffset.NSec)
+	duration, err := time.ParseDuration(chaos.Spec.TimeOffset)
```
cwen0: split this logic to a single help function?  So we can add the unit test for this.

doc/time_chaos.md
```diff
@@ -17,9 +17,7 @@ spec:
   selector:
     labelSelectors:
       "app.kubernetes.io/component": "pd"
-  timeOffset:
-    sec: 100000
-    nsec: 100000
+  timeOffset: "-10m100ns"
```
cwen0: Remember to update the Wiki

controllers/timechaos/types.go
YangKeao: Yes!

Issue #365, opened 2020-03-22
UCP: Using more Validator and Defaulter, more friendly for user


Gallardot:
## Description 

Through implement the Defaulter and the Validator interface, We have  implemented admission webhooks for chaos object. But currently we only have some simple rules.I think more rules can make chaos-mesh more friendly for user.
for example A:
https://github.com/pingcap/chaos-mesh/blob/41b31e5e3375654e24d83932dd624d15a191fa31/api/v1alpha1/common_types.go#L65-L75

**Cron should be match the spec rules**

for example B:
https://github.com/pingcap/chaos-mesh/blob/41b31e5e3375654e24d83932dd624d15a191fa31/api/v1alpha1/networkchaos_types.go#L248-L274

**DelaySpec.Latency and DelaySpec.Jitter should be time format.**

If user typo, execute `kubectl create -f chaos.yaml` is OK. But it doesn't actually work. The user can only check the log of chaos-mesh. It's a waste of time.

So I think make more Validation rules by the `webhook.Validator` and(or)  [crd-validation](https://book.kubebuilder.io/reference/markers/crd-validation.html) is more friendly for user.If user typo, user knows immediately what the problem is.

At least 5 cheker for the chaos objects

1. `Specific format checker`
2. `Value range checker`
3. `Null checker`
4. `Regular checker`
5.  `Combination checker`
6.  etc......

I think this proposal also match the [TiDB Usability Challenge Program](https://github.com/pingcap/community/blob/master/challenge-programs/challenge-program-season-2.md).  😄

## Score 

* 936

## Mentor(s)

* @cwen0

Contact the mentors: #tidb-challenge-program channel in TiDB Community Slack Workspace

## Recommended Skills

* Golang
* Kubernetes

## Learning Materials

* [Development guide of chaos-mesh](https://github.com/pingcap/chaos-mesh/wiki/Development-Overview) 
* [Chaos-mesh Contributing Tutorial](https://yisaer.gitbook.io/chaos-mesh-contributing-tutorials/)
*  [crd-validation](https://book.kubebuilder.io/reference/markers/crd-validation.html)

cwen0:
> I think this proposal also match the TiDB Usability Challenge Program. 😄 

Thanks for your proposal, this is a good suggestion. We can split this proposal to multi easy tasks.

you06:
@Gallardot try again please

Issue #373, opened 2020-03-23
Validate duration and cron schedule conflict in webhook


at15:
## Feature Request

**Is your feature request related to a problem? Please describe:**

From https://github.com/pingcap/chaos-mesh/pull/353#discussion_r396348890

When user give a duration that is smaller than cron interval in networkchaos.
Controller generates warning `nextRecover shouldn't be later than nextStart`

```yaml
  duration: "10s"
  scheduler:
    cron: "@every 5s
```

**Describe the feature you'd like:**

Validate duration and cron schedule in webhook instead of at runtime inside controller.
Reject invalid spec directly when user submit it.

**Describe alternatives you've considered:**

N/A

**Teachability, Documentation, Adoption, Migration Strategy:**

Should mention in doc that user should be aware of the different duration they are specifying to avoid conflict. I think even delay could cause conflict (unexpected behavior), i.e. a delay of 5s with a cron of 1s

at15:
It seems it's not limit to network chaos, there is a `ValidateScheduler` interface for each chaos and most implementations are same, only checks at least one of duration and scheduler is presented.

https://github.com/pingcap/chaos-mesh/blob/master/api/v1alpha1/iochaos_webhook.go#L83-L95

https://github.com/pingcap/chaos-mesh/blob/master/api/v1alpha1/networkchaos_webhook.go#L106-L119

Maybe we should merge them and use a default implementation for most chaos unless they have different requirement when it comes to duration and schedule.

cwen0:
> Maybe we should merge them and use a default implementation for most chaos unless they have different requirement when it comes to duration and schedule. 

@at15 This is a good idea, I've wanted to do this for a long time. Can you help us optimize this?

YangKeao:
> 
> 
> > Maybe we should merge them and use a default implementation for most chaos unless they have different requirement when it comes to duration and schedule.
> 
> @at15 This is a good idea, I've wanted to do this for a long time. Can you help us optimize this?

I have tried to do this with embedded struct. However, the code generator seems cannot work with embedded structs well 😿

Gallardot:
> I have tried to do this with embedded struct. However, it seems that the code generator cannot work with embedded structs well 😿

@YangKeao 

How about this ?

```
type SchedulerSpec struct {
	Cron string `json:"cron,omitempty"`
        Duration *string `json:"duration,omitempty"`
}
```

I think `Duration` and `Cron` are very relevant fields. Why not define them together？

[...]

PR #387, opened 2020-03-28
Using more Validator and Defaulter, more friendly for user

Gallardot:
UCP #365 

#365 

Check the list

common chaos field:

- [x] spec.mode
- [x] spec.value
- [x] spec.duration
- [x] spec.scheduler
- [x] spec.scheduler.cron

iochaos field:

- [x] spec.action
- [x] spec.layer
- [x] spec.delay
- [x] spec.errno
- [x] spec.percent

timechaos field:

- [x] spec.timeOffset
- [x] spec.ClockIds

networkchaos field:

- [x] spec.action
- [x] spec.bandwidth.peakrate
- [x] spec.bandwidth.minburst
- [x] spec.direction
- [x] spec.delay
- [x] spec.loss
- [x] spec.duplicate
- [x] spec.corrupt
- [x] spec.bandwidth
- [x] spec.target

podschaos field:

- [x] spec.action
- [x] spec.containerName

kernelchaos field:

- [x] spec.failKernRequest.failtype
- [x] spec.failKernRequest.probability
- [x] spec.failKernRequest.times

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

Side effects

 - None

Related changes

 - None

```release-note
NONE
```


api/v1alpha1/common_validation.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+// ValidateScheduler validates the scheduler and duration
+func ValidateScheduler(duration *string, scheduler *SchedulerSpec, spec *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	schedulerField := spec.Child("scheduler")
+	durationField := spec.Child("duration")
+
+	if duration != nil && scheduler != nil {
+		_, err := cron.ParseStandard(scheduler.Cron)
```
at15: For #373 you can do type check on the cron, if the syntax is `every 5m` then the type should be  `ConstantDelaySchedule`. However for normal cron syntax it's a bit hard to compare, you need to check the unit.

Gallardot: @at15 
It is better to solve #373  on another PR. Otherwise current PR will get complicated. 😂

zhouqiang-cl:
@cwen0 @YangKeao @yeya24 @Yisaer @AngleNet @at15 This is a big PR.  Please help to take a look. 

thanks

api/v1alpha1/common_validation.go
```diff
@@ -0,0 +1,97 @@
...
+
+const (
+	// ValidateSchedulerError defined the error message for ValidateScheduler
+	ValidateSchedulerError = "schedule and duration should be omitted or defined at the same time"
+
+	// ValidatePodchaosSchedulerError defined the error message for ValidateScheduler of Podchaos
+	ValidatePodchaosSchedulerError = "schedule should be omitted"
+
+	// ValidateValueParseError defined the error message for value parse error
+	ValidateValueParseError = "parse value field error"
```
yeya24: Please change all `defined` to `defines`

api/v1alpha1/common_validation.go
```diff
@@ -0,0 +1,97 @@
...
+	} else if duration == nil && scheduler == nil {
+		return nil
+	}
+
+	allErrs = append(allErrs, field.Invalid(schedulerField, scheduler, ValidateSchedulerError))
+	return allErrs
+}
+
+// ValidateValue validates the value with podmode
+func ValidateValue(value string, mode PodMode, valueField *field.Path) field.ErrorList {
```
yeya24: Actually I think this function name is not that clear. What about changing it to `ValidatePodMode`? Since the value checking is based on the pod mode field

api/v1alpha1/kernelchaos_types.go
```diff
@@ -93,6 +94,8 @@ type FailKernRequest struct {
 	//   1. https://www.kernel.org/doc/html/latest/fault-injection/fault-injection.html
 	//   2. http://github.com/iovisor/bcc/blob/master/tools/inject_example.txt
 	// to learn more
+	// +kubebuilder:validation:Maximum=2
+	// +kubebuilder:validation:Minimum=0
```
yeya24: Is that true?

api/v1alpha1/networkchaos_types.go
```diff
@@ -266,17 +270,17 @@ type DelaySpec struct {
 }
 
 // ToNetem implements Netem interface.
-func (delay *DelaySpec) ToNetem() (*chaosdaemonpb.Netem, error) {
-	delayTime, err := time.ParseDuration(delay.Latency)
+func (in *DelaySpec) ToNetem() (*chaosdaemonpb.Netem, error) {
```
yeya24: Why change all to `in`? `in` is not that readable...

at15: I think it's because IDE like goland would show warning for having all method receivers using same name. `in` is used by generator when generating `DeepCopy`

api/v1alpha1/common_validation.go
Gallardot: Good idea. It makes it clearer.

api/v1alpha1/kernelchaos_types.go
Gallardot:
```
// FailType indicates what to fail, can be set to '0' / '1' / '2'
	// If `0`, indicates slab to fail (should_failslab)
	// If `1`, indicates alloc_page to fail (should_fail_alloc_page)
	// If `2`, indicates bio to fail (should_fail_bio)
	// You can read:
	//   1. https://www.kernel.org/doc/html/latest/fault-injection/fault-injection.html
	//   2. http://github.com/iovisor/bcc/blob/master/tools/inject_example.txt
	// to learn more
```

Yes, refer to its comments.

api/v1alpha1/networkchaos_types.go
Gallardot: Yes,this is caused by goland and generator.

api/v1alpha1/common_validation.go
```diff
@@ -0,0 +1,97 @@
...
+
+// ValidatePodMode validates the value with podmode
+func ValidatePodMode(value string, mode PodMode, valueField *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	switch mode {
+	case FixedPodMode:
+		num, err := strconv.Atoi(value)
+		if err != nil {
+			allErrs = append(allErrs, field.Invalid(valueField, value, ValidateValueParseError))
+		} else if num <= 0 {
```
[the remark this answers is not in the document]
Gallardot:
```
num, err := strconv.Atoi(value)
```

If an error occurred, the `num` is `0`. User will receive two error messages. But I think it's OK for User.I will change it.

cwen0:
@YangKeao @yeya24 PTAL

api/v1alpha1/common_validation.go
```diff
@@ -0,0 +1,101 @@
...
+			allErrs = append(allErrs, field.Invalid(valueField, value,
+				fmt.Sprintf(ValidateValueParseError, err)))
+		}
+
+		if percentage <= 0 || percentage > 100 {
+			allErrs = append(allErrs, field.Invalid(valueField, value,
+				fmt.Sprintf("value of %d is invalid, Must be (0,100] with mode:%s",
+					percentage, mode)))
+		}
+	}
```
YangKeao, line 99: Should we check unknown mode here?

Gallardot: I think we don't need to check unknown mode. Because we've already used the CRD-Validator to do the same things.

```
// +kubebuilder:validation:Enum=one;all;fixed;fixed-percent;random-max-percent
Mode PodMode `json:"mode"`
```

The unknown mode will occur an error.

YangKeao: Great 👍

PR #396, opened 2020-03-31
Set mirrors only if not empty

YangKeao:

Makefile will pass empty MIRROR into Dockerfile, and this will override the default settings for ARG MIRROR.

### Check List 

Tests 

 - Manual test

Code changes

 - Has Go code change


images/chaos-kernel/Dockerfile
```diff
@@ -7,7 +7,8 @@ ARG MIRROR=http://archive.ubuntu.com/ubuntu
 
 RUN apt-get update && apt-get install -y ca-certificates gnupg2 wget
 
-RUN sed -i "s|http://archive.ubuntu.com/ubuntu|$MIRROR|g" /etc/apt/sources.list
+RUN if [ ! -z "$MIRROR" ]; then sed -i "s|http://archive.ubuntu.com/ubuntu|$MIRROR|g" /etc/apt/sources.list; fi
```
zhouqiang-cl: 😯

PR #408, opened 2020-04-09
Add e2e into CI

Yisaer:
Add e2e into ci

```release-note
NONE
```


ci/pingcap_chaos_mesh_build_kind.groovy
```diff
@@ -0,0 +1,252 @@
...
+import groovy.transform.Field
+
+@Field
+def podYAML = '''
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    # we pretend as tidb-operator in order not to meet tidb-operator-e2e job in the same node
+    app: tidb-operator-e2e
```
Yisaer, line 14: we would share same cluster with tidb-operator-e2e, so we pretend ourselves like tidb-operator-e2e in order not to run both kind of e2e job in one node.

test/e2e/config/config.go
```diff
@@ -0,0 +1,31 @@
+// Copyright 2019 PingCAP, Inc.
```
fewdan, line 1: s/2019/2020/g

Yisaer: updated.

PR #417, opened 2020-04-13
Revise install scripts

Yisaer:
Current installing command would install the binary and edit the user's bin path. This request it to install the binary we want into the specific dir ($porject/output/bin) and use these specific binary (e.g: $OUTPUT_BIN/kind) in our scripts.

Related changes

```release-note
NONE
```


hack/kind-cluster-build.sh
```diff
@@ -132,91 +134,16 @@ EOF
...
 echo "start to create k8s cluster"
-kind create cluster --config ${configFile} --image kindest/node:${k8sVersion} --name=${clusterName}
-kind get kubeconfig --name=${clusterName} > ${kubeconfigPath}
+${KIND_BIN} create cluster --config ${configFile} --image kindest/node:${k8sVersion} --name=${clusterName}
+${KIND_BIN} get kubeconfig --name=${clusterName} > ${kubeconfigPath}
 export KUBECONFIG=${kubeconfigPath}
 
-echo "deploy docker registry in kind"
-registryNode=${clusterName}-control-plane
-registryNodeIP=$(kubectl get nodes ${registryNode} -o template --template='{{range.status.addresses}}{{if eq .type "InternalIP"}}{{.address}}{{end}}{{end}}')
```
Yisaer, line 141: As we could use `kind load docker-image`, the registry daemon is no longer needed.

zhouqiang-cl:
Seems we remove install.sh, but in blog we need it?

install.sh
```diff
@@ -1,974 +0,0 @@
...
-# Finally, when all dependencies are installed, chaos-mesh will be installed using helm.
-
-usage() {
-    cat << EOF
-This script is used to install chaos-mesh.
-Before running this script, please ensure that:
-* have installed docker if you run chaos-mesh in local.
-* have installed Kubernetes if you run chaos-mesh in normal Kubernetes cluster
-USAGE:
-    install.sh [FLAGS] [OPTIONS]
```
Yisaer, line 15: we no longer need this scripts.

hack/kind-cluster-build.sh
```diff
@@ -132,91 +134,16 @@ EOF
...
-kubectl apply -f ${ROOT}/manifests/local-volume-provisioner.yaml
-kubectl apply -f ${ROOT}/manifests/tiller-rbac.yaml
+${KUBECTL_BIN} apply -f ${ROOT}/manifests/local-volume-provisioner.yaml
+${KUBECTL_BIN} apply -f ${ROOT}/manifests/tiller-rbac.yaml
 
-kubectl create ns chaos-testing
+$KUBECTL_BIN create ns chaos-testing
 
-if [[ $(helm version --client --short) == "Client: v2"* ]]; then helm init --service-account=tiller --wait; fi
+${HELM_BIN} init --service-account=tiller --wait
```
zhouqiang-cl, line 146: we no longger support helm3?

hack/kind-cluster-build.sh
```diff
@@ -14,10 +14,10 @@ Options:
        -h,--help               prints the usage message
        -n,--name               name of the Kubernetes cluster,default value: kind
        -c,--nodeNum            the count of the cluster nodes,default value: 3
-       -k,--k8sVersion         version of the Kubernetes cluster,default value: v1.17.2
+       -k,--k8sVersion         version of the Kubernetes cluster,default value: v1.15.6
```
zhouqiang-cl, line 17: why do we change to v1.15.6?

hack/kind-cluster-build.sh
Yisaer: It seems we have to use user's default `helm` binary to initialize in this case. I will change back soon

hack/kind-cluster-build.sh
Yisaer: According to https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/, there would exist incompatible things between helm >= 2.14 and kubernetes >= 1.16.0

In order to avoid this problem, I suggest changing the default version for kubernetes as  1.15 

FYI: https://github.com/helm/helm/issues/6374

zhouqiang-cl:
@cwen0 @fewdan PTAL

install.sh
cwen0: We need to keep this script to help users install Chaos Mesh quickly.

zhouqiang-cl: Can we add this script back?

Yisaer: updated.

hack/lib.sh
```diff
@@ -47,7 +51,6 @@ function hack::verify_terraform() {
 }
```
yeya24, line 51: I am wondering why we have terraform here? Do we need to remove that?

hack/lib.sh
```diff
@@ -76,7 +79,6 @@ function hack::ensure_kubectl() {
     if hack::verify_kubectl; then
         return 0
     fi
-    echo "Installing kubectl v$KUBECTL_VERSION..."
     tmpfile=$(mktemp)
```
yeya24, line 82: If I already installed kubectl 1.17.0 in my env, but the requested kubectl version is 1.12.0, will it replace the kubectl in my env

Yisaer: it won't. The ensuring only install the bin in the $project/output/bin.

hack/lib.sh
Yisaer: removed now, we won't use terrform currently.

Yisaer:
@cwen0 @yeya24  PTAL

hack/lib.sh
yeya24: Make sense. Thanks for the clarification.

PR #419, opened 2020-04-14
Init chaos-server component

cwen0:

Init chaos-server component, `chaos-server` will store events to DB and provide API service for dashboard. 

* add apiserver pkg 
* add store pkg 
* refine collectors pkg 
* clean code

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


cmd/chaos-server/main.go
```diff
@@ -0,0 +1,91 @@
...
+	flag.BoolVar(&printVersion, "version", false, "print version information and exit")
+	flag.Parse()
+
+	conf, err := config.EnvironChaosServer()
+	if err != nil {
+		log.Error(err, "main: invalid configuration")
+		os.Exit(1)
+	}
+
+	version.PrintVersionInfo("Chaos collector")
```
yeya24, line 57: rename this?

images/chaos-server/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM alpine:3.10
+
+ARG HTTPS_PROXY
+ARG HTTP_PROXY
+
+RUN apk add tzdata --no-cache
+
+COPY --from=pingcap/binary /src/bin/chaos-server /usr/local/bin/chaos-server
```
yeya24, line 8: Please add a new line

pkg/apiserver/archive/archive.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// Register mounts our HTTP handler on the mux.
+func Register(r *gin.RouterGroup, s *Service) {
+	endpoint := r.Group("/archive")
+
+	endpoint.GET("/all", s.listArchiveExperiments)
+}
+
+func (s *Service) listArchiveExperiments(c *gin.Context) {}
```
yeya24, line 56: Maybe add a todo above the function so that we can remember there is something need to be done

pkg/apiserver/archive/archive.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/gin-gonic/gin"
+
+	"github.com/pingcap/chaos-mesh/pkg/core"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/pingcap/chaos-mesh/pkg/config"
+)
+
+// Service defines a handler service for archive experiments.
```
yeya24, line 26: Is archive also like events in the database? Can I ask what's the difference between archive and event?

pkg/config/config.go
```diff
@@ -0,0 +1,43 @@
...
+
+package config
+
+import (
+	"github.com/kelseyhightower/envconfig"
+)
+
+// ChaosServerConfig defines the configuration for Chaos Dashboard
+type ChaosServerConfig struct {
+	ListenHost           string `enconfig:"LISTEN_HOST" default:"0.0.0.0"`
```
yeya24, line 22: Is it `enconfig` or `envconfig`?

pkg/config/config.go
```diff
@@ -0,0 +1,43 @@
...
+	MetricAddress        string `enconfig:"METRIC_ADDRESS"`
+	EnableLeaderElection bool   `enconfig:"ENABLE_LEADER_ELECTION"`
+	Database             *DatabaseConfig
+}
+
+// DatabaseConfig defines the configuration for databases
+type DatabaseConfig struct {
+	// Archive  Chaos Experiments to DB.
+	Archive    bool
+	Driver     string `envconfig:"DATABASE_DRIVER"     default:"sqlite3"`
```
yeya24, line 33: Grafana didn't support sqlite3 data source. Is it okay to use it as the default?

cwen0: I think this is OK.  If users did not choose MySQL, we can't provide Grafana support.

pkg/apiserver/archive/archive.go
cwen0: This archive just defines for the experiment.  Maybe it ’s better to name it to "experiment".

pkg/config/config.go
yeya24: Make sense! But we need to mention this in our docs later

Makefile
```diff
@@ -129,7 +140,7 @@ tidy:
 	GO111MODULE=on go mod tidy
 	git diff --quiet go.mod go.sum
 
-image: image-chaos-daemon image-chaos-mesh image-chaos-fs image-chaos-scripts image-chaos-grafana image-chaos-dashboard image-chaos-kernel
+image: image-chaos-daemon image-chaos-mesh image-chaos-fs image-chaos-scripts image-chaos-grafana image-chaos-kernel
```
Gallardot, line 143: add image-chaos-server

Makefile
```diff
@@ -161,7 +172,6 @@ docker-push:
 	docker push "${DOCKER_REGISTRY_PREFIX}pingcap/chaos-daemon:latest"
 	docker push "${DOCKER_REGISTRY_PREFIX}pingcap/chaos-scripts:latest"
 	docker push "${DOCKER_REGISTRY_PREFIX}pingcap/chaos-grafana:latest"
-	docker push "${DOCKER_REGISTRY_PREFIX}pingcap/chaos-dashboard:latest"
```
Gallardot, line 164: docker push chaos-server images

cmd/chaos-server/main.go
```diff
@@ -0,0 +1,91 @@
+// Copyright 2019 PingCAP, Inc.
```
fewdan, line 1: s/2019/2020/g

pkg/config/config.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/kelseyhightower/envconfig"
+)
+
+// ChaosServerConfig defines the configuration for Chaos Dashboard
+type ChaosServerConfig struct {
+	ListenHost           string `envconfig:"LISTEN_HOST" default:"0.0.0.0"`
+	ListenPort           int    `envconfig:"LISTEN_PORT" default:"2333"`
+	MetricAddress        string `envconfig:"METRIC_ADDRESS"`
+	EnableLeaderElection bool   `envconfig:"ENABLE_LEADER_ELECTION"`
+	Database             *DatabaseConfig
```
yeya24, line 26: I want to start some discussion about our configuration. Do we really want to use configurations as env vars? This is not common currently, I would say better to use a config file, or just flags. Any thoughts? @cwen0  @fewdan  @Yisaer

cwen0: In my opinion, if we are running these components in a Kubernetes environment, using environment variables is not much different from using a configuration file. In addition, if we think that using config configuration files is better in the future, we can replace them with configuration files. This change is tiny.

yeya24: That's might not be a tiny change. It will cause some back-compatibility issues, so it is hard to change one to another way.
Env vars and flags are not much different. But using the config file approach is quite useful if we want to have something like hot reload in the future (is it possible?).

yeya24: But I think it is okay for now, I don't want to add more work to you. So we can accept this first or change this in another PR if it is really needed.

doc/docs.go
```diff
@@ -0,0 +1,14 @@
...
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package doc
```
fewdan, line 14: What do you plan to put in this file in the future?

pkg/config/config.go
fewdan: I think there is no essential difference between the two. You can maintain the status quo first and make changes if necessary.

doc/docs.go
cwen0: We need to use this directory to generate swagger file

PR #441, opened 2020-04-22
Add timechaos and pause timechaos e2e test case

yeya24:

In order to test time chaos in e2e, I add a test program, which is mainly copied from #222


test/config.go
```diff
@@ -22,6 +22,7 @@ type Config struct {
 	ManagerTag   string
 	DaemonImage  string
 	DaemonTag    string
+	TimerImage   string
```
Yisaer, line 25: I tend to name it as E2eImage as we would put more binary to help e2e test in this image, not only timechaos.

yeya24: You mean we may add some other binaries to this image? Make sense!

Will update it tomorrow

yeya24:
Found a flaky test case https://github.com/pingcap/chaos-mesh/blob/master/pkg/ptrace/ptrace_linux_test.go here

hack/e2e.sh
```diff
@@ -153,6 +154,7 @@ function e2e::image_load() {
     for n in ${names[@]}; do
         $KIND_BIN load docker-image --name $CLUSTER ${DOCKER_REGISTRY}/$n:$IMAGE_TAG
     done
+    $KIND_BIN load docker-image --name $CLUSTER ${DOCKER_REGISTRY}/pingcap/e2e-helper:latest
```
Yisaer, line 157: we could add pingcap/e2e-helper into `names`

```sh
local names=(
        pingcap/chaos-mesh
        pingcap/chaos-daemon
)
```

Makefile
```diff
@@ -217,6 +217,9 @@ endif
 	cp -r manifests test/image/e2e
 	docker build -t "${DOCKER_REGISTRY_PREFIX}pingcap/chaos-mesh-e2e:${IMAGE_TAG}" test/image/e2e
 
+image-e2e-helper:
+	docker build -t "${DOCKER_REGISTRY_PREFIX}pingcap/e2e-helper:latest" test/cmd/e2e_helper
```
Yisaer, line 221: use ${IMAGE_TAG} here

hack/run-e2e.sh
```diff
@@ -80,6 +80,7 @@ e2e_args=(
     --manager-image-tag="${IMAGE_TAG}"
     --daemon-image="${DOCKER_REGISTRY}/pingcap/chaos-daemon"
     --daemon-image-tag="${IMAGE_TAG}"
+    --e2e-image="${DOCKER_REGISTRY}/pingcap/e2e-helper:latest"
```
Yisaer, line 83: use  ${DOCKER_REGISTRY}/pingcap/e2e-helper:${IMAGE_TAG}

hack/e2e.sh
yeya24: I put it below since I think maybe helper image just needs to use fixed `latest` tag. Do we need to use `${IMAGE_TAG}`? 
Do we need to push it to docker hub as well? If yes, I am happy to update the tag

Yisaer: Someday we would use e2e test framework as our stability env and we would push relative image to the internal hub.

yeya24:
Done. @Yisaer @cwen0  PTAL

yeya24:
Seems there is a problem in Github. Why this pr is not updated

yeya24:
OK. Got error in my added e2e. Will update it.

PR #442, opened 2020-04-22
organize the e2e case 

Yisaer:
organize the e2e case 

```release-note
NONE
```

### DIFF
--- a/test/e2e/chaos/basic.go
+++ b/test/e2e/chaos/basic.go
@@ -30,7 +30,7 @@ const (
 	pauseImage = "gcr.io/google-containers/pause:latest"
 )
 
-var _ = ginkgo.Describe("[chaos-mesh] Basic", func() {
+var _ = ginkgo.Describe("[Basic]", func() {
 	f := framework.NewDefaultFramework("chaos-mesh")
 	var ns string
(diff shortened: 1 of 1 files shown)


yeya24:
Is it better to put different chaos in different files? Instead of a single `basic.go`

Yisaer:
> Is it better to put different chaos in different files? Instead of a single `basic.go`

Currently, not necessary. I tend to make each file with a certain e2e environment.

test/e2e/chaos/basic.go
```diff
@@ -59,581 +59,596 @@ var _ = ginkgo.Describe("[chaos-mesh] Basic", func() {
...
-				Selector: v1alpha1.SelectorSpec{
-					Namespaces: []string{
-						ns,
+	// pod chaos case in [PodChaos] context
+	ginkgo.Context("[PodChaos]", func() {
+
+		// podfailure chaos case in [PodFailure] context
+		ginkgo.Context("[PodFailure]", func() {
+
+			ginkgo.It("[Duration]", func() {
```
yeya24, line 68: What's the difference between `Duration` case and `Schedule` case, seems they have almost the same test code, expect the chaos spec is a little different. Can we combine them together into one case?

PR #461, opened 2020-05-02
Validates the scheduler

Gallardot:

#373 

### Check List 

Tests 

 - Unit test

Code changes

 - Has Go code change

Side effects

 - None

Related changes

 - None

```release-note
NONE
```


api/v1alpha1/common_validation.go
```diff
@@ -34,37 +34,67 @@ const (
...
-		if err != nil {
-			allErrs = append(allErrs, field.Invalid(durationField,
-				*duration,
-				fmt.Sprintf("parse duration field error:%s", err)))
+	if duration != nil && scheduler != nil {
+		errs := validateSchedulerParams(duration, durationField, scheduler, schedulerField)
+		if len(errs) != 0 {
+			allErrs = append(allErrs, errs...)
 		}
+	} else if (duration == nil && scheduler != nil) || (duration != nil && scheduler == nil) {
```
cwen0, line 65: I think `duration==nil && scheduler != nil` is ok, sometimes we just want to start to chaos experiment at a specific time and last to the end time.

api/v1alpha1/common_validation.go
```diff
@@ -34,37 +34,67 @@ const (
...
 
-		if len(allErrs) > 0 {
-			return allErrs
+func validateSchedulerParams(duration *time.Duration, durationField *field.Path, spec *SchedulerSpec, schedulerField *field.Path) field.ErrorList {
+	allErrs := field.ErrorList{}
+	if duration != nil && spec != nil {
+		cronField := schedulerField.Child("cron")
+		allErrs = append(validateCron(spec.Cron, cronField))
+
+		scheduler, _ := cronv3.ParseStandard(spec.Cron)
```
cwen0, line 77: why ignore this error?

Gallardot: `validateCron` had validated the `cron`.

If ParseStandard  causes error,  the `scheduler` will be `nil`

So I think we can  ignore this error.

api/v1alpha1/common_validation.go
Gallardot: I am not sure about that.
But I find the controllers except podchaos all need  `scheduler == nil && duration == nil` OR `scheduler != nil && duration != nil`

https://github.com/pingcap/chaos-mesh/blob/8e769a704563d2becd5bd9269961783bca912952/controllers/iochaos/types.go#L47-L51


https://github.com/pingcap/chaos-mesh/blob/8e769a704563d2becd5bd9269961783bca912952/controllers/stresschaos/types.go#L59-L63

https://github.com/pingcap/chaos-mesh/blob/8e769a704563d2becd5bd9269961783bca912952/controllers/kernelchaos/types.go#L66-L70

https://github.com/pingcap/chaos-mesh/blob/8e769a704563d2becd5bd9269961783bca912952/controllers/timechaos/types.go#L58-L62


But podschaos has a different logic

https://github.com/pingcap/chaos-mesh/blob/8e769a704563d2becd5bd9269961783bca912952/controllers/podchaos/types.go#L48-L52



I think the main reason for this difference is that `PodKillAction` and `ContainerKillAction`  have different behaviors. We have addressed this difference in webhook of podschaos.

api/v1alpha1/common_validation.go
Yisaer: I don't suggest ignoring error here. The logic could be changed in future and we may miss the root cause when error happens.

Gallardot: I've already fixed it.  PTAL
@cwen0 @Yisaer

api/v1alpha1/common_validation.go
yeya24: This can be simplified to `else if !(duration == nil && scheduler == nil)`

Gallardot:
> I checked the e2e test and tested this PR in my local environment. Seems the continuous failure of e2e is caused by this PR itself. @Gallardot
> 
> Tested with the time chaos example.
> 
> ```yaml
> apiVersion: pingcap.com/v1alpha1
> kind: TimeChaos
> metadata:
>   name: time-shift-example
>   namespace: chaos-testing
> spec:
>   mode: one
>   selector:
>     labelSelectors:
>       app: io
>   timeOffset: "-1h"
>   duration: "9m"
>   scheduler:
>     cron: "@every 10m"
> ```
> 
> Seems the `nextRecover` is not correct.
> image

I'm terribly sorry about that. I’ll fix it right away. @yeya24

yeya24:
> > I checked the e2e test and tested this PR in my local environment. Seems the continuous failure of e2e is caused by this PR itself. @Gallardot
> > Tested with the time chaos example.
> > ```yaml
> > apiVersion: pingcap.com/v1alpha1
> > kind: TimeChaos
> > metadata:
> >   name: time-shift-example
> >   namespace: chaos-testing
> > spec:
> >   mode: one
> >   selector:
> >     labelSelectors:
> >       app: io
> >   timeOffset: "-1h"
> >   duration: "9m"
> >   scheduler:
> >     cron: "@every 10m"
> > ```
> > 
> > 
> > Seems the `nextRecover` is not correct.
> > image
> 
> I'm terribly sorry about that. I’ll fix it right away. @yeya24

No worries. Thanks for your awesome work! 🚀

Gallardot:
I have fix the continuous failure.
I will migrate `nextStart` and `nextRecover` to `status` in another PR

@yeya24 @cwen0 PTAL

PR #466, opened 2020-05-06
Support helm-less install

### LINKED ISSUE #431: helm-less install
## Feature Request

I'm always frustrated when installation is not documented clearly using k8s primitives.

Please document the installation and runtime requirements for chaos-mesh, so that environments without helm, and where helm has been decided against, can utilise chaos-mesh (or alternatively, explain what runtime services chaos-mesh requires from helm, so we can consider how to deliver those in high-trust environments.

I did consider ranting on Twitter, but it's probably counter-productive ;).

Clear minimal installation instructions will permit other packaging system for k8s to quickly package chaos-mesh up as well.

cwen0:

#431 

* remove helm dependency from install.sh 
* support helm-less install  

```
# install chaos mesh
curl -sSL https://raw.githubusercontent.com/cwen0/chaos-mesh/add_dry_run/install.sh | sh 

# install chaos-mesh with kubectl 
curl -sSL https://raw.githubusercontent.com/cwen0/chaos-mesh/add_dry_run/install.sh | sh -s -- --dry-run  | kubectl apply -f -  

# install kind cluster and chaos mesh
curl -sSL https://raw.githubusercontent.com/cwen0/chaos-mesh/add_dry_run/install.sh | sh -s -- -l kind
```

### Check List 

Tests 


install.sh
```diff
@@ -15,26 +15,26 @@ USAGE:
...
         --volume-provisioner Deploy volume provisioner in local Kubernetes cluster
         --local-registry     Deploy local docker registry in local Kubernetes cluster
+        --dry-run            Simulate an install and generate chaos-mesh manifests files
 OPTIONS:
     -v, --version            Version of chaos-mesh, default value: latest
     -l, --local [kind]       Choose a way to run a local kubernetes cluster, supported value: kind,
                              If this value is not set and the Kubernetes is not installed, this script will exit with 1.
     -n, --name               Name of Kubernetes cluster, default value: kind
+    -c  --crd                The URL of the crd files, default value: https://raw.githubusercontent.com/pingcap/chaos-mesh/master/manifests/crd.yaml
+    -r  --runc               Runtime specifies which container runtime to use. Currently we only supports docker and containerd. default value: docker
```
yeya24, line 33: What about `--runtime`, `runc` is not that appropriate here

install.sh
```diff
@@ -15,26 +15,26 @@ USAGE:
...
         --docker-mirror      Use docker mirror to pull image, dockerhub.azk8s.cn => docker.io, gcr.azk8s.cn => gcr.io
         --volume-provisioner Deploy volume provisioner in local Kubernetes cluster
         --local-registry     Deploy local docker registry in local Kubernetes cluster
+        --dry-run            Simulate an install and generate chaos-mesh manifests files
 OPTIONS:
     -v, --version            Version of chaos-mesh, default value: latest
     -l, --local [kind]       Choose a way to run a local kubernetes cluster, supported value: kind,
                              If this value is not set and the Kubernetes is not installed, this script will exit with 1.
     -n, --name               Name of Kubernetes cluster, default value: kind
+    -c  --crd                The URL of the crd files, default value: https://raw.githubusercontent.com/pingcap/chaos-mesh/master/manifests/crd.yaml
```
yeya24, line 32: What is the use case that we need to expose this flag? Does it need to be configurable?

install.sh
```diff
@@ -971,4 +731,580 @@ azk8spull() {
...
+
+    ensure openssl req -new -key ${tmpdir}/server.key -out ${tmpdir}/server.csr -config ${tmpdir}/csr.conf > /dev/null 2>&1
+    ensure openssl x509 -req -in ${tmpdir}/server.csr -CA ${tmpdir}/ca.crt -CAkey ${tmpdir}/ca.key -CAcreateserial -out ${tmpdir}/server.crt -days 1875 -extensions v3_req -extfile ${tmpdir}/csr.conf > /dev/null 2>&1
+
+    TLS_KEY=$(openssl base64 -A -in ${tmpdir}/server.key)
+    TLS_CRT=$(openssl base64 -A -in ${tmpdir}/server.crt)
+    CA_BUNDLE=$(openssl base64 -A -in ${tmpdir}/ca.crt)
+
+    cat <<EOF
+---
```
yeya24, line 787: Is it possible to not copy all these yaml templates in this script? It would be a pain that we have to update both this script and the helm templates when there is a change.

install.sh
```diff
@@ -971,4 +731,580 @@ azk8spull() {
...
+      containers:
+        - name: chaos-daemon
+          image: pingcap/chaos-daemon:latest
+          imagePullPolicy: Always
+          command:
+            - /usr/local/bin/chaos-daemon
+            - --runtime
+            - ${runc}
+            - --http-port
+            - !!str 31766
```
yeya24, line 964: `!!str` is not needed

install.sh
```diff
@@ -971,4 +731,580 @@ azk8spull() {
...
+            - --grpc-port
+            - !!str 31767
```
yeya24, line 966: `!!str` is not needed

install.sh
yeya24: Using a relative path is enough here?

install.sh
Gallardot: agree

install.sh
cwen0: If we use a `relative path`, we must use exec this script in specific directory.

install.sh
cwen0: I just want to provide a more convenient installation method and do not rely on helm, Do you have any other good solutions?

install.sh
yeya24: Yes, but we have to. It is not appropriate to use `/master` in such script. If we want to run this script in the release branch, then what is the value of `crd`? Do we have to update to `https://raw.githubusercontent.com/pingcap/chaos-mesh/<tag>/manifests/crd.yaml` every time?

cwen0: When we need to release, we also need to change the image to install, so I think we also need to update this, or we add a flag to define the version of chaos-mesh.

yeya24: I still think it is better to use `relative path`. If using `master`, then how to test each PR in CI? We have to update the path to each PR's CRD?

install.sh
cwen0: Maybe we can add a hack script to help use us update this.

install.sh
cwen0: If we delete `!!str`, the program will report an error. 
```
Error from server (BadRequest): error when creating "STDIN": DaemonSet in version "v1" cannot be handled as a DaemonSet: v1.DaemonSet.Spec: v1.DaemonSetSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Command: []string: ReadString: expects " or n, but found 3, error found in #10 byte of ...|tp-port",31766,"--gr|..., bigger context ...|s-daemon","--runtime","containerd","--http-port",31766,"--grpc-port",31767],"image":"pingcap/chaos-d|...
```

install.sh
cwen0: this pr has no relation to CI. Why need to use this pr in CI?

install.sh
Gallardot: Maybe a hack script can be executed  after `make yaml` and checked by CI

install.sh
yeya24: Sorry I mean not CI, just in each branch. If we are in different branches and run this script, then we have to still update the crd url every time? I don't think it is reasonable

cwen0: Maybe we can support to set a `relative path` and use the URL as a default value is OK.

cwen0: Done

install.sh
cwen0: Done

cwen0:
@Gallardot @yeya24 @fewdan PTAL again! Thanks!

Gallardot:
who and when need to execute the ‘make update-install-script’ ?

cwen0:
> who and when need to execute the ‘make update-install-script’ ?

when we update helm charts, we need run `make update-install-script` to update `install.sh`. `helm --dry-run` requires a Kubernetes environment, so we can't add it at CI at present.

cwen0:
Close #431

PR #472, opened 2020-05-08
make chaos-server runnable

cwen0:

* make chaos-server runnable 
* init store pkg and define some struct 
* fix `make check` 

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


pkg/core/archive_experiment.go
```diff
@@ -14,13 +14,56 @@
...
 // ExperimentStore defines operations for working with archive experiments
-type ExperimentStore interface{}
+type ExperimentStore interface {
+	// List returns a archive experiment list from the datastore.
+	List(context.Context) ([]*ArchiveExperiment, error)
+
+	// ListByKind returns a archive experiment list by kind from the datastore.
+	ListByKind(context.Context, string) ([]*ArchiveExperiment, error)
+
+	// Find returns a archive experiment by ID.
```
fewdan, line 33: What is "ID"?  Do we assign a number to each archive?  
s/a archive/an archive/g

pkg/core/archive_experiment.go
```diff
@@ -14,13 +14,56 @@
...
+	FindByName(context.Context, string, string) (*ArchiveExperiment, error)
+
+	// Create persists a new archive experiment to the datastore.
+	Create(context.Context, *ArchiveExperiment) error
+
+	// Update persists an updated experiment to the datastore.
+	Update(context.Context, *ArchiveExperiment) error
+}
 
 // ArchiveExperiment represents a experiment instance.
```
fewdan, line 46: s/a experiment/an experiment/g

pkg/core/archive_experiment.go
```diff
@@ -14,13 +14,56 @@
...
+	List(context.Context) ([]*ArchiveExperiment, error)
+
+	// ListByKind returns a archive experiment list by kind from the datastore.
+	ListByKind(context.Context, string) ([]*ArchiveExperiment, error)
+
+	// Find returns a archive experiment by ID.
+	Find(context.Context, int64) (*ArchiveExperiment, error)
+
+	// FindByName returns a archive experiment by the name and namespace of the experiment.
+	FindByName(context.Context, string, string) (*ArchiveExperiment, error)
```
fewdan, line 37: Here I have a question, that is, every time we delete the experiment, and then archive the experiment to db. If under the same ns, create another experiment with the previously used name. When finally archiving again, there will be multiple experiments with the same name and the same ns in the db, then this interface should return a list?

pkg/core/archive_experiment.go
cwen0: If we store an experiment in DB, the id is necessary. It may be auto-increment.

pkg/core/archive_experiment.go
cwen0: I think we can generate a suffix for each experiment autosomally.

pkg/core/archive_experiment.go
```diff
@@ -14,13 +14,56 @@
...
 
-// ArchiveExperiment represents a experiment instance.
+	// Update persists an updated experiment to the datastore.
+	Update(context.Context, *ArchiveExperiment) error
+}
+
+// ArchiveExperiment represents an experiment instance.
 type ArchiveExperiment struct {
+	ArchiveExperimentMeta
+	Experiment string `gorm:"size:2048"`
```
yeya24, line 49: What is this field? Does it store the chaos yaml or some others?

pkg/core/pod_record.go
```diff
@@ -0,0 +1,50 @@
...
+
+package core
+
+import (
+	"context"
+
+	"github.com/jinzhu/gorm"
+)
+
+// ExperimentStore defines operations for working with pod records
```
yeya24, line 22: This should be `PodRecordStore`

pkg/apiserver/experiment/experiment.go
```diff
@@ -30,7 +30,7 @@ type Service struct {
 	event   core.EventStore
 }
 
-// NewService return a experiment service instance.
+// NewService return an experiment service instance.
```
yeya24, line 33: `returns`

pkg/store/experiment/experiment.go
```diff
@@ -25,7 +27,20 @@ func NewStore(db *dbstore.DB) core.ExperimentStore {
 	return &experimentStore{db}
 }
 
-// TODO: need to define the fields of this struct
 type experimentStore struct {
 	db *dbstore.DB
 }
+
+// TODO: implement core.EventStore interface
```
yeya24, line 34: I think we need a `delete` interface here as well. Do you want to put it in later prs?

pkg/core/archive_experiment.go
fewdan: In other words, the name of each experiment is suffixed to ensure uniqueness?

pkg/core/archive_experiment.go
cwen0: This field will store the whole experiment struct.

pkg/core/archive_experiment.go
cwen0: I think we can add the suffix when creating chaos.

PR #479, opened 2020-05-09
Fix pause resume logic

### LINKED ISSUE #450: Experiment should be resumed immediately after not pausing
## Bug Report

**What did you do?**

Create a chaos with the config below. We have cron 10m and duration 9m.

``` yaml
kind: IoChaos
metadata:
  name: io-delay-example
  namespace: chaos-testing
spec:
  action: delay
  mode: one
  configName: "chaosfs-test"
  path: ""
  selector:
    labelSelectors:
      app: test
  layer: "fs"
  delay: "1s"
  duration: "9m"
  scheduler:
    cron: "@every 10m"
```

After 10s the chaos started, I paused this chaos. Then 10s later, I set paused to false. I expected the chaos continue running. However, this chaos is still paused until its nextStart, which is 10m later. 

After the chaos resumes, we need to still inject the chaos in the current round, instead of waiting until the next round

yeya24:

Fixes #450 

Manual tested and it works.

will add E2E test case for this later.

### DIFF
--- a/controllers/twophase/types.go
+++ b/controllers/twophase/types.go
@@ -19,14 +19,13 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	"k8s.io/client-go/util/retry"
 
 	"github.com/pingcap/chaos-mesh/api/v1alpha1"
 	"github.com/pingcap/chaos-mesh/controllers/reconciler"
 	"github.com/pingcap/chaos-mesh/pkg/utils"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/client-go/util/retry"
-
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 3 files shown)


controllers/twophase/types.go
```diff
@@ -111,17 +109,27 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 
 		status.Experiment.EndTime = &metav1.Time{
 			Time: time.Now(),
 		}
 		status.Experiment.Phase = v1alpha1.ExperimentPhaseFinished
+	} else if status.Experiment.Phase == v1alpha1.ExperimentPhasePaused &&
+		!chaos.GetNextRecover().IsZero() && chaos.GetNextRecover().After(now) {
+		// Resume
+		r.Log.Info("Resuming")
+		if err := applyAction(ctx, r, req, chaos); err != nil {
```
cwen0, line 130: I don't understand this logic. The phase is `Paused` and why needs to apply here?

yeya24: This is the time that we want to resume the chaos, which means changing the `Paused` status to `Running` again, so we need to apply the action and update phase to `Running`. Does this make sense to you?

PR #499, opened 2020-05-12
basic api

fewdan:
[basic api](https://github.com/pingcap/chaos-mesh/issues/483)

add common api

### Check List 

Tests 

 - No code

Code changes

 - Has Go code change

Side effects

 - NO

Related changes

 - NO

```release-note
NONE
```


cwen0:
@fewdan remove core.sqlite file and add it to .gitignore file

pkg/apiserver/common/common.go
```diff
@@ -0,0 +1,155 @@
...
+
+// GetKinds returns the list of chaos kinds
+func (s *Service) GetKinds(c *gin.Context) {
+	ChaosKindList :=  make([]ChaosKind, 0)
+
+	ChaosKindList = append(ChaosKindList, ChaosKind{
+		Name: "PodChaos",
+	})
+
+	ChaosKindList = append(ChaosKindList, ChaosKind{
```
cwen0, line 138: I think we can get these crd types from Kubernetes.

fewdan: Obtained from K8s, if you need to add fields to the structure of chaoskind in the future, it may not be very convenient. For example, add a description field in chaoskind to describe the meaning of this type of chaos.

fewdan: Obtained from K8s, if you need to add fields to the structure of chaoskind in the future, it may not be very convenient. For example, add a description field in chaoskind to describe the meaning of this type of chaos.

fewdan: already edited

pkg/apiserver/common/common.go
```diff
@@ -0,0 +1,164 @@
...
+
+	if err != nil {
+		c.JSON(http.StatusOK, gin.H{
+			"status": apiserver.GetResourcesWrong,
+			"message": "get namespaces wrong",
+			"data": namespaceList,
+		})
+		return
+	}
+
```
yeya24, line 122: We can move `namespaceList :=  make([]Namespace, 0, len(namespace.Items))` here since we know the number of namespaces

pkg/apiserver/common/common.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// GetKinds returns the list of chaos kinds
+func (s *Service) GetKinds(c *gin.Context) {
+	ChaosKindList :=  make([]ChaosKind, 0)
+
+	config, _ := ctrlconfig.GetConfig()
+	apiExtCli, _ := apiextensionsclientset.NewForConfig(config)
+
+	crdList, err := apiExtCli.ApiextensionsV1beta1().CustomResourceDefinitions().List(metav1.ListOptions{})
```
yeya24, line 143: This is good. But we can add a list option to filter `group:pingcap.com` if we want to optimize it further

pkg/apiserver/common/common.go
```diff
@@ -0,0 +1,164 @@
...
+			})
+		}
+	}
+
+	c.JSON(http.StatusOK, gin.H{
+		"status": apiserver.Success,
+		"message": "success",
+		"data": ChaosKindList,
+	})
+}
```
yeya24, line 164: Please add a new line

helm/chaos-mesh/values.yaml
```diff
@@ -21,7 +21,7 @@ controllerManager:
   replicaCount: 1
 
   image: pingcap/chaos-mesh:latest
-  imagePullPolicy: Always
+  imagePullPolicy: Never
```
zhouqiang-cl, line 24: why change to Never

fewdan: It was temporarily modified during the test and I forgot to change it back.

Issue #503, opened 2020-05-12
New info API for chaosfs sidecar


yeya24:
## Feature Request

The current chaosfs sidecar is just like a black box and it is hard to know its current status. The only way I can come up with is to check its logs. However, this is hard to achieve in E2E test.

Now many flaky IO E2E tests happen because it is hard to know whether the chaosfs recovers the fault or not. It would be nice to have such an `Info` grpc API to get the current status of injections.

Yisaer:
Good idea. We can also sync the sidecar information into IOChaos Status.

yeya24:
IMO the IO chaos status should be separated from the sidecar status. They are two different entities.

yeya24:
Actually what I need is just an API to check whether the chaosfs is injected or not and I don't care about the details of injections. So this API only returns a bool value. Is it reasonable? @cwen0 @Yisaer WDYT?

Yisaer:
> Actually what I need is just an API to check whether the chaosfs is injected or not and I don't care about the details of injections. So this API only returns a bool value. Is it reasonable? @cwen0 @Yisaer WDYT?

That's ok to me.

PR #504, opened 2020-05-13
Remove set clean start to true

yeya24:

This flag is used for cleanup all namespaces expect the system ns such as `kube-system`, `kube-public`, etc.

Though this is useful, it is quite dangerous and error-prone. Especially in the case that you need to use different context to switch between the local e2e k8s and other k8s clusters...


yeya24:
Sorry for the flaky again. We need #503.

Issue #482, opened 2020-05-09
Record chaos events to dbstore


fewdan:
## Feature Request

In order to complete chaos-server, we need to store relevant information. We need to watch all chaos object and record chaos events to DB store

PR #507, opened 2020-05-13
Record chaos event to dbstore

cwen0:

#482 

In this pr, we will watch all chaos object and record chaos events to DB store.

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


cwen0:
> Is the information currently stored sufficient for API use? Should pod-related information be added to the event? There is an interface to get the event on the specified pod, that is to say we want to get the event by pod name and pod NS, so should we add pod information to the event?

Updated

cwen0:
@yeya24 If we delete this chaos, the `Experiment.Phase` will be set to `ExperimentPhaseFinishedCWen` at here https://github.com/pingcap/chaos-mesh/blob/master/controllers/common/types.go#L78, So the `finish_time` won't be empty. In addition, we will delete all incomplete events when to start chaos-server.

yeya24:
I will merge it when CI passes if you think it is okay. @cwen0 @fewdan

Issue #484, opened 2020-05-09
[API] Get the state of current experiments


fewdan:
## Feature Request

/experiments/state: Get the state of current experiments

See [here](https://app.swaggerhub.com/apis/ChaosMesh/chaos-server/1.0.0#trial) for details

PR #512, opened 2020-05-14
[API] experiments/state

fewdan:
add experiments/state [#484](https://github.com/pingcap/chaos-mesh/issues/484)

add experiments/stat

### Check List 

Tests 

 - No code

Code changes

 - Has Go code change

Side effects

 - NO

Related changes

 - NO

```release-note
NONE
```


pkg/apiserver/experiment/experiment.go
```diff
@@ -63,3 +67,192 @@ func (s *Service) getExperimentDetail(c *gin.Context) {}
...
+	data["Finished"] = 0
+	getChaosWrong := gin.H{
+		"status": 1003,
+		"message": "get chaos wrong",
+		"data": make(map[string]int),
+	}
+
+	err := s.getPodChaosState(data)
+	if err != nil {
+		c.JSON(200, getChaosWrong)
```
Yisaer, line 224: use `http.StatusOK` instead 200

fewdan: done

pkg/apiserver/experiment/experiment.go
```diff
@@ -47,12 +51,13 @@ func NewService(
 
 // Register mounts our HTTP handler on the mux.
 func Register(r *gin.RouterGroup, s *Service) {
-	endpoint := r.Group("/experiment")
+	endpoint := r.Group("/experiments")
 
 	// TODO: add more api handlers
 	endpoint.GET("/", s.listExperiments)
-	endpoint.GET("/:name", s.getExperimentDetail)
+	endpoint.GET("/detail/:name", s.getExperimentDetail)
```
yeya24, line 58: IMO this API is not RESTful. And do we also need `ns` in this API?

pkg/apiserver/experiment/experiment.go
```diff
@@ -63,3 +68,192 @@ func (s *Service) getExperimentDetail(c *gin.Context) {}
...
+func (s *Service) state (c *gin.Context) {
+	data := make(map[string]int)
+	data["Total"] = 0
+	data["Running"] = 0
+	data["Paused"] = 0
+	data["Failed"] = 0
+	data["Finished"] = 0
+	getChaosWrong := gin.H{
+		"status": 1003,
+		"message": "get chaos wrong",
```
yeya24, line 219: What about `failed to get chaos state`?

pkg/apiserver/experiment/experiment.go
```diff
@@ -63,3 +68,192 @@ func (s *Service) getExperimentDetail(c *gin.Context) {}
...
+	err = s.getKernelChaosState(data)
+	if err != nil {
+		c.JSON(http.StatusOK, getChaosWrong)
+		return
+	}
+	err = s.getStressChaosState(data)
+	if err != nil {
+		c.JSON(http.StatusOK, getChaosWrong)
+		return
+	}
```
yeya24, line 252: I think we can optimize these `getChaosState` functions. IMO what about adding a new interface for these different chaos lists? Then we can just use one function like `s.getChaosState(data, chaosList <newInterfaceName>)`. And in this way, we can also use errGroup to make the query parallel.

pkg/apiserver/experiment/experiment.go
fewdan: I think ns is needed, and this can be added when the API is developed.

pkg/apiserver/experiment/experiment.go
yeya24: But IMO this is okay to put in next PR.

pkg/apiserver/experiment/experiment.go
```diff
@@ -63,3 +68,192 @@ func (s *Service) getExperimentDetail(c *gin.Context) {}
...
+		switch chaos.Status.ChaosStatus.Experiment.Phase {
+		case v1alpha1.ExperimentPhaseRunning:
+			stateInfo["running"]++
+		case v1alpha1.ExperimentPhasePaused:
+			stateInfo["Paused"]++
+		case v1alpha1.ExperimentPhaseFailed:
+			stateInfo["Failed"]++
+		case v1alpha1.ExperimentPhaseFinished:
+			stateInfo["Finished"]++
+		}
```
yeya24, line 90: This switch statement is not needed I guess.
We can simply use 

``` go
for _, chaos := range chaosList.Items {
  stateInfo[string(chaos.Status.ChaosStatus.Experiment.Phase)]++
}
stateInfo["Total"] += len(chaosList.Items)
```

fewdan: Ok, I will change it.

pkg/apiserver/experiment/experiment.go
```diff
@@ -63,3 +68,192 @@ func (s *Service) getExperimentDetail(c *gin.Context) {}
...
+
+func (s *Service) state (c *gin.Context) {
+	data := make(map[string]int)
+	data["Total"] = 0
+	data["Running"] = 0
+	data["Paused"] = 0
+	data["Failed"] = 0
+	data["Finished"] = 0
+	getChaosWrong := gin.H{
+		"status": 1003,
```
yeya24, line 218: Another question, do we have some definition docs for the status code? It is hard for me to know what this means. What about making these status codes constants?

fewdan: I will define errors_code.go separately in apiserver.

cwen0:
@fewdan Please fix CI, first!

pkg/apiserver/experiment/experiment.go
```diff
@@ -63,3 +69,184 @@ func (s *Service) getExperimentDetail(c *gin.Context) {}
...
+	var chaosList v1alpha1.PodChaosList
+	err := s.kubeCli.List(context.Background(), &chaosList)
+	if err != nil {
+		return err
+	}
+	for _, chaos := range chaosList.Items {
+		stateInfo["Total"]++
+		switch chaos.Status.ChaosStatus.Experiment.Phase {
+		case v1alpha1.ExperimentPhaseRunning:
+			stateInfo["running"]++
```
zhouqiang-cl, line 84: s/runnig/Running?

zhouqiang-cl: And did we need to define Running = "Running"?

zhouqiang-cl: Did we need a state like "New"?

pkg/apiserver/experiment/experiment.go
```diff
@@ -63,3 +69,184 @@ func (s *Service) getExperimentDetail(c *gin.Context) {}
...
+			stateInfo["Failed"]++
+		case v1alpha1.ExperimentPhaseFinished:
+			stateInfo["Finished"]++
+		}
+	}
+	return nil
+}
+
+// getIoChaosState returns the state of IoChaos
+func (s *Service) getIoChaosState (stateInfo map[string]int) error {
```
zhouqiang-cl, line 97: Can we wrapper this function, there are many duplicate code here, such as in IOChaos/NetworkChaos

pkg/apiserver/experiment/experiment.go
fewdan: It has been modified and there is currently no "New" phase.

PR #513, opened 2020-05-14
Add injected grpc method

### LINKED ISSUE #503: New info API for chaosfs sidecar

yeya24:

Fixes #503 

Added a new grpc method `Injected` to return whether chaosfs sidecar is injected or not.

Checklist:
- [x] code updated
- [x] proto file updated
- [x] tests added
- [x] e2e updated


cwen0:
@fewdan @Yisaer PTAL

PR #517, opened 2020-05-15
Use annotation to pause chaos

### LINKED ISSUE #447: Use annotation of Chaos Object to pause experiment
## Feature Request

At present, If we want to pause an experiment, we need patch this object, such as: 

```
kubectl patch podchaos pod-kill-example --namespace chaos-testing --type merge --patch 'spec:
 paused: false'
``` 
I think this operation is complicated and incomprehensible. 

I recommend that we can use annotation of the object to pause experiment, in the way, we can pause the experiment using the following command: 

```
kubectl annotate podchaos pod-kill-example -n chaos-testing pause='true'
```

yeya24:

Fixes #447

### DIFF
--- a/api/v1alpha1/common_types.go
+++ b/api/v1alpha1/common_types.go
@@ -21,6 +21,8 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
+const PauseAnnotationKey = "admission-webhook.pingcap.com/pause"
+
 // SelectorSpec defines the some selectors to select objects.
 // If the all selectors are empty, all objects will be used in chaos experiment.
 type SelectorSpec struct {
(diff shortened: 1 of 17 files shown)


api/v1alpha1/common_types.go
```diff
@@ -21,6 +21,8 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
+const PauseAnnotationKey = "admission-webhook.pingcap.com/pause"
```
cwen0, line 24: This feature isn't related to `admission-webhook`, I think we can use the other key, eg: "experiment.pingcap.com/pause` or others.

yeya24: Make sense.

PR #519, opened 2020-05-17
Add a script to generate go code from proto files

yeya24:

Add a script to generate go code from proto files.
We can easily run it via `make proto`


cwen0:
@fewdan PTAL

PR #523, opened 2020-05-18
Implement list experiments

### LINKED ISSUE #492: [API] Get the list of the experiments
## Feature Request

/experiments/all: Get the list of the experiments

See [here](https://app.swaggerhub.com/apis/ChaosMesh/chaos-server/1.0.0#trial) for details

yeya24:

Fixes #492.

There is also some refactoring in this PR.

Changes:

- [x] Refactor and clean up some code in chaos server.
- [x] Add a common interface for chaosList
- [x] Remove all the `/all` endpoint in the urls
- [x] Return plain string array in common APIs

Verification

```
curl 'localhost:2333/api/experiments?namespace=chaos-testing&&kind=PodChaos' | jq .
{
  "data": [
    {
      "name": "pod-failure-example",
      "namespace": "chaos-testing",
      "kind": "PodChaos",
      "created": "2020-05-18T14:48:06-04:00",
      "status": "Running"
    },
    {
      "name": "nginx-failure",
      "namespace": "chaos-testing",
      "kind": "PodChaos",
      "created": "2020-05-18T14:50:40-04:00",
      "status": "Running"
    }
  ],
  "message": "success",
  "status": 0
```


pkg/apiserver/experiment/experiment.go
```diff
@@ -16,17 +16,40 @@ package experiment
...
+		v1alpha1.KindNetworkChaos: &v1alpha1.NetworkChaosList{},
+		v1alpha1.KindStressChaos:  &v1alpha1.StressChaosList{},
+		v1alpha1.KindIOChaos:      &v1alpha1.IoChaosList{},
+		v1alpha1.KindKernelChaos:  &v1alpha1.KernelChaosList{},
+		v1alpha1.KindTimeChaos:    &v1alpha1.TimeChaosList{},
+	}
+)
+
+// Experiment defines the basic information of an experiment
+type Experiment struct {
```
cwen0, line 45: Maybe we also need to add the event field in this struct and store the recent events 

image

pkg/apiserver/experiment/experiment.go
```diff
@@ -16,17 +16,40 @@ package experiment
...
 	"github.com/pingcap/chaos-mesh/pkg/config"
 	"github.com/pingcap/chaos-mesh/pkg/core"
+	"golang.org/x/sync/errgroup"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+var (
+	// TODO(yeya24): don't hardcode this if it is possible
+	kinds = map[string]v1alpha1.ChaosList{
```
cwen0, line 34: we can register these with `init` function in `v1alpha1` pkg.

yeya24: In the v1aloha1 package? Then this map will be public, which is not very appropriate I think. It is error-prone and maybe changed in the runtime. What’s your idea?

pkg/apiserver/experiment/experiment.go
yeya24: From this document https://app.swaggerhub.com/apis/ChaosMesh/chaos-server/1.0.0#/experiment/get_experiments_all I don't see the event field. What about updating the APIs first?

Do you think getting the events of one experiment should be part of the experiments list API, or a separate API?

pkg/apiserver/experiment/experiment.go
cwen0: It can reduce the cost of adding new chaos. I think it is also OK to do so at present like your pr.

pkg/apiserver/experiment/experiment.go
cwen0: At here, we just need the recent events, such as a day, If we use a separate API, the page will pull too much request to chaos-server at the same time.

cwen0: We also can update this in other pr.

yeya24: Then I think two optional parameters `startTime and endTime` is needed for filtering in the frontend.

pkg/apiserver/experiment/experiment.go
yeya24: You are right, we should put it in a commonplace and use init(). But I am not sure in `v1alpha1` package or in `apiserver`.

pkg/apiserver/experiment/experiment.go
```diff
@@ -54,187 +88,104 @@ func Register(r *gin.RouterGroup, s *Service) {
...
+			for _, chaos := range list.ListChaos() {
+				switch chaos.Status {
+				case string(v1alpha1.ExperimentPhaseRunning):
+					data.Running++
+				case string(v1alpha1.ExperimentPhasePaused):
+					data.Paused++
+				case string(v1alpha1.ExperimentPhaseFailed):
+					data.Failed++
+				case string(v1alpha1.ExperimentPhaseFinished):
+					data.Finished++
```
yeya24, line 178: Seems swaggo doesn't support to return a map? So I just use a struct to keep the number of each state. 

```
@Success 200 {object} ChaosState
```

PR #524, opened 2020-05-18
Support experiment deletion

### LINKED ISSUE #522: [API] Delete the running experiment
## Feature Request

`DELETE /experiments/:kind/:ns/:name`

Delete a current experiment specified by its namespace and name.

yeya24:

Fixes #522


pkg/apiserver/experiment/experiment.go
```diff
@@ -67,11 +71,50 @@ func (s *Service) listExperiments(c *gin.Context) {}
...
+			fmt.Sprintf("kind %s is not supported", kind))
+		return
+	}
+	if err := s.kubeCli.Get(ctx, chaosKey, chaos); err != nil {
+		util.JSONEmptyData(c, statuscode.GetResourcesWrong,
+			"failed to get chaos; "+err.Error())
+		return
+	}
+
+	if err := s.kubeCli.Delete(ctx, chaos, &client.DeleteOptions{}); err != nil {
```
fewdan, line 107: Should we also need to delete the experiment in the DB? Or do you want to archive this experiment in db after deleting the experiment?

yeya24: I think deleting the current experiment in the cluster and deleting experiments in the databases should be 2 APIs because only archived experiments are in the database.

But I am not very sure. Maybe @cwen0 has some other ideas. WDYT?

fewdan: Oh, I mean, after deleting the experiment from the cluster, we need to archive the experiment to the db.

yeya24: Sorry I misunderstand your question before. But in another PR I already implemented the archive functionality. So when one experiment is deleted it will be added to dB immediately

pkg/apiserver/experiment/experiment.go
```diff
@@ -67,11 +71,50 @@ func (s *Service) listExperiments(c *gin.Context) {}
...
+	chaosKey := types.NamespacedName{Namespace: c.Param("ns"), Name: c.Param("name")}
+
+	kindsMap := map[string]runtime.Object{
+		v1alpha1.KindPodChaos:     &v1alpha1.PodChaos{},
+		v1alpha1.KindNetworkChaos: &v1alpha1.NetworkChaos{},
+		v1alpha1.KindStressChaos:  &v1alpha1.StressChaos{},
+		v1alpha1.KindIOChaos:      &v1alpha1.IoChaos{},
+		v1alpha1.KindKernelChaos:  &v1alpha1.KernelChaos{},
+		v1alpha1.KindTimeChaos:    &v1alpha1.TimeChaos{},
+	}
```
Yisaer, line 91: Can we make this as a const variable?

pkg/apiserver/experiment/experiment.go
Yisaer, line 107: I think we need to handle the case that the error is not found. if error is not found, I think we can return success with a business error code?

yeya24: You are right! Forget this case

pkg/apiserver/experiment/experiment.go
yeya24: Const map is not possible. I will update it to use a var.

api/v1alpha1/kinds.go
```diff
@@ -0,0 +1,38 @@
...
+	KindKernelChaos = "KernelChaos"
+	// KindStressChaos is the kind for stress chaos
+	KindStressChaos = "StressChaos"
+	// KindTimeChaos is the kind for time chaos
+	KindTimeChaos = "TimeChaos"
+)
+
+var (
+	// Kinds is a collection to keep chaos kind name and its type
+	Kinds = map[string]*ChaosKind{
```
yeya24, line 22: I don't if this is okay. But it is very dangerous since var is changeable during runtime.

pkg/apiserver/experiment/experiment.go
```diff
@@ -144,8 +135,64 @@ func (s *Service) listExperiments(c *gin.Context) {
...
+// @Produce json
+// @Param namespace path string true "namespace"
+// @Param name path string true "name"
+// @Param kind query string true "kind" Enums(PodChaos, IoChaos, NetworkChaos, TimeChaos, KernelChaos, StressChaos)
+// @Success 200 "delete ok"
+// @Router /api/experiments/{namespace}/{name} [delete]
+// @Failure 400 {object} utils.APIError
+// @Failure 500 {object} utils.APIError
+func (s *Service) deleteExperiment(c *gin.Context) {
+	kind := c.Query("kind")
```
cwen0, line 150: Why not add this to `Param' ?

pkg/apiserver/experiment/experiment.go
```diff
@@ -144,8 +135,64 @@ func (s *Service) listExperiments(c *gin.Context) {
...
+		chaosKind *v1alpha1.ChaosKind
+		ok        bool
+	)
+	if chaosKind, ok = v1alpha1.Kinds[kind]; !ok {
+		c.Status(http.StatusBadRequest)
+		_ = c.Error(utils.ErrInvalidRequest.New(kind + " is not supported"))
+		return
+	}
+	if err := s.kubeCli.Get(ctx, chaosKey, chaosKind.Chaos); err != nil {
+		if apierrors.IsNotFound(err) {
```
cwen0, line 171:
```
c.Status(http.StatusNotFound)
_ = c.Error()
```

api/v1alpha1/kinds.go
cwen0: If we use this way, why not register these values in different files.

pkg/apiserver/experiment/experiment.go
yeya24: I am not sure about the pattern. The URL might be too long.

pkg/apiserver/experiment/experiment.go
yeya24: Regarding this comment https://github.com/pingcap/chaos-mesh/pull/524#discussion_r428105456. WDYT? Do you want to treat not found case as an error?

yeya24:
> ```
> If we use this way, why not register these values in different files.
> ```

@cwen0 I don't have a good idea about it. And I think the current way is not very good... Do you have more suggestions?

pkg/apiserver/experiment/experiment.go
cwen0: They are the same variables, so I think we need to unify these.

pkg/apiserver/experiment/experiment.go
cwen0: Yes, I think so.

pkg/apiserver/experiment/experiment.go
yeya24: Do you prefer `/:ns/:name/:kind` or `/:kind/:ns/:name`?

cwen0: `/:kind/:ns/:name`?

yeya24:
E2E failed but it seems this is the flaky test we met before. And it is not related to this PR. So I will merge this PR manually now.

PR #526, opened 2020-05-19
update store/event/event.go

fewdan:
In order to implement event-related interfaces, store/event/event.go was completed.

update store/event/event.go

### Check List 

Tests 

 - No code

Code changes

 - Has Go code change

Side effects

 - NO

Related changes

 - NO

```release-note
NONE
```


pkg/apiserver/event/event.go
```diff
@@ -47,11 +51,45 @@ func NewService(
 
 // Register mounts our HTTP handler on the mux.
 func Register(r *gin.RouterGroup, s *Service) {
-	endpoint := r.Group("/event")
+	endpoint := r.Group("/events")
 
 	// TODO: add more api handlers
 	endpoint.GET("/all", s.listEvents)
```
yeya24, line 57: We can use "endpoint.GET("l", s.listEvents)" here if this makes sense to you

pkg/store/event/event.go
```diff
@@ -44,15 +44,114 @@ type eventStore struct {
...
+		First(et).Error; err != nil && !gorm.IsRecordNotFoundError(err) {
+		return nil, err
+	}
+	pods, err:= e.FindPodsByEventID(context.Background(), et.ID)
+	if err != nil {
+		return nil, err
+	}
+	for _, pod := range pods {
+		et.Pods = append(et.Pods, pod)
+	}
```
yeya24, line 152: Seems you can directly use `et.Pods = pods`

pkg/store/event/event.go
```diff
@@ -44,15 +44,114 @@ type eventStore struct {
...
+			return nil, err
+		}
+
+		pods, err:= e.FindPodsByEventID(context.Background(), et.ID)
+		if err != nil {
+			return nil, err
+		}
+		for _, pod := range pods {
+			et.Pods = append(et.Pods, pod)
+		}
```
yeya24, line 132: ditto

pkg/store/event/event.go
```diff
@@ -44,15 +44,114 @@ type eventStore struct {
...
+	podRecords := make([]*core.PodRecord, 0)
+
+	if err := e.db.Where(
+		&core.PodRecord{PodName: name, Namespace: namespace}).
+		Find(&podRecords).Error; err != nil && !gorm.IsRecordNotFoundError(err) {
+		return nil, err
+	}
+
+	et := new(core.Event)
+	eventList := make([]*core.Event, 0)
```
yeya24, line 118: A small nit.

`eventList := make([]*core.Event, 0, len(podRecords))` would be better

pkg/store/event/event.go
```diff
@@ -44,15 +44,114 @@ type eventStore struct {
...
+	}
+
+	for _, et := range resList {
+		pods, err:= e.FindPodsByEventID(context.Background(), et.ID)
+		if err != nil {
+			return nil, err
+		}
+		for _, pod := range(pods) {
+			et.Pods = append(et.Pods, pod)
+		}
```
yeya24, line 100: ditto

pkg/store/event/event.go
```diff
@@ -44,15 +44,114 @@ type eventStore struct {
...
+
+	return eventList, nil
 }
-func (e *eventStore) ListByPod(context.Context, string, string) ([]*core.Event, error) {
-	return nil, nil
+
+// ListByExperiment returns a event list by the name and namespace of the experiment.
+func (e *eventStore) ListByExperiment(_ context.Context, namespace string, experiment string) ([]*core.Event, error) {
+	var resList []core.Event
+	eventList := make([]*core.Event, 0)
```
yeya24, line 84: The length is known

pkg/store/event/event.go
```diff
@@ -44,15 +44,114 @@ type eventStore struct {
...
+	}
+
+	for _, et := range(resList) {
+		pods, err:= e.FindPodsByEventID(context.Background(), et.ID)
+		if err != nil {
+			return nil, err
+		}
+		for _, pod := range(pods) {
+			et.Pods = append(et.Pods, pod)
+		}
```
yeya24, line 74: Seems you can directly use et.Pods = pods

pkg/apiserver/event/event.go
```diff
@@ -47,11 +51,45 @@ func NewService(
...
+		eventList, err := s.event.List(context.Background())
+		if err != nil {
+			c.JSON(http.StatusOK, gin.H{
+				"status": statuscode.GetResourcesFromDBWrong,
+				"message": "get events wrong",
+				"data": eventList,
+			})
+			return
+		}
+	} else if (name != "" && namespace == "") || (name == "" && namespace != "") {
```
yeya24, line 75: So we must have both `name` and `namespace`? I think there is a case that I only want to see the events in one namespace. WDYT?

fewdan: Understand, I will modify it.

pkg/apiserver/event/event.go
fewdan: Why is it "l"?

yeya24: sorry a typo

pkg/store/event/event.go
```diff
@@ -44,15 +44,106 @@ type eventStore struct {
 	db *dbstore.DB
 }
 
-// TODO: implement core.EventStore interface
-func (e *eventStore) List(context.Context) ([]*core.Event, error) { return nil, nil }
-func (e *eventStore) ListByExperiment(context.Context, string, string) ([]*core.Event, error) {
-	return nil, nil
+// FindPodsByEventID return the list of PodRecords according to the eventID
+func (e *eventStore) FindPodsByEventID(_ context.Context, id uint) ([]*core.PodRecord, error) {
```
Yisaer, line 48: I think `FindPodsByEventID` is misleading. What about `FindPodRecordsByEventID` ?

pkg/apiserver/event/event.go
```diff
@@ -47,11 +52,52 @@ func NewService(
...
+// @Produce json
+// @Success 200 {array} string
+// @Router /api/events/all [get]
+// @Failure 500 {object} utils.APIError
+func (s *Service) listEvents(c *gin.Context) {
+	name := c.Query("name")
+	namespace := c.Query("namespace")
+	eventList := make([]*core.Event, 0)
+
+	if name == "" && namespace == "" {
```
zhouqiang-cl, line 73: Did we need to define a funtion like CheckNameAndNamespace(It can be done in other PR)

pkg/store/event/event.go
```diff
@@ -44,15 +44,135 @@ type eventStore struct {
...
+		eventList = append(eventList, &et)
+	}
+
+	return eventList, nil
 }
-func (e *eventStore) ListByPod(context.Context, string, string) ([]*core.Event, error) {
-	return nil, nil
+
+// ListByNamespace returns the list of events according to the namespace
+func (e *eventStore) ListByNamespace(_ context.Context, namespace string) ([]*core.Event, error) {
```
zhouqiang-cl, line 104: why we leave context _?

yeya24: This is a common pattern if this parameter is not used.

fewdan: Because everything needed is in the DB in the eventStore, the context is temporarily not used.  I don't know if the context will be used in subsequent changes.

pkg/apiserver/event/event.go
```diff
@@ -47,11 +52,52 @@ func NewService(
...
+	endpoint := r.Group("/events")
 
 	// TODO: add more api handlers
 	endpoint.GET("/all", s.listEvents)
 }
 
-// TODO: need to be implemented
-func (s *Service) listEvents(c *gin.Context) {}
+// @Summary Get all events from db.
+// @Description Get all chaos kinds from db.
```
cwen0, line 62: s/chaos kinds/chaos events/g

pkg/apiserver/event/event.go
```diff
@@ -47,11 +52,52 @@ func NewService(
...
+// @Tags events
+// @Produce json
+// @Success 200 {array} string
```
cwen0, line 65: s/{array} string/{array} core.Event/g

pkg/apiserver/event/event.go
```diff
@@ -47,11 +52,52 @@ func NewService(
...
+
+	if name == "" && namespace == "" {
+		resList, err := s.event.List(context.Background())
+		if err != nil {
+			c.Status(http.StatusInternalServerError)
+			_ = c.Error(utils.ErrInternalServer.WrapWithNoMessage(err))
+			return
+		}
+		eventList = resList
+	} else if (name == "" && namespace != "") {
```
cwen0, line 81: `()` is not necessary.

pkg/apiserver/event/event.go
```diff
@@ -47,11 +52,52 @@ func NewService(
...
+		resList, err := s.event.ListByNamespace(context.Background(), namespace)
+		if err != nil {
+			c.Status(http.StatusInternalServerError)
+			_ = c.Error(utils.ErrInternalServer.WrapWithNoMessage(err))
+			return
+		}
+		eventList = resList
+	}else if (name != "" && namespace == "") {
```
cwen0, line 89: we can check this at first.

pkg/store/event/event.go
```diff
@@ -44,15 +44,135 @@ type eventStore struct {
 	db *dbstore.DB
 }
 
-// TODO: implement core.EventStore interface
-func (e *eventStore) List(context.Context) ([]*core.Event, error) { return nil, nil }
-func (e *eventStore) ListByExperiment(context.Context, string, string) ([]*core.Event, error) {
-	return nil, nil
+// FindPodRecordsByEventID return the list of PodRecords according to the eventID
```
cwen0, line 47: s/return/returns/g

pkg/store/event/event.go
```diff
@@ -44,15 +44,135 @@ type eventStore struct {
...
+func (e *eventStore) FindPodRecordsByEventID(_ context.Context, id uint) ([]*core.PodRecord, error) {
```
cwen0, line 48: If this function won't be used by the other pkg, we can change it to `findPodRecordsByEventID`

pkg/apiserver/event/event.go
fewdan: There is no modification here, I don't quite understand what you mean.

pkg/apiserver/event/event.go
```diff
@@ -47,11 +52,52 @@ func NewService(
...
+// @Description Get all chaos events from db.
+// @Tags events
+// @Produce json
+// @Success 200 {array} core.Event
+// @Router /api/events/all [get]
+// @Failure 500 {object} utils.APIError
+func (s *Service) listEvents(c *gin.Context) {
+	name := c.Query("name")
+	namespace := c.Query("namespace")
+	eventList := make([]*core.Event, 0)
```
yeya24, line 71: I think the alert is for this variable. Could you please update it to `var eventList []*core.Event`? @fewdan

yeya24: image

yeya24: Or we don't need to use the variable `resList`, you can just use 

```
var eventList []*core.Event
var err error
eventList, err = s.event.List()
```

yeya24:
Umm, I don't know why this is still not merged even if the E2E is passed. Let me merge this manually.

PR #528, opened 2020-05-19
Add metrics to Chaos Mesh controller

### LINKED ISSUE #148: proposal: add metrics for chaos CRD
From @yeya24 's suggestion:
Chaos Mesh need some metrics to show how many CRDs it have for each kind of injection, we can use `gauge` metric.
If we have such metrics, it can display the history of CRDs and can be added to dashboard

yeya24:

Fixes #148

This PR is based on #523 so please merge that one first.

Verification

image


controllers/metrics/metrics.go
```diff
@@ -0,0 +1,134 @@
+package metrics
```
cwen0, line 1: Add LICENCE

controllers/metrics/metrics.go
```diff
@@ -0,0 +1,134 @@
...
+	c.TemplateNotExist.Collect(ch)
+	c.ConfigNameDuplicate.Collect(ch)
+	c.TemplateLoadError.Collect(ch)
+	c.InjectRequired.Collect(ch)
+	c.Injections.Collect(ch)
+	c.experimentStatus.Collect(ch)
+}
+
+func (c *ChaosCollector) collect() {
+	kinds := map[string]v1alpha1.ChaosList{
```
cwen0, line 101: we also can remove this.

helm/chaos-mesh/templates/prometheus-deployment.yaml
```diff
@@ -32,6 +32,13 @@ spec:
       {{- if .Values.prometheus.serviceAccount }}
       serviceAccount: {{ .Values.prometheus.serviceAccount }}
       {{- end }}
+      initContainers:
+        - name: data-permission-fix
+          image: busybox
+          command: ["/bin/chmod","-R","777", "/data"]
+          volumeMounts:
```
cwen0, line 39: What is this volume used for?

yeya24: The same as https://github.com/pingcap/tidb-operator/blob/62ad3e41f741b3cb0c84a05f8a47e351dd9382b4/pkg/monitor/monitor/util.go#L257:6. Fix the data permission problem

yeya24:
@cwen0 Updated

yeya24:
This PR uses too many `if metrics != nil` pattern. This is not good and redundant indeed. But currently I don't get a good way to add the metrics code into unit tests so that's a temporary solution. Let me refine this later if possible.

PR #545, opened 2020-05-23
Add document to the new IO chaos template mechanism

yeya24:

Document update for the feature #502


doc/io_chaos.md
```diff
@@ -56,11 +69,11 @@ ARGS="--pd=${CLUSTER_NAME}-pd:2379 \
 > * If you are testing a TiDB cluster, you need to modify it at [`_start_tikv.sh.tpl`](https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-cluster/templates/scripts/_start_tikv.sh.tpl).
 > * PD has the same issue with TiKV. You need to modify the data directory of PD at [`_start_pd.sh.tpl`](https://github.com/pingcap/tidb-operator/blob/master/charts/tidb-cluster/templates/scripts/_start_pd.sh.tpl).
 
-## ConfigMap configuration
+## Injection configuration
 
-Chaos Mesh uses sidecar container to inject IOChaos. To fulfill this chaos, you need to configure this sidecar container using a [ConfigMap](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/).
+Injection configuration is another [ConfigMap](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/). that is required to fulfill IO Chaos.
```
fewdan, line 74: s/that/That/g  ?

cwen0:
@dcalvin PTAL, Thanks!

yeya24:
I think we have to review this ASAP since this PR is urgent. We have the new IO chaos implementation for 3 weeks but the doc is still outdated. And this PR is not for fixing English grammars so we just need to review whether this doc is helpful for users to understand our new implementation.

@cwen0 @zhouqiang-cl Please review and see what should change. For the grammar and style problems, we can put in a later PR.

dcalvin:
Sorry for the late response. I was occupied with other priories. Just had a quick run, and I think there is some editing to be done wrt the structure and the language. But I agree we can merge it first if there are no technical issues.

docs/sidecar_configmap.md
```diff
@@ -8,11 +8,123 @@ Chaos Mesh runs a [fuse-daemon](https://www.kernel.org/doc/Documentation/filesys
...
+Currently, Chaos Mesh supports two types of ConfigMaps:
 
-The following content is the sidecar ConfigMap defined for tikv:
+1. Template config. The skeleton of each sidecar config is similar, in order to fulfill different requirements and make the configuration simplified,
+Chaos Mesh supports creating common templates to be used by different applications.
+
+2. Injection config. This configuration will be combined with template config and finally generate a config to inject to the selected pods. 
+Since most applications use different data directories, volume name or container name, you can define different parameters based on the common template created in the first step.
+
+## Template Configuration
```
cwen0, line 21: Can we split `Template Configuration` to a single file? Because most users don't need to care about the details about this config and just need to know which name they need to set. We want most users to ignore this file.

docs/sidecar_configmap.md
```diff
@@ -24,21 +136,46 @@ metadata:
...
     app.kubernetes.io/component: webhook
 data:
-  # the real content of config
-  chaosfs-tikv.yaml: |
+  chaosfs-tikv: |
     name: chaosfs-tikv
     selector:
       labelSelectors:
         "app.kubernetes.io/component": "tikv"
+    template: sidecar-template
```
cwen0, line 144: Can use rename `sidecar-template`? Such as `fuse-sidecar` or `chaosfs-sidecar`?

docs/sidecar_configmap.md
yeya24: What do you mean by `split to a single file`? It is a single file now. Would you mind elaborating your idea?

docs/sidecar_configmap.md
yeya24: Totally okay. I like `chaosfs-sidecar`

docs/sidecar_configmap.md
cwen0: We can introduce the `Template Configuration`  in a single file and we just introduce the `Injection configuration` at this file, because we most users just need to care about `injection configuration`.

yeya24: Sorry I still cannot understand your idea. Do you want to put `Template config` and `Injection config` into one config file? If so, then this is out of scope of this PR because this is another feature request which we don't support yet. 

If you want that feature, you can open another issue to track that implementation.

cwen0: Just the document.

docs/sidecar_configmap.md
yeya24: Done.

docs/sidecar_configmap.md
yeya24: Done. PTAL

PR #546, opened 2020-05-23
[API]listEventsByExperiment

fewdan:
chaos-server listEventsByExperiment

chaos-server listEventsByExperiment

### Check List 

Tests 

 - No code

Code changes

 - Has Go code change

Side effects

 - NO

Related changes

 - NO

```release-note
NONE
```


pkg/apiserver/event/event.go
```diff
@@ -100,3 +103,47 @@ func (s *Service) listEvents(c *gin.Context) {
...
+	} else if startTimeStr == "" {
+		eventList, err = s.event.ListByExperiment(context.Background(), namespace, name)
+		if err != nil {
+			c.Status(http.StatusInternalServerError)
+			_ = c.Error(utils.ErrInternalServer.WrapWithNoMessage(err))
+			return
+		}
+	} else {
+		//startTime, err := time.ParseInLocation("2006-01-02 15:04:05", startTimeStr, time.Local)
+		startTime, err := time.Parse(time.RFC3339, strings.Replace(startTimeStr, " ", "+", -1))
```
fewdan, line 133: Here I have a question about how to pass time data. I didn't find a more appropriate method, do you have any opinions?

fewdan: I tried to change several time formats, but it always "not found" when searching in db.

yeya24:
What about making the event interface looks like https://github.com/pingcap/chaos-mesh/pull/538/files#diff-8df376d7f7bbb5300e6558f0b3439aefR28? I prefer making the list API in one method.

pkg/apiserver/event/event.go
```diff
@@ -56,18 +58,19 @@ func Register(r *gin.RouterGroup, s *Service) {
 
 	// TODO: add more api handlers
 	endpoint.GET("/", s.listEvents)
```
yeya24, line 60: This API should be `endpoint.GET("", s.listEvents)`. Would you mind updating this in this PR?

```
curl localhost:2333/api/events
```

fewdan:
@yeya24 I have edited it. PTAL

PR #553, opened 2020-05-25
Fix CI

Gallardot:

#551 , Force the developer to `make check` before create a PR

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


Makefile
```diff
@@ -164,7 +164,7 @@ vet:
 tidy:
 	@echo "go mod tidy"
 	GO111MODULE=on go mod tidy
-	git diff --quiet go.mod go.sum
+	git diff --quiet go.mod go.sum api config controllers manifests pkg test
```
yeya24, line 167: I don't think this is the change I want to see. `tidy`should only check the go module dependency.

Something like https://github.com/pingcap/tidb-operator/blob/master/hack/verify-all.sh is better

Gallardot: Done.  PTAL

Issue #414, opened 2020-04-11
Protect some namespaces


Gallardot:
## Feature Request

**Is your feature request related to a problem? Please describe:**

**Describe the feature you'd like:**

Currently we can inject chaos into any k8s namespace. But sometimes we might only allow chaos to be injected into a specific scope of namespace.For example,we are not allowed to inject chaos into kube-system namespace . Maybe we need an allow list or a disable list.

https://github.com/pingcap/chaos-mesh/blob/46a8a5c5b2d9358b67cf687585239288c70c4390/pkg/webhook/inject/inject.go#L36-L39

I noticed that there was a ignore list when sidecar was injected. But that's not enough

**Describe alternatives you've considered:**

I don't think this is a requirement that changes frequently, so we can now configure it in an environment variable.

And then change the logic of `selectPods`
https://github.com/pingcap/chaos-mesh/blob/46a8a5c5b2d9358b67cf687585239288c70c4390/pkg/utils/selector.go#L78
```

```

**Teachability, Documentation, Adoption, Migration Strategy:**

Gallardot:
merge on #661 , close this

PR #563, opened 2020-05-27
Protected namespaces

Gallardot:
#414 
### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


pkg/utils/consts.go
```diff
@@ -14,26 +14,31 @@
...
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
 var (
-	InitLog            = ctrl.Log.WithName("setup")
-	DashboardNamespace string
-	DataSource         string
+	InitLog           = ctrl.Log.WithName("setup")
+	AllowedNamespaces []string
+	IgnoredNamespaces []string
```
yeya24, line 26: I think a blacklist is okay to protect namespaces so `IgnoredNamespaces` is enough. What do you think?

Gallardot: > I think a blacklist is okay to protect namespaces so `IgnoredNamespaces` is enough. What do you think?

@yeya24 

If only a few namespaces were allowed to execute chaos, then setting AllowedNamespaces would be a good idea.
If only a few namespaces were not allowed to execute chaos, then setting IgnoredNamespaces would be a good idea.

If both AllowedNamespaces and IgnoredNamespaces are set, only AllowedNamespaces will take effect.

pkg/utils/selector.go
```diff
@@ -90,8 +90,12 @@ func selectPods(ctx context.Context, c client.Client, selector v1alpha1.Selector
...
 					return nil, err
 				}
-
-				pods = append(pods, pod)
+				if IsAllowedNamespaces(pod.Namespace) {
+					pods = append(pods, pod)
+				} else {
+					log.Info("filter pod by namespaces",
+						"pod", pod.Name, "namespace", pod.Namespace)
+				}
```
yeya24, line 98: I prefer we don't check namespace here since there is also a check at line 122.

Or maybe you can keep this and use pods = append(pods, filterByNamespaces(podList.Items)...)

pkg/utils/selector.go
```diff
@@ -360,8 +366,44 @@ func filterByPhaseSelector(pods []v1.Pod, phases labels.Selector) ([]v1.Pod, err
...
+		}
+	}
+	return filteredList
+}
+
+// IsAllowedNamespaces return whether namespace allows the execution of a chaos task
+func IsAllowedNamespaces(namespace string) bool {
+	if len(AllowedNamespaces) > 0 {
+		for _, ns := range AllowedNamespaces {
+			if strings.HasPrefix(namespace, ns) || strings.HasSuffix(namespace, ns) {
```
yeya24, line 387: Sorry, but I don't really understand why do we need `HasPrefix` and `HasSuffix` here? Is `namespace==ns` not enough? 

I think we only want to support equal match.

helm/chaos-mesh/README.md
```diff
@@ -31,6 +31,8 @@ The following tables list the configurable parameters of the Chaos Mesh chart an
 | `controllerManager.tolerations` |  Toleration labels for chaos-controller-manager pod assignment | `[]` |
 | `controllerManager.affinity` |  Map of chaos-controller-manager node/pod affinities | `{}` |
 | `controllerManager.podAnnotations` |  Pod annotations of chaos-controller-manager | `{}`|
+| `controllerManager.allowedNamespaces` |  The namespace list allow the execution of a chaos task. Commas separate multiple namespaces. | ``|
```
yeya24, line 34: allows

helm/chaos-mesh/README.md
```diff
@@ -31,6 +31,8 @@ The following tables list the configurable parameters of the Chaos Mesh chart an
...
+| `controllerManager.ignoredNamespaces` |  The namespace list ignore the chaos task. Commas separate multiple namespaces. Configuring `allowedNamespaces` at the same time will ignore this configuration. | ``|
```
yeya24, line 35: ignores

helm/chaos-mesh/templates/controller-manager-deployment.yaml
```diff
@@ -54,6 +54,14 @@ spec:
...
           - name: BPFKI_PORT
             value: !!str {{ .Values.bpfki.grpcPort }}
+          {{- if .Values.controllerManager.allowedNamespaces }}
+          - name: ALLOWED_NAMESPACES
+            value: {{ .Values.controllerManager.allowedNamespaces }}
+          {{- end }}
+          {{- if .Values.controllerManager.ignoredNamespaces }}
+          - name: IGNORED_NAMESPACES
+            value: {{ .Values.controllerManager.ignoredNamespaces }}
+          {{- end }}
```
yeya24, line 64: I don' t like using env vars to pass config and I prefer using two flags.
Using env var lacks sth because you cannot find this config when you use `./controller-manager -h`.

Gallardot: > I don' t like using env vars to pass config and I prefer using two flags.
> Using env var lacks sth because you cannot find this config when you use `./controller-manager -h`.

I think this is just a configuration style difference. If we need a consistent configuration style, we can reduce the use of environment variables as configurations

pkg/utils/selector.go
Gallardot: the logic for checking namespace here is different from that in line 122.

pkg/utils/selector.go
Gallardot: > Sorry, but I don't really understand why do we need `HasPrefix` and `HasSuffix` here? Is `namespace==ns` not enough?
> 
> I think we only want to support equal match.

Sometimes ,We may deploy three environments for Development / Integration / Testing in the same cluster.
We may use label or prefix or suffix to distinguish between different environments.
So I think support equal match is not enough. @yeya24

yeya24: Sorry but I disagree with this. The __fix matching is dangerous and this shouldn't be a default behavior.

pkg/utils/selector.go
yeya24: What is the difference? I can only see that the namespace of pods is checked twice

pkg/utils/selector.go
Gallardot: Perhaps we can provide a configuration item to select an  equal match or a __fix match? @yeya24

pkg/utils/selector.go
Gallardot: https://github.com/pingcap/chaos-mesh/blob/f22ba60db393d0643597ad6f131b068fa43a2152/pkg/utils/selector.go#L82-L106

if `len(selector.Pods) > 0` , We will ignore the other options. So I think we should do a check here as well.

pkg/utils/selector.go
yeya24: If you want to support __fix, I would say it is better to support regex directly. __fix is not clear enough.
IMO equal matches can already cover most of the cases. Let's get @cwen0 's opinion.

pkg/utils/selector.go
yeya24: Hmm, that's true. But seems the default behavior is a little weird. If we set the podSelector, then we don't do other filterings? @cwen0

pkg/utils/selector.go
Gallardot: We have about 200 namespaces in a k8s cluster, and they are the Development/Integration/Testing environment for business. If only support equal matches，We need to set a lot of namespace in `AllowedNamespaces`. If we create a new namespace, we also need to update this.

helm/chaos-mesh/values.yaml
```diff
@@ -26,6 +26,9 @@ controllerManager:
   nameOverride: ""
   fullnameOverride: ""
 
+  allowedNamespaces: ""
```
cwen0, line 29: How about using `controllerManager.env` to manage all environment variables like chaos-dashboard and define a config struct to unmarshal them?

pkg/utils/selector.go
cwen0: How about using simple regular expressions?

helm/chaos-mesh/values.yaml
Gallardot: Good idea. I'll take a look

pkg/utils/selector.go
Gallardot: @cwen0 @yeya24 

OK,I'll re-implement it using regular expressions.

Gallardot:
@cwen0 @yeya24 
All done! PTAL. Thank you.

Issue #520, opened 2020-05-18
[API] Update chaos experiment by API


cwen0:
## Feature Request

**Is your feature request related to a problem? Please describe:**

1. /experiments/edit: Edit the experiment 

We need to create chaos experiment by API, [API document](https://app.swaggerhub.com/apis/ChaosMesh/chaos-server/1.0.0#/experiment/patch_experiments_edit)

PR #570, opened 2020-05-28
Support updating chaos experiments by API

cwen0:

#520 

Support updating chaos experiments by API

### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


pkg/apiserver/experiment/experiment.go
```diff
@@ -717,3 +699,244 @@ func (s *Service) patchExperiment(exp *ExperimentBase, annotations map[string]st
...
+		v1alpha1.KindIOChaos:      s.updateIOChaos,
+		v1alpha1.KindStressChaos:  s.updateStressChaos,
+		v1alpha1.KindTimeChaos:    s.updateTimeChaos,
+		v1alpha1.KindKernelChaos:  s.updateKernelChaos,
+	}
+
+	f, ok := updateFuncs[exp.Target.Kind]
+	if !ok {
+		c.Status(http.StatusBadRequest)
+		_ = c.Error(utils.ErrInvalidRequest.New("Target kind is not available"))
```
fewdan, line 732: How about changing to "_ = c.Error(utils.ErrInvalidRequest.New(exp.Target.Kind + " is not supported")) "?

pkg/apiserver/experiment/experiment.go
```diff
@@ -224,49 +227,28 @@ func (s *Service) createExperiment(c *gin.Context) {
...
+		v1alpha1.KindIOChaos:      s.createIOChaos,
+		v1alpha1.KindStressChaos:  s.createStressChaos,
+		v1alpha1.KindTimeChaos:    s.createTimeChaos,
+		v1alpha1.KindKernelChaos:  s.createKernelChaos,
+	}
+
+	f, ok := createFuncs[exp.Target.Kind]
+	if !ok {
 		c.Status(http.StatusBadRequest)
 		_ = c.Error(utils.ErrInvalidRequest.New("Target kind is not available"))
```
fewdan, line 242: How about changing to "_ = c.Error(utils.ErrInvalidRequest.New(exp.Target.Kind + " is not supported")) "?

pkg/apiserver/experiment/experiment.go
```diff
@@ -86,6 +86,7 @@ func Register(r *gin.RouterGroup, s *Service) {
 	endpoint.POST("/new", s.createExperiment)
 	endpoint.GET("/detail/:kind/:namespace/:name", s.getExperimentDetail)
 	endpoint.DELETE("/:kind/:namespace/:name", s.deleteExperiment)
+	endpoint.PUT("/update", s.updateExperiment)
```
yeya24, line 89: I am just wondering do we need to make `kind`, `namespace` and `name` in the API endpoint? We don't do this for `/new` as well. What do you think? @cwen0

cwen0: These messages are included in the request body, I think it is unnecessary to add them in endpoint and they are repeated messages.

yeya24: I mean remove name and namespace from body and add them to URL. Do you think this makes sense?

cwen0: I think it is ok at present, if we need this way, we can update this in the future.

yeya24: Actually I am not against the current way, but let's choose one solution and don't change it in the future since it is necessary to keep compatibility.

It is okay to use the current way.

tools.json
```diff
@@ -23,10 +23,6 @@
     {
       "Repository": "github.com/mgechev/revive",
       "Commit": "7773f47324c2bf1c8f7a5500aff2b6c01d3ed73b"
-    },
-    {
-      "Repository": "github.com/swaggo/swag",
-      "Commit": "eebe67fc31c39a02a763bcaa815baab805112cf3"
```
yeya24, line 29: Is it intended?

PR #585, opened 2020-06-02
Fix missing Error check in e2e test

Yisaer:
Fix missing Error check in e2e test

```release-note
NONE
```


yeya24:
@Yisaer Does this fix the flaky? Seems this is another thing.

Yisaer:
> @Yisaer Does this fix the flaky? Seems this is another thing.

I'm not sure. In the existed log the error happened in verify recovering the chaos, but it may still be caused by the applying the chaos that we forgot to verify applying chaos error here. I think this request would expose the root error about the flaky e2e test.

To solve the flaky test, we need to get the logs of chaos controller which I am working on it.

yeya24:
Please fix CI and then we can merge. Anyway this change LGTM

PR #592, opened 2020-06-04
[API] store duration field to event table

cwen0:

* Add duration field 

example: "300ms", "-1.5h" or "2h45m",  Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".


controllers/twophase/types.go
```diff
@@ -150,7 +151,7 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 			return ctrl.Result{}, err
 		}
 
-		if err := applyAction(ctx, r, req, chaos); err != nil {
+		if err := applyAction(ctx, r, req, *duration, chaos); err != nil {
```
yeya24, line 154: We cannot set the duration when the chaos starts since the duration is not determined. What if it is paused and the chaos is not resumed? Like a chaos has duration 5m, but it only runs 1m and pauses. After 5m, the pause resumes and starts the next round. In this case, the actual duration is only 1m.

cwen0: We can reset it when we pause the chaos. This field is required by the front-end.

yeya24: Yes. This is a good idea. I agree we should change update this when we pause

cwen0:
@yeya24 PTAL again, Thanks!

Issue #596, opened 2020-06-05
PodKill and ContainerKill should only support schedule mode in frontend


yeya24:
## Bug Report

1. Choose podkill or container kill in the `Target` page, then click `Next`
![image](https://user-images.githubusercontent.com/25150124/83825081-445dc780-a6a6-11ea-9a8f-341e6f9e78b1.png)

2. It shows `The action you chose must be immediate`. However podkill and container kill only works at schedule mode so https://github.com/pingcap/chaos-mesh/blob/14493ac9df745ba3b223c541c98e52c76e493879/ui/src/lib/formikhelpers.ts#L76 function just does the opposite thing.
![image](https://user-images.githubusercontent.com/25150124/83825085-4aec3f00-a6a6-11ea-825f-b23bdbd68f8b.png)

g1eny0ung:
Solved in #597.

PR #597, opened 2020-06-05
fix: only schedule mode in pod(container)-kill

g1eny0ung:

This PR solves #596.


ui/src/lib/formikhelpers.ts
```diff
@@ -73,7 +73,7 @@ export function parseSubmitValues(values: Experiment) {
   return values
 }
 
-export function mustImmediate(formikValues: Experiment) {
+export function mustSchedule(formikValues: Experiment) {
   if (
     formikValues.target.pod_chaos.action === 'pod-kill' ||
     formikValues.target.pod_chaos.action === 'container-kill'
```
yeya24, line 79: There is another problem here. Since we only clean up the formikValues when we submit the experiment, so if we first click `Pod Chaos`, and then click `Network Chaos` and submit, then this condition `formikValues.target.pod_chaos.action === 'pod-kill'` is still satisfied, however the actual kind we are using is `NetworkChaos`. I reproduced this several times.

A possible way I can think of now is to add a check `formikValues.target.kind === 'PodChaos'`, WDYT?

g1eny0ung: There has a solution for this is clear other `Chaos` when you select a new `Chaos`. For example, when you click the `Network Chaos` tab, the others will be reset to the default values.

What about this way?

g1eny0ung: > A possible way I can think of now is to add a check formikValues.target.kind === 'PodChaos', WDYT? I am not sure if there is a better solution. Can we reset the experiment value each time when we move the stepper?

This way can also solve this problem. But I think if we don't clean the **old chaos** which users selected, the user will be confused when returning to the target step. Which one did he choose? There will be many chaos actions visually.

yeya24: Yes, I agree we should clear the old status. But I think it is not easy. :) Anyway please fix this using the way you like, I am new to frontend so please ignore my idea if it is wrong

g1eny0ung: Ok, I will fix this soon.

g1eny0ung: @yeya24 fix in 548f153 and thx for mentioned this problem.~

yeya24:
Hey, @g1eny0ung  I test this feature locally. But seems there is another problem.

After I click NetworkChaos and enter something, then I switch back to PodChaos. Then I cannot select any actions in the `SelectField`. Is it related to your change?

Screenshot from 2020-06-05 07-56-15

g1eny0ung:
@yeya24 I follow your description and test several times. It's ok for me to begin switching to `NetworkChaos`, select an action and type something and then switch to `PodChaos` and select an action.

yeya24:
> @yeya24 I follow your description and test several times. It's ok for me to begin switching to `NetworkChaos`, select an action and type something and then switch to `PodChaos` and select an action.

Would you minding holding this PR today? I will try to provide detailed reproduce procedures.

g1eny0ung:
> > @yeya24 I follow your description and test several times. It's ok for me to begin switching to `NetworkChaos`, select an action and type something and then switch to `PodChaos` and select an action.
> 
> Would you minding holding this PR today? I will try to provide detailed reproduce procedures.

Yep. It's ok for me.

yeya24:
> > > @yeya24 I follow your description and test several times. It's ok for me to begin switching to `NetworkChaos`, select an action and type something and then switch to `PodChaos` and select an action.
> > 
> > 
> > Would you minding holding this PR today? I will try to provide detailed reproduce procedures.
> 
> Yep. It's ok for me.

The steps are:
1. Select `NetworkChaos` and `Delay`, enter 1s and click `Next`
2. Go back to the `Target` page and try to select actions. 
3. You will find you can select actions but you cannot see anything in the SelectField.

But it is weird that when I go back to the `Target` page, and click `Pod Lifecycle`, then I can choose actions. If not click, then it will reproduce this problem. Is this expected?

g1eny0ung:
> > > > @yeya24 I follow your description and test several times. It's ok for me to begin switching to `NetworkChaos`, select an action and type something and then switch to `PodChaos` and select an action.
> > > 
> > > 
> > > Would you minding holding this PR today? I will try to provide detailed reproduce procedures.
> > 
> > 
> > Yep. It's ok for me.
> 
> The steps are:
> 
> 1. Select `NetworkChaos` and `Delay`, enter 1s and click `Next`
> 2. Go back to the `Target` page and try to select actions.
> 3. You will find you can select actions but you cannot see anything in the SelectField.
> 
> But it is weird that when I go back to the `Target` page, and click `Pod Lifecycle`, then I can choose actions. If not click, then it will reproduce this problem. Is this expected?

Know about this problem. The reason is I will set the `Chaos kind` when clicking the left vertical tabs. If click next in target form and then back, the `Chaos kind` will not be updated except you click the tabs again. I will fix it soon.

g1eny0ung:
@yeya24 Now the problem of selection should be solved.

And also, the target form will remember the last changes when you switch back to it.

PR #598, opened 2020-06-05
Define a environment variables config struct for chaos controller

Gallardot:
### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


pkg/config/config.go
```diff
@@ -14,9 +14,26 @@
...
+var ControllerCfg *ChaosControllerConfig
+
+func init() {
+	conf, err := EnvironChaosController()
+	if err != nil {
+		log.Error(err, "Chaos Controller: invalid environment configuration")
+		os.Exit(1)
+	}
+	ControllerCfg = &conf
+}
```
yeya24, line 35: I don't think this is correct. If you want to use `init()` here, please separate the config package for chaos-dashboard and chaos-controller. Otherwise in the chaos-dashboard component it will execute this function, too

Gallardot: Thank you. I have fixed it. PTAL

yeya24: The `init` function is package level. So even you separate the go file for dashboard and controller, the init function will still be executed. https://www.digitalocean.com/community/tutorials/understanding-init-in-go

I recommend we don't use init function in the config package. What about using a global var to keep `ControllerConfig` in `controllers/common/types.go`?

Gallardot: @yeya24 
My mistake, I made a mistake in my test environment, I used an earlier version of Dashboard.
I have fixed it. PTAL. Thank you.

controllers/common/types.go
```diff
@@ -15,8 +15,11 @@ package common
 
 import (
 	"context"
+	"os"
 	"time"
 
+	"github.com/pingcap/chaos-mesh/pkg/config"
```
yeya24, line 21: Could you please group the import?

controllers/podchaos/containerkill/types.go
```diff
@@ -18,6 +18,8 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/pingcap/chaos-mesh/controllers/common"
```
yeya24, line 21: Please group import

Gallardot:
> I think it is good now, only a small nit. Let's merge this after the E2E passes. Thanks for the work @Gallardot

@yeya24  All done. PTAL. Thank you.

controllers/common/types.go
```diff
@@ -36,6 +38,20 @@ const (
 	AnnotationCleanFinalizerForced = `forced`
 )
 
+var log = ctrl.Log.WithName("controller")
+
+//ControllerCfg is a global variable to keep the configuration for Chaos Controller
+var ControllerCfg *config.ChaosControllerConfig
+
+func init() {
+	conf, err := config.EnvironChaosController()
```
cwen0, line 47: At present, we also have some configuration defined by the flag. Maybe we need to manage all configuration in this way and we also can unify this in a single pr.

Gallardot: > At present, we also have some configuration defined by the flag. Maybe we need to manage all configuration in this way and we also can unify this in a single pr.

Got it. I will do it in a single PR.

Issue #599, opened 2020-06-05
Failed to create containerKill chaos using chaos dashboard


yeya24:
## Bug Report

**What did you do?**

Try to create a container kill chaos using chaos dashboard.

But I see this action fails because the container_name not set.

![Screenshot from 2020-06-05 11-58-06](https://user-images.githubusercontent.com/25150124/83897712-ecfe3c80-a723-11ea-8f28-081cc133429a.png)

![Screenshot from 2020-06-05 11-58-19](https://user-images.githubusercontent.com/25150124/83897722-eff92d00-a723-11ea-98e7-596a8208599e.png)

The problem is that we are deleting fields when submitting https://github.com/pingcap/chaos-mesh/blob/14493ac9df745ba3b223c541c98e52c76e493879/ui/src/lib/formikhelpers.ts#L67.

This works well for network chaos but it will delete the `container_name` field for container kill chaos.

yeya24:
This is fixed by #603. Close this now.

PR #603, opened 2020-06-08
fix: save container_name

g1eny0ung:

This PR solves #599.

### DIFF
--- a/ui/src/lib/formikhelpers.ts
+++ b/ui/src/lib/formikhelpers.ts
@@ -64,6 +64,11 @@ export function parseSubmitValues(values: Experiment) {
           continue
         }
 
+        // Handle container-kill action
+        if (chaos.action === 'container-kill' && action === 'container_name') {
+          continue
+        }
+
         if (action !== chaos.action) {
           delete chaos[action]
         }


ui/src/lib/formikhelpers.ts
```diff
@@ -64,6 +64,11 @@ export function parseSubmitValues(values: Experiment) {
           continue
         }
 
+        // Handle container-kill action
+        if (chaos.action === 'container-kill' && action === 'container_name') {
+          continue
+        }
```
yeya24, line 70: Though this can fix the problem, I am not sure when we add more chaos types, do we also have to do something similar to this?

g1eny0ung: I have also considered this problem. But since there have some minor questions in API `/experiment/new`, I have to remove other empty actions to prevent them from interfering with the selected.

So, it's easier to remove other actions with a `for each loop`. But to be honest, the configuration file of the chaos experiment is relatively complicated. For now, I think we can allow some **patch** operations, like this PR change.

Admittedly, this approach is not very elegant. After all APIs are stabilized, I think this part of the code can be considered for refactoring.

yeya24: Great. Then let's merge this now.

PR #608, opened 2020-06-09
[add] get detail of archive

fewdan:
get detail of archive

get detail of archive

### Check List 

Tests 

 - No code

Code changes

 - Has Go code change

Side effects

 - NO

Related changes

 - NO

```release-note
NONE
```


cwen0:
@fewdan In my view, the report of the experiment should be general information of the experiment, such as the running time, all injected pods etc.

fewdan:
> @fewdan In my view, the report of the experiment should be general information of the experiment, such as the running time, all injected pods etc.

This interface has been changed to "detail". It can be distinguished from "report". And this interface can be used to obtain the yaml information for creating chaos.

pkg/apiserver/archive/experiment.go
```diff
@@ -81,3 +83,39 @@ func (s *Service) listExperiments(c *gin.Context) {
...
+
+// @Summary Get the details of chaos experiment.
+// @Description Get the details of chaos experiment.
+// @Tags archives
+// @Produce json
+// @Param namespace query string false "namespace"
+// @Param name query string false "name"
+// @Param kind query string false "kind" Enums(PodChaos, IoChaos, NetworkChaos, TimeChaos, KernelChaos, StressChaos)
+// @Success 200 {array} core.ArchiveExperiment
+// @Router /api/archives/report [get]
```
yeya24, line 95: This should be updated.

pkg/apiserver/archive/experiment.go
```diff
@@ -81,3 +83,39 @@ func (s *Service) listExperiments(c *gin.Context) {
...
+func (s *Service) experimentDetail(c *gin.Context) {
+	kind := c.Query("kind")
+	name := c.Query("name")
+	ns := c.Query("namespace")
+
+	if kind == "" || name == "" || ns == "" {
+		c.Status(http.StatusBadRequest)
+		_ = c.Error(utils.ErrInvalidRequest.New("kind, name or namespace is empty"))
+		return
+	}
```
yeya24, line 106: That's the problem we have talked about at the start of design. Since archives are collections of experiments in the database, we can only use id to uniquely identify one experiment. Using `kind`, `name` and `namespace` is not sufficient.

TBH I don't know how can we get the ID.

fewdan: Later we plan to add suffix to name when creating chaos to ensure that name is unique.

yeya24:
Please rebase master, thanks.

pkg/store/experiment/experiment.go
```diff
@@ -72,6 +73,51 @@ func (e *experimentStore) ListMeta(_ context.Context, kind, ns, name string) ([]
...
+	if uid == "" {
+		var err error
+		uid, err = e.getUID(context.TODO(), kind, namespace, name)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if err := e.db.Where(
+		"namespace = ? and name = ? and kind = ? and uid = ?", namespace, name, kind, uid).
```
yeya24, line 113: A small question, since we are using uid here, is it still necessary to use `namespace` `name` and `kind` in the SQL?

pkg/store/experiment/experiment.go
```diff
@@ -72,6 +73,51 @@ func (e *experimentStore) ListMeta(_ context.Context, kind, ns, name string) ([]
...
+	UID := archives[0].UID
+	st := archives[0].StartTime
+
+	for _, archive := range archives {
+		if st.Before(archive.StartTime) {
+			st = archive.StartTime
+			UID = archive.UID
+		}
+	}
+	return UID, nil
```
yeya24, line 98: Not really understand this. So if there are multiple archives, we only want to get the earliest UID?

pkg/store/experiment/experiment.go
fewdan: I am not sure whether the UID is globally unique. If the UID is not unique, adding ns and name is just to reduce the probability of UID duplication.

pkg/store/experiment/experiment.go
fewdan: If there are multiple archives with the same name, we need to get the one with the latest creation time. 
Because the detail interface obtains a specific archive, you can select the archive you want through the UID field. But if no UID is passed, the latest archive is selected by default.

pkg/apiserver/archive/experiment.go
```diff
@@ -81,3 +84,74 @@ func (s *Service) listExperiments(c *gin.Context) {
...
+// @Tags archives
+// @Produce json
+// @Param namespace query string false "namespace"
+// @Param name query string false "name"
+// @Param kind query string false "kind" Enums(PodChaos, IoChaos, NetworkChaos, TimeChaos, KernelChaos, StressChaos)
+// @Param uid query string false "uid"
+// @Success 200 {array} core.ArchiveExperiment
+// @Router /api/archives/detailSearch [get]
+// @Failure 500 {object} utils.APIError
+func (s *Service) experimentDetailSearch(c *gin.Context) {
```
cwen0, line 99: If we define this API as a search API, I think `kind`, `name`, and `ns` may be empty and we should return a list of the experiment instead of the single object.

PR #623, opened 2020-06-12
[add] add uid in events and archive_experiments

fewdan:
add uid in events and archive_experiments

add uid in events and archive_experiments

### Check List

Tests

- No code

Code changes

- Has Go code change

Side effects

- NO

Related changes

- NO

```release-note
NONE
```


yeya24:
Can I ask what this UID for? Do you want to connect the chaos object in Kubernetes with the event in DB? Could you please introduce some use cases?

fewdan:
> Can I ask what this UID for? Do you want to connect the chaos object in Kubernetes with the event in DB? Could you please introduce some use cases?

@yeya24 
There is a problem that if chaos with the same name is deleted and then created, it cannot be distinguished in db.
So if uid is introduced, you can distinguish chaos with the same name (because their uid is different). In this case, only need to modify the API, without modifying other parts of chaos-mesh, can meet the requirements.

yeya24:
> > Can I ask what this UID for? Do you want to connect the chaos object in Kubernetes with the event in DB? Could you please introduce some use cases?
> 
> @yeya24
> There is a problem that if chaos with the same name is deleted and then created, it cannot be distinguished in db.
> So if uid is introduced, you can distinguish chaos with the same name (because their uid is different). In this case, only need to modify the API, without modifying other parts of chaos-mesh, can meet the requirements.

Got it. I can understand that this field is used to differentiate different chaos but with the same name.

But from https://github.com/pingcap/chaos-mesh/pull/608#discussion_r437809694, if we plan to add unique suffix to chaos name, then is it still necessary to have the UID?

fewdan:
@yeya24 
We have discussed, and the conclusion is that there are some shortcomings in adding a suffix, so we decided to use UID+name+NS+kind to distinguish different chaos.
The disadvantages of adding a suffix are:
1. The name is changed, it is inconvenient for the user to edit or delete chaos.
2. The user cannot operate by "kubectl delete -f xxx.yaml".
If you use UID, you only need to modify the chaos-dashboard part to meet the user's requirements, and it is more reasonable.
So after this PR merge, I will modify other APIs.

pkg/core/event.go
```diff
@@ -66,6 +66,7 @@ type Event struct {
 	Namespace  string
 	Kind       string
 	Message    string
+	UID        string     `gorm:"index:uid"`
```
cwen0, line 69: s/UID/ExperimentID/g

cwen0, line 69: s/UID/ExperimentID/g

PR #628, opened 2020-06-15
feat: add events page

g1eny0ung:

As the title.


ui/src/components/EventsTable/index.tsx
```diff
@@ -0,0 +1,225 @@
...
+          {events &&
+            stableSort<SortedEvent>(events, getComparator(order, orderBy))
+              .slice(page * rowsPerPage, page * rowsPerPage + rowsPerPage)
+              .map((e) => (
+                <TableRow key={e.ID} hover>
+                  <TableCell>{e.Experiment}</TableCell>
+                  <TableCell>{e.Kind}</TableCell>
+                  <TableCell>{e.Namespace}</TableCell>
+                  <TableCell>{format(e.CreatedAt)}</TableCell>
+                  <TableCell>{format(e.UpdatedAt)}</TableCell>
```
yeya24, line 198: `Created` and `Updated` are not needed in the table. These are just metadata fields from gorm.Model. Shouldn't expose these to users.

ui/src/components/EventsTable/index.tsx
```diff
@@ -0,0 +1,225 @@
...
+                  <TableCell>{format(e.StartTime)}</TableCell>
+                  <TableCell>{format(e.FinishTime)}</TableCell>
```
yeya24, line 200: Screenshot from 2020-06-12 20-52-51
Getting invalid format here since the event is not finished yet. We should add a check in this case.

yeya24, line 200: Screenshot from 2020-06-15 11-39-27

Getting invalid date here since the event is not finished. We should add a check this case

yeya24:
I got a bug while I am using, but I am not sure what's the cause in the code.

Screenshot from 2020-06-15 11-40-57
The experiment aaa was deleted several minutes ago. But I am still seeing this in the chart.

g1eny0ung:
> I got a bug while I am using, but I am not sure what's the cause in the code.
> 
> Screenshot from 2020-06-15 11-40-57
> The experiment aaa was deleted several minutes ago. But I can still see this in the chart.

I also encountered this problem. Seems it's a bug of the `events` API.

ui/src/components/EventsTable/index.tsx
g1eny0ung: Already removed in 5647ea8.

ui/src/components/EventsTable/index.tsx
g1eny0ung: Resolved in 5647ea8.

yeya24:
> > I got a bug while I am using, but I am not sure what's the cause in the code.
> > Screenshot from 2020-06-15 11-40-57
> > The experiment aaa was deleted several minutes ago. But I can still see this in the chart.
> 
> I also encountered this problem. Seems it's a bug of the `events` API.

Will investigate this today. Good work!

g1eny0ung:
I append some patches to this PR, includes:

- Now can pause and restart an experiment
- Add detail link button at the end of each row of the events table
- Show pods info in the events table

yeya24:
@g1eny0ung Would you mind taking a look at the lgtm bot? Can you solve this?

g1eny0ung:
> @g1eny0ung Would you mind taking a look at the lgtm bot? Can you solve this?

I viewed its warning notification several hours ago. its actually "used", just the variable definition not used.

The reason I wanna keep it because it can help other people to understand what this code does if he wants to deep into it.

As u can see, I already put a comment `// eslint-disable-next-line` on the top of them. ESLint is a JS linter which is widely used in front-end, It has checked these codes (plays the same role as the bot).

I suggest banning the bot from checking TS codes because the code has already been linted during development.

yeya24:
I think maybe it is a little weird here because this is a table for events. Do you think it is better to use an `Event detail` here? Instead of a button for `Experiment detail`.

Screenshot from 2020-06-16 12-29-07

Another thing is that the events are stored in database, they are records for the past. But the `experiment detail` API will query from the current cluster, it represents the current status. It is possible that the experiment is not in the cluster anymore.

Screenshot from 2020-06-16 12-32-38

g1eny0ung:
@yeya24 I think you're right. There is a feature I haven't come true is when clicking the detail button, it will jump to the experiment detail page, I wanna replace the events table in the detail page to an event detail panel.

image

Replace this area to event detail which you selected.

What do you think this way?

yeya24:
> @yeya24 I think you're right. There is a feature I haven't come true is when clicking the detail button, it will jump to the experiment detail page, I wanna replace the events table in the detail page to an event detail panel.
> 
> image
> 
> Replace this area to event detail which you selected.
> 
> What do you think this way?

Wow, that's much better!

g1eny0ung:
@yeya24 The feature of `Event detail` already done.

And this thing:

> Another thing is that the events are stored in database, they are records for the past. But the experiment detail API will query from the current cluster, it represents the current status. It is possible that the experiment is not in the cluster anymore.

@fewdan will help us filter out the events of survival experiments.

yeya24:
TBH, I think the events table is a little narrow. 

Screenshot from 2020-06-17 12-22-30

What's your idea about this status bar, do we need to have this every page?  Seems not very important in events page. Especially the `New Experiment` button.

Screenshot from 2020-06-17 12-23-48

But these things are not very important in this PR. We can improve it later.

g1eny0ung:
@yeya24 For cases where the table will be narrow on a small screen. I add a patch commit bffa717.

it will make the left area (which table in) in the detail page from 2/3 to 3/4.

And also, now the sidebar will remember your last choice to initialize its initial width, mini or normal. You can toggle it to mini when your screen is small relatively.

zhouqiang-cl:
@cwen0 PTAL

PR #629, opened 2020-06-16
[update] get the list of events by uid

fewdan:
If expName and expNS are specified, but no UID is specified, the corresponding events will be returned based on the latest UID.

get the list of events by uid

### Check List

Tests

- No code

Code changes

- Has Go code change

Side effects

- no

Related changes

```release-note
NONE
```


pkg/store/event/event.go
```diff
@@ -237,12 +237,19 @@ func (e *eventStore) DeleteIncompleteEvents(_ context.Context) error {
...
 	var startTime time.Time
 
+	if uid == "" && experimentName != "" && experimentNamespace != "" {
+		uid, err = e.getUID(context.Background(), experimentNamespace, experimentName)
+		if err != nil {
+			return resList, err
+		}
+	}
+
 	if podName != "" {
```
cwen0, line 253: If the `UID` is not empty, we also can list events by `UID`.

PR #631, opened 2020-06-16
Add flag to install dependency only

cwen0:

* Add a new flag to install dependency only by `install.sh` 
* Update install.sh 
* Use `IfNotPresent` instead of `Always`

### DIFF
--- a/helm/chaos-mesh/values.yaml
+++ b/helm/chaos-mesh/values.yaml
@@ -21,7 +21,7 @@ controllerManager:
   replicaCount: 1
 
   image: pingcap/chaos-mesh:latest
-  imagePullPolicy: Always
+  imagePullPolicy: IfNotPresent
 
   nameOverride: ""
   fullnameOverride: ""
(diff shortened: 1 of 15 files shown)


helm/chaos-mesh/values.yaml
```diff
@@ -21,7 +21,7 @@ controllerManager:
   replicaCount: 1
 
   image: pingcap/chaos-mesh:latest
-  imagePullPolicy: Always
+  imagePullPolicy: IfNotPresent
```
Yisaer, line 24: Why change to `IfNotPresent`

cwen0: I think `always` is not necessary. Sometimes if the pod was restarted and we change some flags of it, which may cause the pod crashing.

Gallardot: I think we should keep it set to `always`. Because our image is `latest`.  Unless we specify a particular version.

cwen0: Got it

PR #636, opened 2020-06-17
docs: should use bash

g1eny0ung:

When I try to install chaos mesh with kind by https://chaos-mesh.org/docs/installation/get_started_on_kind#install-chaos-mesh, then I run into this error:

```
sh: 381: Syntax error: Bad for loop variable
```

After rechecking the `install.sh`, I think this command should use `bash`.

### DIFF
--- a/website/docs/installation/get_started_on_kind.md
+++ b/website/docs/installation/get_started_on_kind.md
@@ -18,7 +18,7 @@ curl -sSL https://raw.githubusercontent.com/pingcap/chaos-mesh/master/install.sh
 
 `install.sh` is an automation shell script that helps you install dependencies such as `kubectl`, `Helm`, `kind`, and `kubernetes`, and deploy Chaos Mesh itself.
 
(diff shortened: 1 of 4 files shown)


website/docs/installation/get_started_on_kind.md
```diff
@@ -13,7 +13,7 @@ Before deployment, make sure [Docker](https://docs.docker.com/install/) is insta
 ## Install Chaos Mesh
 
 ```bash
-curl -sSL https://raw.githubusercontent.com/pingcap/chaos-mesh/master/install.sh | sh -s -- --local kind
+curl -sSL https://raw.githubusercontent.com/pingcap/chaos-mesh/master/install.sh | bash -s -- --local kind
```
cwen0, line 16: we also need to update this in `pages/index.js`

g1eny0ung: Done.

PR #640, opened 2020-06-18
api-server: add the fallback page

cwen0:

* use `index.html` as the fallback page

### DIFF
--- a/tools/assets_generate/main.go
+++ b/tools/assets_generate/main.go
@@ -18,7 +18,7 @@ import (
 	"net/http"
 	"os"
 
-	"github.com/shurcooL/vfsgen"
+	"github.com/cwen0/vfsgen"
 )
 
 func main() {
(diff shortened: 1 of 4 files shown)


tools/assets_generate/main.go
```diff
@@ -18,7 +18,7 @@ import (
 	"net/http"
 	"os"
 
-	"github.com/shurcooL/vfsgen"
+	"github.com/cwen0/vfsgen"
```
yeya24, line 21: Is this change intended? You'd like to use own fork here?

cwen0: The origin repo doesn't support setting the fallback and the repo is rarely updated, but we need this feature. https://github.com/cwen0/vfsgen/commit/c35322782328a576b31b802094d540f346ac0b39

cwen0: I will try to push this pr to the origin repo.

PR #656, opened 2020-06-28
Add a new http fileserver support to support fallback

cwen0:

Refer https://github.com/shurcooL/vfsgen/pull/82#issuecomment-650583868 suggestion，we add a new HTTP fileserver to supports fallback.


pkg/httpfs/fs.go
```diff
@@ -0,0 +1,245 @@
...
+		if fi.IsDir() {
+			indexPath := pathpkg.Join(path, "index.html")
+			f0, err := fs.root.Open(indexPath)
+			if err == nil {
+				defer f0.Close()
+				fi0, err := f0.Stat()
+				if err == nil {
+					path = indexPath
+					f = f0
+					fi = fi0
```
fewdan, line 154: Do you need to change the name of the variable here? It looks a bit confusing.

pkg/httpfs/fs.go
```diff
@@ -0,0 +1,245 @@
...
+		}
+	}
+)
+
+type fileServer struct {
+	root http.FileSystem
+	opt  FileServerOptions
+}
+
+func (fs *fileServer) ServeHTTP(w http.ResponseWriter, req *http.Request) {
```
zhouqiang-cl, line 94: Add comment?

PR #659, opened 2020-06-29
Document: modify IOCChaos to IOChaos

tsthght:

In user_guides/sidecar_configmap.md, change `IOCChaos` to `IOChaos`.

In user_guides/sidecar_configmap.md, change `IOCChaos` to `IOChaos`.


tsthght:
PTAL @cwen0

PR #661, opened 2020-06-29
Protected namespaces

Gallardot:
#414 
### Check List 

Tests 

 - Unit test
 - E2E test
 - No code

Code changes

 - Has Go code change

Side effects

Related changes

```release-note
NONE
```


Gallardot:
> The code looks good. Just curious about the design, do we only need one regex instead of a list of regex?

@yeya24 I think one regular expression is enough.

For example, you want chaos tasks to be performed on a namespace with either the `DEV-` prefix or the `TEST-` suffix. The `allowedNamespaces ` can be set to `TEST-\w+|\w+-DEV`.

Issue #557, opened 2020-05-25
Archive deleted experiments multiple times


yeya24:
## Bug Report

```
select name,start_time,finish_time from archive_experiments where namespace='tipocket-cdc-kafka';
```

Seems the same experiment is stored in the DB several times.

![image](https://user-images.githubusercontent.com/25150124/82839967-1060f600-9e9f-11ea-94aa-ec7a3bd69f63.png)

The cause is that the cleanup of chaos is blocked by the finalized. But since the chaos daemon network problem, the recovery is always failing. But we stored the archive several times.

```
apiVersion: pingcap.com/v1alpha1
kind: TimeChaos
metadata:
  creationTimestamp: "2020-05-14T22:56:43Z"
  deletionGracePeriodSeconds: 0
  deletionTimestamp: "2020-05-14T22:58:57Z"
  finalizers:
  - tipocket-cdc-kafka/tipocket-cdc-kafka-upstream-tikv-1
  generation: 24
  name: tipocket-cdc-kafka-tipocket-cdc-kafka-upstream-tikv-1-time-chaos
  namespace: tipocket-cdc-kafka
  resourceVersion: "76364895"
  selfLink: /apis/pingcap.com/v1alpha1/namespaces/tipocket-cdc-kafka/timechaos/tipocket-cdc-kafka-tipocket-cdc-kafka-upstream-tikv-1-time-chaos
  uid: d3a19fc8-db45-4d43-b4a1-2c39ad826e68
spec:
  clockIds:
  - CLOCK_REALTIME
  mode: one
  paused: false
  selector:
    namespaces:
    - tipocket-cdc-kafka
    pods:
      tipocket-cdc-kafka:
      - tipocket-cdc-kafka-upstream-tikv-1
  timeOffset: -159ms
  value: ""
status:
  experiment:
    phase: Failed
  phase: ""
```

cwen0:
@yeya24 Can we determine whether the experiment is actually deleted?

yeya24:
> @yeya24 Can we determine whether the experiment is actually deleted?

I am not sure. Maybe @Yisaer knows better.

But IMO we should deal with the finalized problem indeed. If the injected pods are deleted, we should cleanup the chaos.

You can see there are no pods, but all the time chaos is left.

![image](https://user-images.githubusercontent.com/25150124/82855996-e5949300-9eda-11ea-8aff-e0bf4f058d33.png)

yeya24:
[...]

PR #662, opened 2020-06-29
Store experiment when it was created

### LINKED ISSUE #657: Fail to archive experiment
## Bug Report

When to archive the experiment, we found that the experiment may be deleted by `chaos-operator`, so this action will cause the archive action to fail.

cwen0:

Fix #657 #557 

* store experiment when it was created 
* add `archived` field to archive_eperiment struct, when some experiment was deleted and update `archived` to `true`


yeya24:
TBH I don't think it is a good choice. The experiment object can be updated/changed after it is created. Then do we want to update the archive experiment in db every time?

cwen0:
> TBH I don't think it is a good choice. The experiment object can be updated/changed after it is created. Then do we want to update the archive experiment in db every time?

Yes, I think we can update it in DB every time.

cwen0:
@YangKeao @yeya24 PTAL, thanks!

PR #672, opened 2020-07-02
[update] update /api/events and add kind and finishTime

fewdan:
Signed-off-by: fewdan <fuyihao@syvip.com>
update /api/events and add kind and finishTime

update /api/events and add kind and finishTime

### Check List

Tests

- No code

Code changes

- Has Go code change

Side effects

- NO

Related changes

- NO

```release-note
NONE
```


pkg/apiserver/event/event.go
```diff
@@ -66,20 +66,26 @@ func Register(r *gin.RouterGroup, s *Service) {
...
-	experimentName := c.Query("experimentName")
-	experimentNamespace := c.Query("experimentNamespace")
-	uid := c.Query("uid")
+	filter := core.Filter{
+		PodName:             c.Query("podName"),
+		PodNamespace:        c.Query("podNamespace"),
+		StartTimeBeginStr:   c.Query("startTimeBegin"),
+		StartTimeEndStr:     c.Query("startTimeEnd"),
+		FinishTimeBeginStr:  c.Query("finishTimeBegin"),
+		FinishTimeEndStr:    c.Query("finishTimeEnd"),
```
yeya24, line 75: I think it is too complicated now... :) Do we need to have begin and end range for start and finish? TBH, the use case is not much.
Even in Grafana, it only uses start time and end time, it is not necessary to support range for both.

fewdan: Because the time saved in the database is accurate to the second, if the interval query is not provided and the user directly specifies a time, then the time may not be accurate to the second. If it is an interval, the user does not need to provide the precise time, and the experience may be better.

yeya24: @g1eny0ung Hello, would you mind sharing your idea about this? Since you are a direct user of this API.

g1eny0ung: @yeya24 @fewdan First, I'm okay with this implementation too. But actually...I also think it's a little complicated for "users". The users represent developers who want to use chaos mesh API to develop something.

And for front-end usage, I think I will use the `StartTimeBeginStr` and `FinishTimeEndStr`, According to my understanding, at the time point of `StartTimeEndStr`, this event has already started. Same for `FinishTimeBeginStr`.

yeya24: @fewdan I think the precision is not very important in this API. So it is okay to just use just startTime and finishTime here. As @g1eny0ung also just needs the two time.

Adding the range time support for start and finish is weird and will cause confusion.

pkg/store/event/event.go
```diff
@@ -262,44 +262,83 @@ func (e *eventStore) DeleteIncompleteEvents(_ context.Context) error {
...
+			}
+		}
 	}
 	if err != nil {
 		return resList, err
 	}
 
-	if startTimeStr != "" {
-		startTime, err = time.Parse(time.RFC3339, strings.Replace(startTimeStr, " ", "+", -1))
+	if filter.StartTimeBeginStr != "" {
```
cwen0, line 293: If we need to check these variables, we should check them before selecting data from db.

pkg/store/event/event.go
```diff
@@ -262,44 +262,83 @@ func (e *eventStore) DeleteIncompleteEvents(_ context.Context) error {
...
-		if experimentNamespace != "" && event.Namespace != experimentNamespace {
+		if filter.UID != "" && event.ExperimentID != filter.UID {
 			continue
 		}
-		if uid != "" && event.ExperimentID != uid {
+		if filter.Kind != "" && event.Kind != filter.Kind {
 			continue
 		}
-		if startTimeStr != "" && event.StartTime.Before(startTime) && !event.StartTime.Equal(startTime) {
+		if filter.StartTimeBeginStr != "" && event.StartTime.Before(startTimeBegin) && !event.StartTime.Equal(startTimeBegin) {
```
cwen0, line 332: Can we filter these by SQL?

pkg/store/event/event.go
```diff
@@ -262,44 +262,65 @@ func (e *eventStore) DeleteIncompleteEvents(_ context.Context) error {
...
+			continue
+		}
+		if filter.FinishTimeEndStr != "" && event.FinishTime.After(finishTimeEnd) && !event.FinishTime.Equal(finishTimeEnd) {
 			continue
 		}
```
yeya24, line 325: I think the all the if statements above are not necessary since you are already using `constructQueryArgs` to query from the database.

pkg/store/event/event.go
```diff
@@ -262,44 +262,65 @@ func (e *eventStore) DeleteIncompleteEvents(_ context.Context) error {
...
-
-	if podName != "" {
-		resList, err = e.ListByPod(context.Background(), podNamespace, podName)
-	} else if podNamespace != "" {
-		resList, err = e.ListByNamespace(context.Background(), podNamespace)
-	} else if uid != "" {
-		resList, err = e.ListByUID(context.Background(), uid)
-	} else {
-		resList, err = e.List(context.Background())
+	var startTimeBegin, finishTimeEnd time.Time
```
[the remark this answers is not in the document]
yeya24: Updated: these vars are not necessary IMO

pkg/store/event/event.go
```diff
@@ -262,44 +262,65 @@ func (e *eventStore) DeleteIncompleteEvents(_ context.Context) error {
...
+
+	if filter.StartTimeBeginStr != "" {
+		startTimeBegin, err = time.Parse(time.RFC3339, strings.Replace(filter.StartTimeBeginStr, " ", "+", -1))
```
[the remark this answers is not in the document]
fewdan: OK. So now the RFC3339 format string is supported first, and then other formats will be added later?

pkg/core/event.go
```diff
@@ -83,3 +83,15 @@ type PodRecord struct {
 	Message   string
 	Action    string
 }
+
+// Filter represents the filter to list events
+type Filter struct {
+	PodName             string
+	PodNamespace        string
+	StartTimeBeginStr   string
+	FinishTimeEndStr    string
```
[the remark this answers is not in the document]
fewdan: I will modify them.

pkg/store/event/event.go
fewdan: @yeya24
These "if" are still needed, because if we specify "podName" and "podNamespace", then we will call "ListByPod" or "ListByNamespace". In this case, "constructQueryArgs" is not used, so it is not filtered according to other parameters. Therefore, it is necessary to add "if" in the following loop to judge.

g1eny0ung: Check these two situations to judge whether to run these extra `if` checks?

Like:

```go
if filter.PodName != "" || filter. PodNamespace != "" {
        // if ...
        // if ...
}
```

pkg/store/event/event.go
```diff
@@ -262,44 +262,65 @@ func (e *eventStore) DeleteIncompleteEvents(_ context.Context) error {
...
+		if err != nil {
+			return nil, fmt.Errorf("the format of the startTimeBegin is wrong")
+		}
 	}
-	if err != nil {
-		return resList, err
+	if filter.FinishTimeStr != "" {
+		finishTime, err = time.Parse(time.RFC3339, strings.Replace(filter.FinishTimeStr, " ", "+", -1))
+		if err != nil {
+			return nil, fmt.Errorf("the format of the finishTimeEnd is wrong")
```
yeya24, line 280: Please also update the error

pkg/store/event/event.go
```diff
@@ -262,44 +262,65 @@ func (e *eventStore) DeleteIncompleteEvents(_ context.Context) error {
...
-	} else if uid != "" {
-		resList, err = e.ListByUID(context.Background(), uid)
-	} else {
-		resList, err = e.List(context.Background())
+	var startTime, finishTime time.Time
+
+	if filter.StartTimeStr != "" {
+		startTime, err = time.Parse(time.RFC3339, strings.Replace(filter.StartTimeStr, " ", "+", -1))
+		if err != nil {
+			return nil, fmt.Errorf("the format of the startTimeBegin is wrong")
```
yeya24, line 274: Please also update the error

pkg/store/event/event.go
yeya24: I have tested the code and seems there is a bug when comparing event time caused by different timezones.

The `event.FinishTime` is stored in the database with current timezone, for example in my case it is `2020-07-07 09:06:20 -0400 -0400`. But `finishTime` is parsed to UTC time `2020-07-07 09:10:11.826 +0000 UTC`, so during the comparison, event.FinishTime.After(finishTime) returns true and all the events are filtered.

pkg/store/event/event.go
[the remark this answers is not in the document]
yeya24: Yes. We can add it later. Just follow what Prometheus does here https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L1540.

fewdan:
@yeya24 
Added an "events/dry" interface, you can call this interface for grafana. "events/dry" does not return pod records.

fewdan:
@yeya24 
Added an "events/dry" interface, you can call this interface for grafana. "events/dry" does not return pod records.

pkg/store/event/event.go
```diff
@@ -262,45 +262,68 @@ func (e *eventStore) DeleteIncompleteEvents(_ context.Context) error {
...
+		if err != nil {
+			return nil, fmt.Errorf("the format of the finishTime is wrong")
+		}
 	}
 
-	if startTimeStr != "" {
-		startTime, err = time.Parse(time.RFC3339, strings.Replace(startTimeStr, " ", "+", -1))
-		if err != nil {
-			return nil, fmt.Errorf("the format of the time is wrong")
+	if filter.PodName != "" {
```
cwen0, line 284: Can we use this method: 

```
if podName != ""{
      db = db.Where(pdname=?, podname) 
} 

if ns != "" { 
        db = db.Where(xxx=?, )
}  

if startime != "" { 
          db = db.Where(xxxx) 
} 

if len(args) != 0 { 
       db = db.Where(query, args....) 
} 

db.Find(&resList)
```

yeya24:
@fewdan  Great work! Thanks! I will do some testings and review again today.

fewdan:
@yeya24
I still have a small modification, you can wait for me to complete the modification before testing.

pkg/apiserver/event/event.go
```diff
@@ -56,30 +56,74 @@ func Register(r *gin.RouterGroup, s *Service) {
...
+// @Produce json
+// @Param startTime query string false "The start time of events"
+// @Param endTime query string false "The end time of events"
+// @Param experimentName query string false "The name of the experiment"
+// @Param experimentNamespace query string false "The namespace of the experiment"
+// @Param kind query string false "kind" Enums(PodChaos, IoChaos, NetworkChaos, TimeChaos, KernelChaos, StressChaos)
+// @Success 200 {array} core.Event
+// @Router /api/events/dry [get]
+// @Failure 500 {object} utils.APIError
+func (s *Service) listDryEvents(c *gin.Context) {
```
yeya24, line 117: It is not necessary to have a new API for this. One additional parameter is enough.

pkg/apiserver/event/event.go
```diff
@@ -56,30 +56,74 @@ func Register(r *gin.RouterGroup, s *Service) {
...
+	filter := core.Filter{
+		StartTimeStr:        c.Query("startTime"),
+		FinishTimeStr:       c.Query("finishTime"),
+		ExperimentName:      c.Query("experimentName"),
+		ExperimentNamespace: c.Query("experimentNamespace"),
+		Kind:                c.Query("kind"),
+	}
+
+	eventList, err := s.event.DryListByFilter(context.Background(), filter)
```
yeya24, line 126: This function is not necessary, it is okay to reuse events.ListByFilter

PR #673, opened 2020-07-02
[add] add force delete

fewdan:
Signed-off-by: “fewdan” <fuyihao@syvip.com>

add force delete

add force delete

### Check List

Tests

- No code

Code changes

- Has Go code change

Side effects

- NO

Related changes

- NO

```release-note
NONE
```

### DIFF
--- a/pkg/apiserver/experiment/experiment.go
+++ b/pkg/apiserver/experiment/experiment.go
@@ -25,11 +25,13 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	"github.com/pingcap/chaos-mesh/api/v1alpha1"
+	"github.com/pingcap/chaos-mesh/controllers/common"
 	"github.com/pingcap/chaos-mesh/pkg/apiserver/utils"
 	"github.com/pingcap/chaos-mesh/pkg/config"
 	"github.com/pingcap/chaos-mesh/pkg/core"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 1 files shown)


pkg/apiserver/experiment/experiment.go
```diff
@@ -885,6 +890,26 @@ func (s *Service) deleteExperiment(c *gin.Context) {
...
+
+		annotations := chaosMeta.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+		annotations[common.AnnotationCleanFinalizer] = common.AnnotationCleanFinalizerForced
+		chaosMeta.SetAnnotations(annotations)
+		if err := s.kubeCli.Update(context.Background(), chaosKind.Chaos); err != nil {
+			c.Status(http.StatusInternalServerError)
+			_ = c.Error(utils.ErrInternalServer.WrapWithNoMessage(fmt.Errorf("update chaos annotation error")))
```
yeya24, line 908: A small nit here. I am wondering if we can make this error message more clear, since we are doing a force delete via adding an annotation, the real deletion is performed at the controller side.

But users might feel strange when they see `update chaos  annotation error` when they think they just want to delete a chaos.

pkg/apiserver/experiment/experiment.go
```diff
@@ -852,6 +854,7 @@ func (s *Service) getExperimentDetail(c *gin.Context) {
 // @Param namespace path string true "namespace"
 // @Param name path string true "name"
 // @Param kind path string true "kind" Enums(PodChaos, IoChaos, NetworkChaos, TimeChaos, KernelChaos, StressChaos)
+// @Param force path string true "force" Enums(true, false)
```
yeya24, line 857: Do we want to make this true by default?

pkg/apiserver/experiment/experiment.go
fewdan: Okay, I will modify this information.

pkg/apiserver/experiment/experiment.go
fewdan: I think it is possible to set a default value, so would "false" be more reasonable?

yeya24: Yes, we are using non-force delete before, in the frontend as well. I think we should use the same. WDYT?

PR #679, opened 2020-07-07
chore: complement chaos

g1eny0ung:

This PR complement the experiment creation, now you can create `IO Chaos`, `Kernel Chaos`, `Time Chaos` and `Stress Chaos` from the dashboard.

Similarly, you can also use YAML to create any type of experiment.

For local review, please first run `yarn` to install updated deps.


yeya24:
I really like this refactoring to put labels and annotations at the first page instead of the scope page. It is much clear now that these are labels of the chaos experiment, not the label selector in the scope

Screenshot from 2020-07-08 15-22-16

ui/src/lib/formikhelpers.ts
```diff
@@ -129,6 +134,65 @@ export function yamlToExperiments(yamlObj: any): Experiment {
...
+    spec.stressors.cpu = {
+      ...defaultExperimentSchema.target.stress_chaos.stressors.cpu,
+      ...spec.stressors.cpu,
+    }
+
+    spec.stressors.memory = {
+      ...defaultExperimentSchema.target.stress_chaos.stressors.memory,
+      ...spec.stressors.memory,
+    }
+  }
```
yeya24, line 174: About stress chaos @cwen0 Maybe we should refactor the API and add actions such  as `cpu, memory, mixes` just like IO chaos.

These two configs can both set at the same time, or you can only set cpu and memory. You can find an example here 
https://github.com/pingcap/chaos-mesh/blob/master/examples/burn-cpu.yaml

Screenshot from 2020-07-08 15-58-06

ui/src/components/NewExperiment/Stepper/Target/IO.tsx
```diff
@@ -0,0 +1,120 @@
...
+        {actions.map((option: string) => (
+          <MenuItem key={option} value={option}>
+            {toTitleCase(option)}
+          </MenuItem>
+        ))}
+      </SelectField>
+
+      {values.target.io_chaos.action !== '' && (
+        <AdvancedOptions>
+          {values.target.io_chaos.action === ('delay' || 'mixed') && (
```
yeya24, line 65: I am not really sure but this seems like a bug. Should it be `(values.target.io_chaos.action === 'delay' || values.target.io_chaos.action === 'mixed')` ?

ui/src/components/NewExperiment/Stepper/Target/IO.tsx
```diff
@@ -0,0 +1,120 @@
...
+            <TextField
+              id="target.io_chaos.delay"
+              name="target.io_chaos.delay"
+              label="Delay"
+              helperText="The value of delay action"
+            />
+          )}
+
+          {values.target.io_chaos.action === ('errno' || 'mixed') && (
```
yeya24, line 74: ditto

g1eny0ung:
> Overall this looks really awesome! This is a huge effort, only with some small nits about help messages. I don't think I am able to dive deeply to review the frontend code. If someone else can help then it would be great!

Unfortunately, now I am the only person that mainly dev with the front end. 🤣

So far, this is not a big problem. In the early stage of the dashboard, one person can maintain a rapid development status. I hope that more people can participate in the future, so that I can be free to do some other work, such as the API part.

ui/src/components/NewExperiment/Stepper/Target/IO.tsx
g1eny0ung: Ah...a place that shouldn’t be wrong, I’ll fix it right away!

ui/src/components/NewExperiment/Stepper/Target/Stress.tsx
```diff
@@ -0,0 +1,70 @@
...
+        name="target.stress_chaos.stressors.memory.size"
+        label="Size"
+        helperText="Memory size"
+      />
+      <LabelField
+        id="target.stress_chaos.stressors.memory.options"
+        name="target.stress_chaos.stressors.memory.options"
+        label="Options of stress-ng"
+        helperText="Type string and end with a space to generate the stress-ng options"
+      />
```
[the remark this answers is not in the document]
g1eny0ung: Got it now.

ui/src/components/NewExperiment/Stepper/Target/IO.tsx
```diff
@@ -0,0 +1,120 @@
...
+            <TextField
+              id="target.io_chaos.errno"
+              name="target.io_chaos.errno"
+              label="Errno"
+              helperText="The value of errno action"
```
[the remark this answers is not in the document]
g1eny0ung: I agree.

ui/src/components/NewExperiment/Stepper/Target/Kernel.tsx
```diff
@@ -0,0 +1,100 @@
...
+        helperText="The fails with probability"
+        InputProps={{
+          endAdornment: <InputAdornment position="end">%</InputAdornment>,
+        }}
+      />
+      <TextField
+        type="number"
+        id="target.kernel_chaos.fail_kernel_req.times"
+        name="target.kernel_chaos.fail_kernel_req.times"
+        helperText="The max times of fails"
```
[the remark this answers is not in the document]
g1eny0ung: Yep. Should be `failures`.

ui/src/components/NewExperiment/Stepper/Target/Time.tsx
```diff
@@ -0,0 +1,39 @@
...
+  }, [])
+
+  return (
+    <>
+      <LabelField
+        id="target.time_chaos.clock_ids"
+        name="target.time_chaos.clock_ids"
+        label="Clock ids"
+        helperText="Type string and end with a space to generate the clock ids"
+      />
```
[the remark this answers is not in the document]
g1eny0ung: Got it now.

g1eny0ung:
@yeya24 I didn't reply to all comments. I will mark as resolved directly when I fixed.

ui/src/lib/formikhelpers.ts
cwen0: > About stress chaos @cwen0 Maybe we should refactor the API and add actions such as `cpu, memory, mixes` just like IO chaos.
> 
> These two configs can both set at the same time, or you can only set cpu and memory. You can find an example here
> https://github.com/pingcap/chaos-mesh/blob/master/examples/burn-cpu.yaml @g1eny0ung. In the current page, I have to set both CPU and memory stress
> 
> Screenshot from 2020-07-08 15-58-06

I agree with this. We can add the `action` field.

g1eny0ung: I'm already giving StressChaos a select field to select the action. It's only used in the front end, the API will not be affected.

For API refactoring, we can open an issue to tracking it. It's not very urgent, we can finish it in the future.

Issue #676, opened 2020-07-03
NetworkChaos Experiment Inject Latency Between Pod and External Endpoint


DannyeB:
## Feature Request

**Is your feature request related to a problem? Please describe:**
I would like to be able to inject latency between a pod in the cluster and an external endpoint. This might be an externally hosted database or Redis instance etc

**Describe the feature you'd like:**
In the NetworkChaos experiment, I should be able to use an external URL as the target. Any traffic coming from my selected pods going to the external URL should get latency applied. This is what I was thinking it would look like:

```yaml
apiVersion: pingcap.com/v1alpha1
kind: NetworkChaos
metadata:
  name: block-external-sample
  namespace: chaos-testing
spec:
  action: delay
  mode: all
  selector:
    labelSelectors:
      app: microserviceb
    namespaces:
      - default
  direction: to
  target:
    selector:
      external: "https://mydburl.com"
  delay:
    latency: "300ms"
    correlation: "80"
    jitter: "90ms"

```

**Describe alternatives you've considered:**
I have thought about setting up a proxy pod inside my cluster and applying the latency to that. However, then I need to change my connection strings and that is not a very practical solution for me.

**Teachability, Documentation, Adoption, Migration Strategy:**
A lot of cloud services I work with, in Kubernetes use externally hosted solutions for the stateful parts such as databases. We need to cover the failure scenarios when these go down or the connections become slow.

nunziox:
[This is not a contribution]

I was going to open the same issue :) We really would need this as well.

nunziox:
What is the ETA for this ?

YangKeao:
@nunziox I am working on this issue. I hope it can be merged this week :yum:

PR #680, opened 2020-07-07
External target for netem and partition chaos

YangKeao:

#676 , you can see an example here:

```
kind: NetworkChaos
metadata:
  name: network-delay-example
  namespace: chaos-testing
spec:
  action: delay
  mode: one
  selector:
    labelSelectors:
  delay:
    latency: "90ms"
    correlation: "25"
    jitter: "90ms"
  direction: to
  externalTargets:
    - "8.8.8.8"
    - "www.google.com"
    - "8.8.0.0/16"
  duration: "10s"
  scheduler:
    cron: "@every 15s"
```

```
kind: NetworkChaos
metadata:
  name: network-partition-example
  namespace: chaos-testing
spec:
  action: partition
  mode: one
  selector:
    labelSelectors:
  direction: to
  externalTargets:
    - "8.8.8.8"
    - "8.8.0.0/16"
    - "www.google.com"
  duration: "10s"
  scheduler:
    cron: "@every 15s"
```

Add an `externalTarget` field to specify external endpoints as targets. This extra field was used to avoid breaking compatibility, in the future we may need to redesign these fields.

### Check List

Tests

I have tested this feature manually. With a simple ubuntu pod and `ping` external targets, everything works well.

Code changes

- Has Go code change

Related changes

```release-note
Support external target for netem and partition chaos
```


api/v1alpha1/networkchaos_types.go
```diff
@@ -166,6 +166,10 @@ type NetworkChaosSpec struct {
 	// Target represents network target, this applies on netem and network partition action
 	// +optional
 	Target *Target `json:"target,omitempty"`
+
+	// ExternalTargets represents network targets outside k8s
+	// +optional
+	ExternalTargets []string `json:"externalTargets,omitempty"`
```
cwen0, line 172: Can we use some way to represent the range of IPs? such as "192.168.0.0/24"

YangKeao: It supports CIDR input now :smile:

controllers/networkchaos/netem/utils.go
```diff
@@ -0,0 +1,39 @@
+// Copyright 2019 PingCAP, Inc.
```
fewdan, line 1: s/2019/2020/g

controllers/networkchaos/netutils/cidr.go
```diff
@@ -0,0 +1,58 @@
+// Copyright 2019 PingCAP, Inc.
```
fewdan, line 1: ditto

helm/chaos-mesh/values.yaml
```diff
@@ -21,7 +21,7 @@ controllerManager:
   replicaCount: 1
 
   image: pingcap/chaos-mesh:latest
-  imagePullPolicy: Always
+  imagePullPolicy: Never
```
cwen0, line 24: Fix this?

YangKeao: Fixed

PR #696, opened 2020-07-14
Use a docker container to generate protobuf related file

### LINKED ISSUE #684: Generate protobuf/grpc related files in docker
## Feature Request

Nowadays different developers have different developing environments. The versions of their protoc/proto-gen-go/proto-gen-go-grpc differ, which may have hidden compatibility problems.

Generate protobuf/grpc related file in docker, which will give a stable compiling environment.

We also need to push these generated codes to github, so `--mount $(pwd):/mnt -user $(id -u):$(id -g)` is needed.

YangKeao:

Fix #684 

The protobuf related environment is quite complicated, especially with `protoc-gen-go`. Currently, the architecture of `protoc-gen-go` is changing rapidly, and it will be split into two package `protoc-gen-go` and `protoc-gen-go-grpc`, which may bring in serious compatibility problem (which has cost me a whole Friday to solve).

With this PR merged, we can generate protobuf and grpc related go files without being affected by the package version of host.

### Check List

Tests

- Manual test 

`make proto` will generate related files

Code changes


Makefile
```diff
@@ -246,8 +249,19 @@ yaml: manifests ensure-kustomize
 	$(KUSTOMIZE_BIN) build config/default > manifests/crd.yaml
 
 # Generate Go files from Chaos Mesh proto files.
+ifeq ($(IN_DOCKER),1)
```
cwen0, line 252: `ifneq`?

YangKeao: No. `ifeq ($(IN_DOCKER),1)` means we need to run the following command in the container. And if we are on the host, we should use `docker run` to enter a container.

hack/protoc/Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+
+ENV GO111MODULE on
+ENV PROTOC_VERSION 3.12.3
+
+RUN apk add --no-cache wget unzip protoc protobuf-dev make bash
+
+RUN go get github.com/golang/protobuf/protoc-gen-go@v1.2.0
+
+RUN mkdir /.cache
+RUN chmod -R 777 /.cache
```
yeya24, line 14: Please add one new line then it is good to go!

PR #702, opened 2020-07-15
[add] Update Incomplete Events

fewdan:

If chaos is deleted before an event is over, then the incomplete event would be stored in dbtastore, which means the event would never save the finish_time.
UpdateIncompleteEvents can update the finish_time when the chaos is deleted.

add UpdateIncompleteEvents

### Check List

Tests

- No code

Code changes

- Has Go code change

Side effects

- NO

Related changes

- NO

```release-note
NONE
```


pkg/store/event/event.go
```diff
@@ -423,6 +423,16 @@ func (e *eventStore) getUID(_ context.Context, ns, name string) (string, error)
 	return UID, nil
 }
 
+// UpdateIncompleteEvents updates the incomplete event by the namespace and name
+func (e *eventStore) UpdateIncompleteEvents(_ context.Context, ns, name string) error {
+	return e.db.Model(core.Event{}).
+		Where(
+			"namespace = ? and experiment = ? and finish_time IS NULL",
```
cwen0, line 430: Do we need to handle the `Not found` error?

fewdan: I just tried it and there is no "not found error". I don't know if it is because update has no such errors.

cwen0: Got

PR #722, opened 2020-07-20
chaosdaemon: fix the bug of goroutine leak.

johncming:

It's a bug that lead to goroutine leak if an error return before `g.Wait()`

Move the `g.Go()` code to right postion.

### Check List

Tests

- No code

Code changes

- Has Go code change

```release-note
NONE
```

### DIFF
--- a/pkg/chaosdaemon/server.go
+++ b/pkg/chaosdaemon/server.go
@@ -43,6 +43,16 @@ type Config struct {
 	Profiling bool
 }
 
+// Get the http address
+func (c *Config) HttpAddr() string {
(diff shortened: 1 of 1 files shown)


cwen0:
@johncming Thanks for your contribution! Can you fix  [DCO](https://github.com/chaos-mesh/chaos-mesh/pull/722/checks?check_run_id=889990633)?

johncming:
> @johncming Thanks for your contribution! Can you fix [DCO](https://github.com/chaos-mesh/chaos-mesh/pull/722/checks?check_run_id=889990633)?

DCO added.

PR #729, opened 2020-07-22
[update] update experiment/detail and archives/detail

fewdan:
1. update experiment/detail, add "ExperimentBase"
2. update archives/detail, change the format to something like "experiment/detail".


pkg/apiserver/archive/experiment.go
```diff
@@ -157,7 +212,44 @@ func (s *Service) experimentDetail(c *gin.Context) {
...
+			Action:     data.Action,
+			UID:        data.UID,
+			StartTime:  data.StartTime,
+			FinishTime: data.FinishTime,
+			Archived:   data.Archived,
+		},
+		ExperimentInfo: info,
+	}
+
+	c.JSON(http.StatusOK, expDetail)
```
g1eny0ung, line 252: Seems this part of code and above are dealing with the same thing, can we bump them into a common function?

pkg/apiserver/experiment/experiment.go
```diff
@@ -505,26 +387,39 @@ func (s *Service) getPodChaosDetail(namespace string, name string) (ExperimentIn
...
+				Kind:      chaos.Kind,
+				Namespace: chaos.Namespace,
+				Name:      chaos.Name,
+			},
+			Created: chaos.GetChaos().StartTime.Format(time.RFC3339),
+			Status:  chaos.GetChaos().Status,
+			UID:     chaos.GetChaos().UID,
+		},
+		ExperimentInfo: info,
+	}, nil
```
g1eny0ung, line 403: Seems `ExperimentDetail` can be bumped too.

pkg/core/archive_experiment.go
```diff
@@ -63,21 +62,419 @@ type ArchiveExperiment struct {
...
+	StressngStressors string              `json:"stressng_stressors,omitempty"`
+}
+
+func (e *ArchiveExperiment) ParsePodChaos() (ExperimentInfo, error) {
+	chaos := &v1alpha1.PodChaos{}
+	if err := json.Unmarshal([]byte(e.Experiment), &chaos); err != nil {
+		return ExperimentInfo{}, err
+	}
+
+	info := ExperimentInfo{
```
cwen0, line 216: Can we abstract a function to parse `ExperimentInfo`?

pkg/apiserver/archive/experiment.go
```diff
@@ -157,7 +212,44 @@ func (s *Service) experimentDetail(c *gin.Context) {
...
+	case v1alpha1.KindIOChaos:
+		info, err = data.ParseIOChaos()
+	case v1alpha1.KindNetworkChaos:
+		info, err = data.ParseNetworkChaos()
+	case v1alpha1.KindTimeChaos:
+		info, err = data.ParseTimeChaos()
+	case v1alpha1.KindKernelChaos:
+		info, err = data.ParseKernelChaos()
+	case v1alpha1.KindStressChaos:
+		info, err = data.ParseStressChaos()
```
cwen0, line 227: Add `default` option

pkg/apiserver/archive/experiment.go
fewdan: These two are not the same, one is to return a list of archives, and the other is to return a specific archive.

g1eny0ung: Got it, my mistake. 🤣

PR #738, opened 2020-07-24
Use bigheap to inject heap stress

### LINKED ISSUE #737: memory in StressChaos is virtual memory but not real memory
## Bug Report

**What did you do?**

Apply StressChaos with memory stressors.

Increase the memory usage and cause OOM.

**What did you see instead?**

Increase virtual memory usage, which is not a regular scenario.

YangKeao:

Fix #737 

Use `--bigheap` to inject memory stress rather than `--vm` which is virtual memory.

### Check List

Tests

- Manual test

Code changes

- Has Go code change

Side effects

Remove `Size` field of memory stressor, because we cannot set the total size of `bigheap` stressor, only the speed is configurable.

Maybe we should provide some options to choose the memory stressors between `brk`, `bigheap` and `stack`, do you think it is needed?


api/v1alpha1/stresschaos_types.go
```diff
@@ -233,19 +232,7 @@ type Stressors struct {
...
-				if err != nil {
-					return "", err
-				}
-				stressors += fmt.Sprintf(" --vm-bytes %d", size)
-			} else {
-				stressors += fmt.Sprintf("--vm-bytes %s",
-					in.MemoryStressor.Size)
-			}
-		}
+		stressors += fmt.Sprintf(" --bigheap %d", in.MemoryStressor.Workers)
```
cwen0, line 235: Can we control how much memory it takes up?

YangKeao: No. The stress-ng doesn't provide such arguments. We can only control how fast it eats up ram (but these arguments are also hard to understand and calculate).

cwen0:
@fewdan @g1eny0ung Do we also need to update Chaos Dashboard after this pr is merged?

g1eny0ung:
@cwen0 Yep. Due to the memory `size` field will be unavailable. It's needed to change both the API and the dashboard interface.

fewdan:
The dashboard needs to be modified, but not many changes.

PR #742, opened 2020-07-27
[update]Add limit in api/events

fewdan:

Add "limit" in "api/events" to indicate the maximum length of events_list.


pkg/core/event.go
```diff
@@ -73,27 +71,33 @@ type EventStore interface {
 
 // Event represents an event instance.
 type Event struct {
-	gorm.Model
```
yeya24, line 76: Why `gorm.Model` is not needed?

fewdan: The content in gorm.Model is "ID", "CreatedAt", "UpdatedAt", "DeletedAt". I took out these 4 things directly, so that it is convenient to add the "json" tag.

yeya24: Got it.

pkg/store/event/event.go
```diff
@@ -341,8 +375,15 @@ func (e *eventStore) DryListByFilter(_ context.Context, filter core.Filter) ([]*
 		resList []*core.Event
 		err     error
 		db      *dbstore.DB
+		limit   int
 	)
 
+	if filter.LimitStr != "" {
```
yeya24, line 381: The two functions `DryListByFilter` and `ListByFilter` has many duplicates because they share the same logic, can we abstract some of them to another function?

fewdan: "DryListByFilter" is mainly used by grafana, and may be accessed more frequently than "ListByFilter". In order not to affect "ListByFilter", it is split out. To merge into the same function is not complicated to implement, mainly based on stability considerations.

yeya24: I am not saying to combine the two functions into one. But the two share some parts of duplicate code (except the pod records part). Is it okay to pull the duplicate part out and make it another function?

yeya24: But this actually is not part of this pr, so we can have an issue for it first and not do it in this pr. It depends on you

fewdan: Oh oh, I got it. I just got it wrong. Let me change it in the follow-up pr.

PR #747, opened 2020-07-28
[update] update website/doc , remove '$' in shell

fewdan:

update website/doc , remove '$' in shell


pkg/apiserver/event/event.go
```diff
@@ -71,6 +71,7 @@ func Register(r *gin.RouterGroup, s *Service) {
 // @Param experimentNamespace query string false "The namespace of the experiment"
 // @Param uid query string false "The UID of the experiment"
 // @Param kind query string false "kind" Enums(PodChaos, IoChaos, NetworkChaos, TimeChaos, KernelChaos, StressChaos)
+// @Param limit query string false "The max length of events list"
```
yeya24, line 74: Are these limit changes related?

fewdan: @yeya24   please see https://github.com/chaos-mesh/chaos-mesh/pull/742  .
I made a mistake when switching branches.

PR #768, opened 2020-07-31
Refine build system

YangKeao:

Accelerate the build process of chaos-mesh:

1. Ignore most files in `.dockerignore` to avoid unintended cache miss.

2. Use buildkit to use cache volume.

The total build time decrease from 210s to 36s :beers: 

**require: docker version >= 18.09**

### Check List

Tests

- Manual test

Code changes

- Has Go code change

Side effects

Related changes


YangKeao:
> LGTM
> 
> Does this work with the lower version of Docker?

Only work with docker >= 18.09

PR #775, opened 2020-08-03
add +optional comment

vincent178:
Signed-off-by: vincent178 <vh7157@gmail.com>

add `// +optional` comment for optional fields.


YangKeao:
It seems that `optional` has no effect on array items... But as a comment it's ok :sweat_smile:

PR #776, opened 2020-08-03
Use bigheap to inject heap stress

### LINKED ISSUE #737: memory in StressChaos is virtual memory but not real memory

YangKeao:

Fix #737 

Use `--bigheap` to inject memory stress rather than `--vm` which is virtual memory.

### Check List

Tests

- Manual test

Code changes

- Has Go code change

Side effects

Remove `Size` field of memory stressor, because we cannot set the total size of `bigheap` stressor, only the speed is configurable.

Maybe we should provide some options to choose the memory stressors between `brk`, `bigheap` and `stack`, do you think it is needed?


YangKeao:
Sorry, I have deleted the branch accidentally :sob: . So I have to recreate a new one.

PR #781, opened 2020-08-03
[update] update Reconcile

fewdan:

If "cron" is modified, it will take effect immediately.


controllers/twophase/types.go
```diff
@@ -162,15 +162,24 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		nextStart, err := utils.NextTime(*chaos.GetScheduler(), status.Experiment.StartTime.Time)
+		if err != nil {
+			r.Log.Error(err, "failed to get next start time")
+			return ctrl.Result{}, err
+		}
 		nextTime := chaos.GetNextStart()
 
-		if !chaos.GetNextRecover().IsZero() && chaos.GetNextRecover().Before(nextTime) {
-			nextTime = chaos.GetNextRecover()
+		if nextStart.Equal(nextTime) {
```
cwen0, line 172: Though we don't update the `Scheduler`, we also can't make sure `nextStart` equal to `nextTime'. 

You can check [P144](https://github.com/chaos-mesh/chaos-mesh/blob/master/controllers/twophase/types.go#L144) and [P212](https://github.com/chaos-mesh/chaos-mesh/blob/master/controllers/twophase/types.go#L212). We generate the `nextStart` using `now`, but we set `status.Experiment.StartTime = &metav1.Time{Time: time.Now()}`.  
`now` may not be the same value as `time.Now()`.

controllers/twophase/types.go
```diff
@@ -136,41 +137,56 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		if startTime, err = applyAction(ctx, r, req, dur, chaos); err != nil {
 			return ctrl.Result{Requeue: true}, err
 		}
 
 	} else if chaos.GetNextStart().Before(now) {
-		nextStart, err := utils.NextTime(*chaos.GetScheduler(), now)
+		tempStart, err := utils.NextTime(*chaos.GetScheduler(), now)
 		if err != nil {
-			r.Log.Error(err, "failed to get next start time")
+			r.Log.Error(err, "failed to calculate the temp start time")
```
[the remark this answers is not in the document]
fewdan: Understand, I will modify the log information.

controllers/twophase/types.go
```diff
@@ -136,41 +137,56 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 		}
 
-		nextRecover := now.Add(*duration)
-		if nextStart.Before(nextRecover) {
+		tempRecover := now.Add(*duration)
+		if tempStart.Before(tempRecover) {
 			err := fmt.Errorf("nextRecover shouldn't be later than nextStart")
-			r.Log.Error(err, "nextRecover is later than nextStart. Then recover can never be reached",
-				"nextRecover", nextRecover, "nextStart", nextStart)
+			r.Log.Error(err, "Then recover can never be reached. It may be because cron is greater than duration.")
```
yeya24, line 154: This sentence doesn't make sense to me. Can you be more specific here? It is not good to put `may be` in the error log.

yeya24:
Another thing is that can you please add some test cases for the `Reconcile` method? Then it would help us review the pr and know what are changed as well.

WangXiangUSTC (commented):
LGTM

cwen0 (dismissed):

controllers/twophase/types.go
```diff
@@ -136,41 +137,58 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+		if tempStart.Before(tempRecover) {
 			err := fmt.Errorf("nextRecover shouldn't be later than nextStart")
-			r.Log.Error(err, "nextRecover is later than nextStart. Then recover can never be reached",
-				"nextRecover", nextRecover, "nextStart", nextStart)
+			r.Log.Error(err, "Then recover can never be reached.", "scheduler", *chaos.GetScheduler(), "duration", *duration)
 			return ctrl.Result{}, err
 		}
 
-		if err := applyAction(ctx, r, req, *duration, chaos); err != nil {
+		if startTime, err = applyAction(ctx, r, req, *duration, chaos); err != nil {
```
YangKeao, line 158: Is the `startTime` here added to make sure that `nextStart` is generated with `status.Experiment.StartTime`, so that we can compare it in the following code?

YangKeao: But why not get it from `status.Experiment.StartTime` directly?

fewdan: Thanks, I have modified it.

YangKeao (approved):
LGTM

ti-srebot (approved):

PR #785, opened 2020-08-05
fix: correct the usage of pods preview

g1eny0ung:

This PR changes the pods' table in `NewExperiment Scope`. From now it will not override selectors arbitrarily.

We want it to act as a preview area. As **Namespaces**, **Label Selectors**... change, it will also be updated synchronously. In default, all pods will be selected if you only change the selectors. Uncheck some pods will limit the scope to specified pods.

Also, the **Pods Preview** will limit you to select at least one pod. If not, the experiment scope will be meaningless.

### DIFF
--- a/ui/src/components/FormField/AutocompleteMultipleField.tsx
+++ b/ui/src/components/FormField/AutocompleteMultipleField.tsx
@@ -17,7 +17,7 @@ const AutocompleteMultipleField: React.FC<AutocompleteMultipleFieldProps & TextF
 }) => {
   const { values, setFieldValue } = useFormikContext<Experiment>()
 
-  const firstRenderRef = useRef(false) // This ref prevent to exec the callback function when labels are empty in the first render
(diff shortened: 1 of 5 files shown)


ui/src/components/FormField/AutocompleteMultipleField.tsx
```diff
@@ -17,6 +17,7 @@ const AutocompleteMultipleField: React.FC<AutocompleteMultipleFieldProps & TextF
 }) => {
   const { values, setFieldValue } = useFormikContext<Experiment>()
 
+  const firstRenderRef = useRef(false) // This ref prevent to exec the callback function when labels are empty in the first render
```
fewdan, line 20: s/prevent/prevents/g ?

g1eny0ung: Should be!

PR #759, opened 2020-07-30
Support container-wise StressChaos injection

YangKeao:
Signed-off-by: Yang Keao <keao.yang@yahoo.com>

### What problem does this PR solve?

Fix #755 

Support container wise injection and run stress-ng in container's cgroup in default. 

### Check List

Tests

- Manual test

Code changes

- Has Go code change

PR #794, opened 2020-08-07
[update] add container-wise StressChaos in dashboard

fewdan:

After #759 merged, the container name was added to the StressChaos.
This PR adds container name to the dashboard

 add container-wise StressChaos in dashboard

Tests

- No code

Code changes

- Has Go code change

Side effects

- NO

Related changes

- NO

```release-note
NONE
```

### DIFF
--- a/pkg/apiserver/experiment/experiment.go
+++ b/pkg/apiserver/experiment/experiment.go
@@ -342,6 +342,10 @@ func (s *Service) createStressChaos(exp *core.ExperimentInfo) error {
 		chaos.Spec.Duration = &exp.Scheduler.Duration
 	}
 
+	if exp.Target.StressChaos.ContainerName != nil {
+		chaos.Spec.ContainerName = exp.Target.StressChaos.ContainerName
+	}
+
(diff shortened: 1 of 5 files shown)


pkg/apiserver/experiment/experiment.go
```diff
@@ -342,6 +342,10 @@ func (s *Service) createStressChaos(exp *core.ExperimentInfo) error {
 		chaos.Spec.Duration = &exp.Scheduler.Duration
 	}
 
+	if exp.Target.StressChaos.ContainerName != nil {
```
cwen0, line 345: Do we need to check if `exp.Target.StressChaos` is `nil`?

g1eny0ung:
@cwen0 The dashboard has already followed the changes, PTAL.

pkg/apiserver/experiment/experiment.go
g1eny0ung: Seems all `Create Experiment API`s should need to be checked since all `ChaosInfo`s are pointers. 😢

We can plan a `refactor` for our server APIs.

PR #806, opened 2020-08-12
Fix upload install scripts and crd files workflows

cwen0:

When pushing a tag, we need to upload the install scripts and CRD files regardless of whether the files were changed.

* Use AWS CLI directly 
* Add a single file to release files when pushing a tag


.github/workflows/upload_files.yml
```diff
@@ -33,6 +30,8 @@ jobs:
           if [[ "$GIT_TAG" == "master" ]]; then
             GIT_TAG="latest"
           fi
+          cloud-assets-utils verify-and-sync install.sh -aws true -aws-bucket ${{ secrets.AWS_BUCKET_NAME }} -replace-first-path-to ${GIT_TAG}
```
cwen0, line 33: @g1eny0ung PTAL

PR #810, opened 2020-08-13
refactor: better pr template

g1eny0ung:

I think we can use task lists as our checklist. For the old PR template, it's harder to remove the unrelated checks. By using task lists, you can check your changes and leave others as unchecked.

image

```release-note
NONE
```


.github/pull_request_template.md
```diff
@@ -3,33 +3,32 @@
...
+- [ ] No code
 
 Code changes
 
-- Has Go code change
-- Has CI related scripts change
-- Has Terraform scripts change
+- [ ] Has Go code change
+- [ ] Has CI related scripts change
+- [ ] Has Terraform scripts change
```
yeya24, line 21: I think we can remove all these `Code changes`. They are not really necessary because what are changed is obvious. And we don't really have terraform scripts. What do you think :)?

g1eny0ung: I think so. All these three changes are obvious in the code. What about @zhouqiang-cl @cwen0 ?

PR #826, opened 2020-08-20
New implementation of IoChaos

YangKeao:

This PR submits a new implementation of IoChaos. With this implementation, there is no need to inject a sidecar container, so that it's easier to use.

Note:

* The toda directory is managed with `git submodules`

* Keep in mind that all information (such as attributes, content) could be stored in the page cache.

* The definition of IoChaos has changed, so the dashboard needs to be modified. (I have tried to make this modification minimal)

Known issues:

* If you delete the chaos-daemon while injecting, the target pod may panic.

TODO:

- [x] Resolve the deadlock in toda

- [x] Add e2e test.

- [ ] Add more examples.

- [x] Support mmap

- [x] Change current directory

- [x] Fix Dockerfile for building UI and swagger

- [x] Auto resolve soft link (to volume mount path)

### Checklist

Tests

- [ ] Manual test


pkg/chaosdaemon/iochaos_server.go
```diff
@@ -0,0 +1,122 @@
...
+	}, nil
+}
+
+func (s *daemonServer) killIoChaos(ctx context.Context, pid int64, startTime int64) error {
+	log.Info("killing toda", "pid", pid)
+
+	ins, err := process.NewProcess(int32(pid))
+	if err != nil {
+		log.Info("cannot find process", "pid", pid)
+		return nil
```
yeya24, line 98: Is it intended to return nil here?

YangKeao: If the process has exited unintendedly, (e.g. the pod was restarted), I think we shouldn't block the experiment with a failed recovery, so nil is returned here.

I have refined the subprocess management in #838, which will make sure when the chaos-daemon died, the subprocess will be killed (with a SIGTERM). So that a better way to handle this in Go may be waiting for all subprocesses and recording the unexpected exits, I will try to implement this feature in the future.

BTW, a "standard" way in Linux world could be handling SIGCHLD signal and `waitpid` in the signal handler. However, it's difficult to set up a real signal handler and communicate with the data structure in Go safely. As for the signal handling method in Go, when we receive a SIGCHLD through the channel (which is out of the signal interrupts), the subprocess will have disappeared and the `waitpid` will return an error :( .

YangKeao:
Yeah! Yeah! Yeah! It passes e2e test!!!

api/v1alpha1/iochaos_types.go
```diff
@@ -125,11 +81,28 @@ type IoChaosSpec struct {
...
-	Methods []string `json:"methods,omitempty"`
+	Methods []IoMethod `json:"methods,omitempty"`
+
+	// Percent defines the percentage of injection errors and provides a number from 0-100.
+	// default: 100.
+	// +optional
+	Percent int `json:"percent,omitempty"`
+
+	// VolumePath represents the mount path of injected volume
+	VolumePath string `json:"volumePath"`
```
YangKeao, line 92: `VolumePath` is a newly added field. It's used to specify the path of target volume. @g1eny0ung

api/v1alpha1/iochaos_types.go
```diff
@@ -81,42 +52,27 @@ type IoChaosSpec struct {
...
-	// such as "300ms", "-1.5h" or "2h45m".
-	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
-	// +optional
-	Duration *string `json:"duration,omitempty"`
-
-	// Layer represents the layer of the I/O action.
-	// Supported value: fs.
-	// Default layer: fs
-	// +kubebuilder:validation:Enum=fs
-	Layer IOLayer `json:"layer"`
```
YangKeao, line 97: `Layer` field was deleted. @g1eny0ung

api/v1alpha1/iochaos_types.go
```diff
@@ -81,42 +52,27 @@ type IoChaosSpec struct {
...
+	// Action defines the specific pod chaos action.
+	// Supported action: latency / fault / attrOverride
+	// +kubebuilder:validation:Enum=latency;fault;attrOverride
+	Action IoChaosType `json:"action"`
```
YangKeao, line 58: `Action` is changed from `delay / errno / mixed` to `latency / fault / attrOverride`. @g1eny0ung 

Multiple IoChaos can be applied to the same pod, but they should have the same volume path.

api/v1alpha1/iochaos_types.go
```diff
@@ -125,11 +81,28 @@ type IoChaosSpec struct {
 	// Methods defines the I/O methods for injecting I/O chaos action.
 	// default: all I/O methods.
 	// +optional
-	Methods []string `json:"methods,omitempty"`
+	Methods []IoMethod `json:"methods,omitempty"`
+
+	// Percent defines the percentage of injection errors and provides a number from 0-100.
+	// default: 100.
+	// +optional
+	Percent int `json:"percent,omitempty"`
```
YangKeao, line 89: The type of `Percent` was changed from `string` to `int` @g1eny0ung

api/v1alpha1/iochaos_types.go
```diff
@@ -125,11 +81,28 @@ type IoChaosSpec struct {
...
 
-	// Addr defines the address for sidecar container.
+	// Duration represents the duration of the chaos action.
+	// It is required when the action is `PodFailureAction`.
+	// A duration string is a possibly signed sequence of
+	// decimal numbers, each with optional fraction and a unit suffix,
+	// such as "300ms", "-1.5h" or "2h45m".
+	// Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
 	// +optional
-	Addr string `json:"addr,omitempty"`
```
YangKeao, line 132: This field was removed.

api/v1alpha1/iochaos_types.go
```diff
@@ -81,42 +52,27 @@ type IoChaosSpec struct {
...
 	// +optional
 	Delay string `json:"delay,omitempty"`
 
 	// Errno defines the error code that returned by I/O action.
 	// refer to: https://www-numi.fnal.gov/offline_software/srt_public_context/WebDocs/Errors/unix_system_errors.html
-	//
-	// If `Errno` is empty, the operator will generate an error code for it randomly.
 	// +optional
-	Errno string `json:"errno,omitempty"`
+	Errno uint32 `json:"errno,omitempty"`
```
YangKeao, line 71: Type of `Errno` was changed from `string` to `uint32` @g1eny0ung

YangKeao (commented):
I don't know whether I have broken the backend of the dashboard. If you face any problem with fitting the frontend, please tell me and I will try my best to help you. @g1eny0ung

g1eny0ung:
@YangKeao The dashboard is easy to sync these changes. I will address these comments in these two days.

api/v1alpha1/iochaos_types.go
```diff
@@ -81,42 +52,27 @@ type IoChaosSpec struct {
...
 
-	// Percent defines the percentage of injection errors and provides a number from 0-100.
-	// default: 100.
+	// Attr defines the overrided attribution
 	// +optional
-	Percent string `json:"percent,omitempty"`
+	Attr *AttrOverrideSpec `json:"attr,omitempty"`
```
g1eny0ung, line 75: @YangKeao Seems there is also a new field, need docs to describe it.

YangKeao: I have described it in [document](https://github.com/chaos-mesh/chaos-mesh/blob/f117548019f6babf1be03bd185c27ba12704e537/website/docs/user_guides/io_chaos.md). It's a [struct](https://github.com/chaos-mesh/chaos-mesh/blob/f117548019f6babf1be03bd185c27ba12704e537/api/v1alpha1/podiochaos_types.go#L96) in go and a map in yaml (with limited keys)

Dockerfile
```diff
@@ -1,30 +1,65 @@
...
+    --mount=type=cache,target=/root/.cache/go-build \
+    --mount=type=cache,target=/src/ui/node_modules \
     IMG_LDFLAGS=$LDFLAGS make binary
+
+FROM pingcap/chaos-build-base AS rust_build
+
+ARG HTTPS_PROXY
+ARG HTTP_PROXY
+
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain nightly-2020-07-01 -y
```
cwen0, line 32: We don't need to build `toda` every time.  Just use `toda` as a single project and use the release image at this repo.

api/v1alpha1/iochaos_types.go
g1eny0ung: Got it.

Dockerfile
YangKeao: I agree with you :+1:

pkg/bpm/bpm.go
```diff
@@ -0,0 +1,355 @@
+// Copyright 2019 Chaos Mesh Authors.
```
cwen0, line 1: s/2019/2020

api/v1alpha1/podiochaos_webhook.go
```diff
@@ -0,0 +1,89 @@
...
+
+// RegisterPodIoHandler registers handler into webhook
+func RegisterPodIoHandler(newHandler PodIoChaosHandler) {
+	podIoChaosHandler = newHandler
+}
+
+// SetupWebhookWithManager setup PodIoChaos's webhook with manager
+func (in *PodIoChaos) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	mgr.GetWebhookServer().
+		Register("/mutate-chaos-mesh-org-v1alpha1-podiochaos", &webhook.Admission{Handler: &PodIoChaosWebhookRunner{}})
```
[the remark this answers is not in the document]
YangKeao: Because we need to use a lower level interface of webhook.

website/docs/user_guides/io_chaos.md
```diff
@@ -4,86 +4,9 @@ title: IOChaos Experiment
 sidebar_label: IOChaos Experiment
 ---
 
-This document helps you build IOChaos experiments.
+This document helps you run IOChaos experiments.
```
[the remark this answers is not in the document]
YangKeao: It has been moved to #903

api/v1alpha1/iochaos_webhook.go
```diff
@@ -109,30 +107,10 @@ func (in *IoChaosSpec) validateDelay(delay *field.Path) field.ErrorList {
...
-			if err != nil {
-				allErrs = append(allErrs, field.Invalid(errno, in.Errno,
-					fmt.Sprintf("parse errno field error:%s for action:%s", err, in.Action)))
-			}
-		}
-	}
-	return allErrs
-}
-
-func (in *IoChaosSpec) validatePercent(percentField *field.Path) field.ErrorList {
```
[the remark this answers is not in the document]
YangKeao: Oops, it's a mistake.

api/v1alpha1/podiochaos_types.go
```diff
@@ -0,0 +1,200 @@
...
+	Methods []IoMethod `json:"methods,omitempty"`
+
+	// Percent represents the percent probability of injecting this action
+	Percent int `json:"percent"`
+}
+
+// IoFault represents the fault to inject and their weight
+type IoFault struct {
+	Errno  uint32 `json:"errno"`
+	Weight int32  `json:"weight"`
```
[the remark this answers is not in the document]
YangKeao: `toda` provides the ability to set multiple `IoFault` and randomly choose one according to the weight. But to keep the consistency of IoChaos API, I didn't provide this ability in IoChaos. Instead, we could implement this feature in `PodIoChaos` layer, to support multiple io faults on the same pod in the future.

cwen0 (dismissed):
LGTM

cwen0 (approved):

fewdan (approved):
LGTM

ti-srebot (approved):

Issue #834, opened 2020-08-21
rpc error: code = Unknown desc = process does not have children


asos-craigmorten:
## Bug Report

**What version of Kubernetes are you using?**

```
$ kubectl version

Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.8", GitCommit:"9f2892aab98fe339f3bd70e3c470144299398ace", GitTreeState:"clean", BuildDate:"2020-08-14T11:06:25Z", GoVersion:"go1.14.7", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.3", GitCommit:"2e7996e3e2712684bc73f0dec0200d64eec7fe40", GitTreeState:"clean", BuildDate:"2020-05-20T12:43:34Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

**What did you do?**

Set up a local Kubernetes cluster using minikube:

```bash
minikube start --driver=virtualbox
minikube addons enable metrics-server
```

Ran a simple target application:

```bash
kubectl create ns echo

kubectl create deployment echo \
  --image=k8s.gcr.io/echoserver:1.4 \
  -n echo

kubectl expose deployment echo \
  -n echo \
  --type=LoadBalancer \
  --port=8080

minikube service echo -n echo
```

Cloned chaos-mesh (latest master) and installed using Helm as per chaos-mesh docs (with dashboard enabled).

```
git clone https://github.com/chaos-mesh/chaos-mesh

kubectl apply -f ./chaos-mesh/manifests/crd.yaml

kubectl create ns chaos-mesh

helm upgrade --install chaos-mesh ./chaos-mesh/helm/chaos-mesh \
  -n chaos-mesh \
  --set dashboard.create=true

minikube service chaos-dashboard -n chaos-mesh
```

Attempted to run a simple stress test impacting the echoserver's CPU to see how chaos-mesh works. See the following `stress-cpu.yaml`:

```bash
apiVersion: chaos-mesh.org/v1alpha1
kind: StressChaos
metadata:
  name: stress-cpu
  namespace: chaos-mesh
spec:
  mode: all
  selector:
    labelSelectors:
      app: echo
    namespaces:
    - echo
  stressors:
    cpu: 
      workers: 1
      load: 100
  value: ""
  duration: "1m"
  scheduler:
    cron: "@every 2m"
```

Run the test using:

[...]

PR #838, opened 2020-08-24
Refine the subprocess management

YangKeao:

Maybe will solve #834 . And it's a preparation for new iochaos (which also needs subprocess management)

By setting `prctl(PR_SET_PDEATHSIG, SIGTERM)` for `nsenter` and the subprocess of `nsenter`. As the `nsenter` will fork when we enter a new pid namespace, and nothing is done for killing subprocess or suiciding when the parent dies.

For example: 

If you call `nsenter -p/proc/.../ns/pid bash -c "while true; do sleep 1; date; done"` then even you kill the nsenter process, the subprocess of it will continue running until it gets killed. The suicide program is used to make sure that the subprocess will be terminated when its parent died.

I'm not sure this method is 100% reliable, but half a loaf is better than none.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

1. Stop the chaos-daemon will stop all stress-ng process.

2. Apply and recover the StressChaos works normally.


api/v1alpha1/stresschaos_types.go
```diff
@@ -119,9 +119,6 @@ type StressInstance struct {
 	// UID is the instance identifier
 	// +optional
 	UID string `json:"uid"`
-	// StartTime specifies when the instance starts
-	// +optional
-	StartTime *metav1.Time `json:"startTime"`
```
cwen0, line 124: Why delete `StartTime`? This `StartTime` was used to avoid the PID was reused.

YangKeao: Oops. It's really a mistake. Checking whether the target is the child process of chaos-daemon cannot replace this start-time check.

I will fix it right now.

pkg/chaosdaemon/util.go
```diff
@@ -212,59 +213,111 @@ func GetNsPath(pid uint32, typ nsType) string {
...
-		f := c.(func(context.Context, []nsOption, string, ...string) *exec.Cmd)
-		return f(ctx, options, cmd, args...)
+func (b *processBuilder) EnableSuicide() *processBuilder {
+	b.suicide = true
+
+	return b
+}
+
+func (b *processBuilder) Build(ctx context.Context) *exec.Cmd {
+	// The call routine is suicide -> pause -> nsenter --(fork)-> suicide -> process
```
cwen0, line 270: s/suicide -> pause /pause->suicide/g

YangKeao: Fixed

cwen0 (approved):
LGTM

fewdan (approved):
LGTM

PR #851, opened 2020-08-26
Always recreate controller-manager while upgrading helm charts

YangKeao:
Signed-off-by: Yang Keao <keao.yang@yahoo.com>

### What problem does this PR solve?

Fix #850 

### What is changed and how does it work?

Add a random annotation to controller-manager, so that the deployment YAML changes every time.

PR #853, opened 2020-08-26
Fix stress chaos map

YangKeao:

A possible solution to #847 

Fix two bugs:

1. Instances map is not protected with a lock :angry: 

2. Check instances before applying chaos. This will fix extra `stress-ng` processes due to fail and retry.

The logic about retrying in chaos-mesh is really chaos :crying_cat_face: . We should improve it in near-future #851 .

And, sorry for not finding these bugs when I was reviewing. I'm still not adapted with dealing these problems without the help of compiler or other tools...

### Checklist

Tests

- [x] Manual test


fewdan (approved):
LGTM

cwen0 (approved):
LGTM

ti-srebot (approved):

PR #865, opened 2020-09-02
Fix the error in building tc commands

YangKeao:

```
2020-09-02T03:47:24.659Z        INFO    chaos-daemon-server     build command   {"command": "nsenter -n/proc/20669/ns/net -- tc qdisc add dev eth0 parent 1: handle 2: netem  duplicate 80.000000 25.000000"}
2020-09-02T03:47:24.685Z        ERROR   chaos-daemon-server     error while adding tc   {"error": "error code: exit status 1, msg: Illegal \"limit\"\n"}
```

There is an extra space between "netem" and "duplicate" (which will lead to an "" in args).

BTW: What is the best practice of concating command line arguments? Directly working on a `[]string` seems too complicated (as I will need to convert integer/float to string everywhere).

### Checklist

Tests

- [x] Unit test


cwen0 (approved):
LGTM

WangXiangUSTC (approved):
LGTM

Issue #487, opened 2020-05-11
Support namespace scoped rbac chaos-mesh


Yisaer:
## Description

Currently Chaos-mesh required ClusterScope rbac to run in Kubernetes. However, in some strict environments, we need chaos-mesh run as a namespace-scoped rbac. We want to support namespace scoped rbac chaos-mesh 

Close #477 

## Score

* 300

## Mentor(s)

* [@yisaer](https://github.com/Yisaer)

Contact the mentors: **#tidb-challenge-program** channel in [TiDB-Operator Community](https://join.slack.com/t/tidbcommunity/shared_invite/enQtNzc0MzI4ODExMDc4LWYwYmIzMjZkYzJiNDUxMmZlN2FiMGJkZjAyMzQ5NGU0NGY0NzI3NTYwMjAyNGQ1N2I2ZjAxNzc1OGUwYWM0NzE)  Slack Workspace
As the logic is not simple and the detail is still needed to be discussed, we are welcome for every challenger to contact the mentor if you have any questions.

## Recommended Skills
* Golang
* Kubernetes

yeya24:
@Yisaer Is this issue conflict with https://github.com/pingcap/chaos-mesh/issues/355?

Yisaer:
> @Yisaer Is this issue conflict with #355?

I don't think so. If the chaos-mesh only have the namespaced scope rbac, it should watch the configmaps in the certain namespace with certain labels.

yeya24:
> > @Yisaer Is this issue conflict with #355?
> 
> I don't think so. If the chaos-mesh only have the namespaced scope rbac, it should watch the configmaps in the certain namespace with certain labels while it do need some work to do.

Thanks for the clarification. Make sense. So we need something like a `--mode` flag to specify whether chaos mesh is clusterscoped?

Yisaer:
@yeya24  Exactly.

STRRL:
I am trying to work on this issue.

Here is my idea:
- Make new fields `ClusterScoped` and `Namespace` in struct `ChaosControllerConfig`; `ClusterScoped`(bool) represent that running in cluster mode or not. `Namespace` represent the namespace could injected chaos into.
- Make same above configurations in helm's yaml file
- New validation about configurations: `ClusterScoped`, `Namespace`, `AllowedNamespaces`, `IgnoredNamespaces`   
- While  `ClusterScoped` is `false`, and `Namespace` is specified, running controller manager with certain NS.

[...]

PR #613, opened 2020-06-10
Unified configuration for chaos controller

Gallardot:
### What problem does this PR solve? 
#598 legacy, Unified configuration for chaos controller
### Check List 

Tests 

 - Unit test
 - E2E test
 - Manual test (add detailed scripts or steps below)
 - No code

Code changes

 - Has Go code change
 - Has CI related scripts change
 - Has Terraform scripts change

Side effects

 - Breaking backward compatibility

Related changes

 - Need to update the documentation

### Does this PR introduce a user-facing change?:

```release-note
NONE
```

PR #872, opened 2020-09-06
[add] Support namespace scoped chaos

STRRL:
Make chaos-mesh controller could work without permission of cluster scope, like pod list from cluster scope. 

Related issue: https://github.com/chaos-mesh/chaos-mesh/issues/487

This PR makes three things:
- Add new configurations to switch on namespace scope mode, and target namespace for chaos injecting within  namespace scope mode.
- Make controller manager only watch chaos objects belongs to target namespace; Make pod selector only list pods belongs to target namespace;
- Create configuration validation and other logic to reject out-of-scoped behaviors.

### Checklist

Todos:
- [x] kubebuilder controller manager support namespace scoped mode
- [x] K8sConfigMapWatcher support namespace scoped mode
- [x] helm configs
~- [ ]  enhance webhook (like, using namespaceSelector)~
- [x] controller manager could skip chaos that does not belong to target namespace.
~~- [ ] e2e tests~~ It's hard to test as expected operation is NOOP.
- [x] new rbac yaml, as a mininum permission list, for controller serviceaccount 
- [ ] found the mininum permission list to install chaos-mesh
- [ ] update the document

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Manually test steps:

WIP

Side effects

Related changes

No

```release-note
NONE
```


cwen0:
@Yisaer PTAL

cmd/controller-manager/main.go
```diff
@@ -79,12 +81,29 @@ func main() {
...
+		setupLog.Info("Chaos controller manager is running in namespace scoped mode.", "namespace", common.ControllerCfg.Namespace)
+		mgr, err = ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+			Scheme: scheme,
+			// one more thing is the specific namespace
+			Namespace:          common.ControllerCfg.Namespace,
+			MetricsBindAddress: common.ControllerCfg.MetricsAddr,
+			LeaderElection:     common.ControllerCfg.EnableLeaderElection,
+			Port:               9443,
+		})
+	}
```
Gallardot, line 105:
```
var ns string = ""
if common.ControllerCfg.ClusterScoped {
		setupLog.Info("Chaos controller manager is running in cluster scoped mode.")
	} else {
		setupLog.Info("Chaos controller manager is running in namespace scoped mode.", "namespace", common.ControllerCfg.Namespace)
                 ns = common.ControllerCfg.Namespace
}

mgr, err = ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
			Scheme: scheme,
			Namespace:          ns,
			MetricsBindAddress: common.ControllerCfg.MetricsAddr,
			LeaderElection:     common.ControllerCfg.EnableLeaderElection,
			Port:               9443,
		})
```

I think it's better that way.  Default watches all namespaces.

Gallardot (changes requested):

cmd/controller-manager/main.go
STRRL: I agree with U :)

STRRL:
I found that `K8sConfigMapWatcher` also need some patches, I will finish these changes very soon.

STRRL:
Hi @cwen0  @Gallardot  mentors, one more thing, shall we consider to split out namespace which inject chaos and namespace which running controller?

For example, running `chaos-controller-manager` in ns `chaos-control-plane`, and put `chaos-daemon` DaemonSet  into another ns `chaos-testing`.

cmd/controller-manager/main.go
Yisaer, line 105: I tend to init manager in this way:
```golang
if common.ControllerCfg.ClusterScoped{
 // init option
}else{
 // init option
}
// init manager.

```

cmd/controller-manager/main.go
```diff
@@ -20,6 +20,8 @@ import (
 	"os"
 	"time"
 
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+
```
Yisaer, line 24: fmt goimports.

pkg/utils/selector.go
```diff
@@ -79,11 +79,18 @@ func SelectAndFilterPods(ctx context.Context, c client.Client, spec SelectSpec)
 // It returns all pods that match the configured label, annotation and namespace selectors.
 // If pods are specifically specified by `selector.Pods`, it just returns the selector.Pods.
 func SelectPods(ctx context.Context, c client.Client, selector v1alpha1.SelectorSpec) ([]v1.Pod, error) {
+	// TODO: refactor: make different selectors to replace if-else logics
```
Yisaer, line 82: agreed.

Yisaer:
Basically LGTM except the selector is too complex for now. Maybe we need to find a better way to refactor it. Adding an e2e test to verify the namespace-scoped would be better. @STRRL

STRRL:
Hi @Yisaer , could you take a little while on this  https://github.com/chaos-mesh/chaos-mesh/pull/872#issuecomment-688285264 ? Please give me your opinion about this question. 

Thanks!

STRRL:
> Hi @cwen0 @Gallardot mentors, one more thing, shall we consider to split out namespace which inject chaos and namespace which running controller?
> 
> For example, running `chaos-controller-manager` in ns `chaos-control-plane`, and create a `IOChaos` in another ns `chaos-testing`.
> 
> Or we should play in ONLY ONE namespace.

I am working on this, it could improve isolation of our system.

STRRL:
> Basically LGTM except the selector is too complex for now. Maybe we need to find a better way to refactor it.

Hi @Yisaer . I create new issue to track this. https://github.com/chaos-mesh/chaos-mesh/issues/886

I do not recommend refactoring while appending new features. :D

Yisaer:
> > Basically LGTM except the selector is too complex for now. Maybe we need to find a better way to refactor it.
> 
> Hi @Yisaer . I create new issue to track this. #886
> 
> I do not recommend refactoring while appending new features. :D

Thanks for your contribution. I didn't mean refactoring selector in this request either. This should be tracked by a new issue and still need to be discussed.

helm/chaos-mesh/templates/controller-manager-deployment.yaml
```diff
@@ -47,6 +47,14 @@ spec:
...
               fieldRef:
                 fieldPath: metadata.namespace
+          - name: TEMPLATE_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+          - name: TARGET_NAMESPACE
+            value: {{ .Values.controllerManager.targetNamespace }}
+          - name: CLUSTER_SCOPED
+            value: "{{ .Values.clusterScoped }}"
```
Yisaer, line 57: Most controllers I knew is using `starting flag` to indicate the `scope`, Could you keep the same style?

helm/chaos-mesh/values.yaml
```diff
@@ -38,6 +38,10 @@ controllerManager:
   allowedNamespaces: ""
   ignoredNamespaces: ""
 
+  # targetNamespace only works with clusterScoped is false(namespace scoped mode).
+  # It means namespace which will be injected chaos
+  targetNamespace: chaos-testing
```
Yisaer, line 43: I think we don't need to specify the `targetNamespace`. The effected scoped obviously located at where the chaos-mesh deployed in its namespace.

Yisaer (commented):
I suggest using [namespaceSelector](https://kubernetes.io/zh/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-namespaceselector) for webhook while `namespace-scoped` is enabled.

As `namespaceSelector` need apiserver >= 1.15(need to be verified), the `if-else` template rendering could refer to  [here](https://github.com/pingcap/tidb-operator/blob/4e2bbe8b9a8bbde989562144ff1c3f93889557d7/charts/tidb-operator/templates/admission/admission-webhook-registration.yaml#L39-L43)

Yisaer:
> ning `chaos-controller-manager` in ns `chaos-control-plane`, and create a `IOChaos` in another ns `chaos-testing`.
> 
> Or we should play in ONLY ONE name

For this question, my answer is no (do not split the manager and chaos in different namespaces). The origin demand for this feature is to limit the range for the chaos mesh (namespace scoped is common for the most controllers while it is more important for chaos-mesh comparing to other controllers as we do need to control the effected range by chaos-mesh)

STRRL:
Hi @Yisaer mentor, I still do not figure out why. 😅

Deploying chaos-controller-manager and applications which are the target of chaos-testing into SAME namespace is DANGEROUS.
 
For example, if user using a bad selector, it might inject chaos into controller itself.

Other operators like prometheus-operator and openkruise have their own namespace which only running controller. I am very sure that your will NOT put your application into `kube-prometheus` or`kube-system`  (doge.

> And this question also remind me that the manager's `servceiaccount` should be bound by `role` instead of `clusterRole`. The role is the only way to guarantee that the manager won't affect other namespaces.

`role` and `clusterRole` is not the key, binding is the key. We could create a `RoleBinding` related to a `ClusterRole`. It is the way to grant a serviceaccount cross-namespace permissions.

It seems that we have opposite opinions, I am very pleasure to discuss this question on wechat or slack. :D

helm/chaos-mesh/templates/controller-manager-deployment.yaml
STRRL: Do you mean a command line flags like `--scoped=xxx`?

Or could you give me an example?

Yisaer:
`RoleBinding` related to a `ClusterRole` is great! Thanks for pointing it out! @STRRL 

In this way, deploying `chaos-mesh` into its own namespace and affect other limit namespaces is ok to me.

helm/chaos-mesh/templates/controller-manager-deployment.yaml
Yisaer: Exactly, using command line flags is what I mean.

STRRL: Currently, all configuration things in chaos-controller is stored by environment variables. I just want to keep same style.

I agree with your suggestion, cli flags is more friendly to help information and auto-completion. But we still have a lot work to do for transferring from ENV to FLAG.

How about creating another issue to track this?

Gallardot: > I agree with your suggestion, cli flags is more friendly to help information and auto-completion. But we still have a lot work to do for transferring from ENV to FLAG.
> 
> How about creating another issue to track this?

https://github.com/chaos-mesh/chaos-mesh/pull/613 https://github.com/chaos-mesh/chaos-mesh/pull/598
@STRRL 
Please don't do it unless it is absolutely necessary.
We've already talked about this in the WeChat group, and we've decided to use env, and keep the same style.

STRRL: @Gallardot @Yisaer 
Aha, actually I don't care using ENV or FLAG. I think the best practice is a configuration layer, stay on the top of the dependency tree. 

Yes, it is `config.ChaosControllerConfig`(and related things). So whatever what we want, we could just modify this pkg. (Maybe we just using viper someday, to build a ~~complex~~  efficient configuration layer.

BTW, a nested `WatcherConfig *watcher.Config` is a really bad idea. It is the inner state of `K8sConfigMapWather`, we should not put it into config struct rudely.

STRRL: So, keep the ENV. 😀

STRRL:
> I suggest using [namespaceSelector](https://kubernetes.io/zh/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-namespaceselector) for webhook while `namespace-scoped` is enabled.

Hi @Yisaer mentor, tldr I think we don't need  namespaceSelector.

NamespaceSelector could control a webhook will apply on a certain object or not. I think if we put a Chaos Object out of the target namespace, the webhook should see this action and reject it, let the incorrect action fail-fast. So our webhook should apply on each namespace.

Yisaer:
> I think if we put a Chaos Object out of the target namespace, the webhook should see this action and reject it, let the incorrect action fail-fast. So our webhook should apply on each namespace.

I think using `RuleBinding` with the `ClusterRole` is conflict with this. @STRRL

STRRL:
> I think using `RuleBinding` with the `ClusterRole` is conflict with this. @STRRL

Well, after my testing, I found that webhook still works well to validating a Chaos Object which belongs to a namespace that controller-manager didn't controlled.

I think the webhooks is a mechanism for apiserver creating something, api-server push Object content to webhook positively(so controller-manager do NOT operate actions like get, list).

I need more time to figure out the detail of webhooks. For now, it just works, so I think fail-fast for rejecting invalid out-of-namespaced ChaosObject is reasonablenessary.

STRRL:
Hi @Yisaer @Gallardot . PTAL. :)

- Basically it does not change core logic. 
- Appending a NOOP for controller while receiving out-of-namespace chaos.
- Changing helm rbac files. Only left a minimum permission for controller-manager serviceaccount.

controllers/httpchaos_controller.go
```diff
@@ -42,6 +44,13 @@ type HTTPChaosReconciler struct {
 func (r *HTTPChaosReconciler) Reconcile(req ctrl.Request) (result ctrl.Result, err error) {
 	logger := r.Log.WithValues("reconciler", "httpfaultchaos")
 
+	if !common.ControllerCfg.ClusterScoped && req.Namespace != common.ControllerCfg.TargetNamespace {
+		// NOOP
+		logger.Info("ignore chaos which belongs to an unexpected namespace within namespace scoped mode",
+			"chaosName", req.Name, "expectedNamespace", common.ControllerCfg.TargetNamespace, "actualNamespace", req.Namespace)
+		return
```
Gallardot, line 51: Maybe a little bit clearer
```suggestion
		return ctrl.Result{}, nil
```

controllers/iochaos_controller.go
```diff
@@ -42,6 +44,14 @@ type IoChaosReconciler struct {
 
 // Reconcile reconciles an IOChaos resource
 func (r *IoChaosReconciler) Reconcile(req ctrl.Request) (result ctrl.Result, err error) {
+	logger := r.Log.WithValues("reconciler", "iochaos")
+
+	if !common.ControllerCfg.ClusterScoped && req.Namespace != common.ControllerCfg.TargetNamespace {
+		// NOOP
+		logger.Info("ignore chaos which belongs to an unexpected namespace within namespace scoped mode",
+			"chaosName", req.Name, "expectedNamespace", common.ControllerCfg.TargetNamespace, "actualNamespace", req.Namespace)
+		return
```
Gallardot, line 53: ditto

controllers/networkchaos_controller.go
```diff
@@ -43,6 +45,13 @@ type NetworkChaosReconciler struct {
 func (r *NetworkChaosReconciler) Reconcile(req ctrl.Request) (result ctrl.Result, err error) {
 	logger := r.Log.WithValues("reconciler", "networkchaos")
 
+	if !common.ControllerCfg.ClusterScoped && req.Namespace != common.ControllerCfg.TargetNamespace {
+		// NOOP
+		logger.Info("ignore chaos which belongs to an unexpected namespace within namespace scoped mode",
+			"chaosName", req.Name, "expectedNamespace", common.ControllerCfg.TargetNamespace, "actualNamespace", req.Namespace)
+		return
```
Gallardot, line 52: ditto

controllers/podchaos_controller.go
```diff
@@ -43,6 +45,13 @@ type PodChaosReconciler struct {
 func (r *PodChaosReconciler) Reconcile(req ctrl.Request) (result ctrl.Result, err error) {
 	logger := r.Log.WithValues("reconciler", "podchaos")
 
+	if !common.ControllerCfg.ClusterScoped && req.Namespace != common.ControllerCfg.TargetNamespace {
+		// NOOP
+		logger.Info("ignore chaos which belongs to an unexpected namespace within namespace scoped mode",
+			"chaosName", req.Name, "expectedNamespace", common.ControllerCfg.TargetNamespace, "actualNamespace", req.Namespace)
+		return
```
Gallardot, line 52: ditto

controllers/stresschaos_controller.go
```diff
@@ -44,6 +46,13 @@ type StressChaosReconciler struct {
 func (r *StressChaosReconciler) Reconcile(req ctrl.Request) (result ctrl.Result, err error) {
 	logger := r.Log.WithValues("reconciler", "stresschaos")
 
+	if !common.ControllerCfg.ClusterScoped && req.Namespace != common.ControllerCfg.TargetNamespace {
+		// NOOP
+		logger.Info("ignore chaos which belongs to an unexpected namespace within namespace scoped mode",
+			"chaosName", req.Name, "expectedNamespace", common.ControllerCfg.TargetNamespace, "actualNamespace", req.Namespace)
+		return
```
Gallardot, line 53: ditto

controllers/timechaos_controller.go
```diff
@@ -43,6 +45,13 @@ type TimeChaosReconciler struct {
 func (r *TimeChaosReconciler) Reconcile(req ctrl.Request) (result ctrl.Result, err error) {
 	logger := r.Log.WithValues("reconciler", "timechaos")
 
+	if !common.ControllerCfg.ClusterScoped && req.Namespace != common.ControllerCfg.TargetNamespace {
+		// NOOP
+		logger.Info("ignore chaos which belongs to an unexpected namespace within namespace scoped mode",
+			"chaosName", req.Name, "expectedNamespace", common.ControllerCfg.TargetNamespace, "actualNamespace", req.Namespace)
+		return
```
Gallardot, line 52: ditto

helm/chaos-mesh/templates/controller-manager-rbac.yaml
```diff
@@ -10,167 +10,161 @@ metadata:
...
+  name: {{ .Release.Name }}:chaos-controller-manager-control-plane
+subjects:
+  - kind: ServiceAccount
+    name: {{ .Values.controllerManager.serviceAccount }}
+    namespace: {{ .Release.Namespace }}
+
+{{- if .Values.clusterScoped }}
+---
+# binding for target namespace,working with cluster scoped mode.
+kind: ClusterRoleBinding
```
Gallardot, line 126: ClusterRoleBinding for clusterScoped

helm/chaos-mesh/templates/controller-manager-rbac.yaml
```diff
@@ -10,167 +10,161 @@ metadata:
...
+subjects:
+  - kind: ServiceAccount
+    name: {{ .Values.controllerManager.serviceAccount }}
+    namespace: {{ .Release.Namespace }}
+{{- end }}
+
+{{- if not .Values.clusterScoped }}
+---
+# binding for target namespace, working with namespace scoped mode.
+kind: RoleBinding
```
Gallardot, line 150: RoleBinding for if `clusterScoped` == false
Why not just like this?
```
{{- if .Values.clusterScoped }}
kind: ClusterRoleBinding
{{- else }}
kind: RoleBinding
{{- end }}
.....
```

Gallardot (changes requested):
Rest LGTM

controllers/httpchaos_controller.go
STRRL: Agree :D

Gallardot (approved):
LGTM

Yisaer (commented):
Supporting multiple namespaces under namespace-scoped could be supported later. The rest LGTM

controllers/common/types_test.go
```diff
@@ -0,0 +1,78 @@
...
+		WatcherConfig: &watcher.Config{
+			ClusterScoped:   false,
+			TargetNamespace: "ns1",
+		},
+		ClusterScoped:   false,
+		TargetNamespace: "ns2",
+	}
+	err := validate(&c)
+	assert.NotNil(t, err)
+}
```
Yisaer, line 78: I think these 4 test cases could be merged into one test suite and managed by table-driven style.

Yisaer (commented):
The testcases could be table-driven style.

helm/chaos-mesh/templates/controller-manager-deployment.yaml
```diff
@@ -47,6 +47,14 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
+          - name: TEMPLATE_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+          - name: TARGET_NAMESPACE
+            value: {{ .Values.controllerManager.targetNamespace }}
```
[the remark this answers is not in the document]
STRRL: Yes! I will create another issue to track this.

controllers/common/types_test.go
STRRL: OK. I will migrate these testcases with ginkgo. :)

helm/chaos-mesh/templates/controller-manager-deployment.yaml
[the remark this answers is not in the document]
STRRL: I created issue: https://github.com/chaos-mesh/chaos-mesh/issues/966

STRRL:
HI, @Yisaer  , PTAL

cwen0:
@Gallardot @Yisaer PTAL

controllers/common/types_test.go
Gallardot: Well,I think what @Yisaer  meant was look  like this https://github.com/chaos-mesh/chaos-mesh/blob/5d409d77ae86c3896b15e2809c6b977999368479/api/v1alpha1/iochaos_webhook_test.go#L35-L40

STRRL: Oops, I have a misunderstanding on table-driven style. 😅

I will refactor these test cases right now.

STRRL: @Gallardot PTAL. :)

Gallardot (approved):

cwen0 (approved):
LGTM

ti-srebot (approved):

PR #874, opened 2020-09-07
feat: add hostNetwork option for controller-manager and update doc accordingly

Colstuwjx:

In our helm chart, the controller-manager didn't support `hostNetwork=true`, and for me, it's quite need, so I proposed this PR.

Updated the `controller-manager-deployment.yaml` and add hostNetwork option support, by default, it's false, so there is nothing changed if we're keep using the original values.

```release-note
NONE
```


cwen0 (approved):
Thanks for your contribution! LGTM

zhouqiang-cl (approved):
LGTM

ti-srebot (approved):

Issue #878, opened 2020-09-08
Specifying fieldSelectors in PodChaos causes an error


masayag:
## Bug Report

**What version of Kubernetes are you using?**
Client Version: 4.4.11
Server Version: 4.6.0-0.nightly-2020-08-31-220837
Kubernetes Version: v1.19.0-rc.2+f71a7ab-dirty

**What did you do?**
I used the 'fieldSelectors' in selector element of the PodChaos resource in order to select a pod.

The following request was applied:
```
apiVersion: chaos-mesh.org/v1alpha1
kind: PodChaos
metadata:
  name: control-plance-pod-failure
  namespace: chaos-testing
spec:
  action: pod-failure
  mode: one
  value: ""
  duration: "30s"
  selector:
    namespaces:
      - "openshift-kube-apiserver"
    labelSelectors:
      "app": "openshift-kube-apiserver"
      "apiserver": "true"
    fieldSelectors:
      "metadata.name": "kube-apiserver-test-infra-cluster-master-1"
  scheduler:
    cron: "@every 2m"
```
When using kubectl with 'fieldSelectors', the response is valid:
```
oc get pods -n openshift-kube-apiserver --field-selector  metadata.name=kube-apiserver-test-infra-cluster-master-1
NAME                                         READY   STATUS    RESTARTS   AGE
kube-apiserver-test-infra-cluster-master-1   5/5     Running   0          11m
```

**What did you expect to see?**
I'd expect a pod named 'kube-apiserver-test-infra-cluster-master-1' to be the target of the failure.

**What did you see instead?**
The podchaos resource was created, but produced an error:
```
oc describe podchaos -n chaos-testing control-plance-pod-failure
...
Spec:
  Action:          pod-failure
  Container Name:  
  Duration:        30s
  Grace Period:    0
  Mode:            one
  Scheduler:
    Cron:  @every 2m
  Selector:
    Field Selectors:
      metadata.name:  kube-apiserver-test-infra-cluster-master-1
    Label Selectors:
      Apiserver:  true
      App:        openshift-kube-apiserver
    Namespaces:
      openshift-kube-apiserver
  Value:  
Status:
  Experiment:
    Phase:  Failed
  Phase:    
  Scheduler:
Events:
  Type     Reason             Age                From                 Message
  ----     ------             ----               ----                 -------
  Warning  ChaosInjectFailed  1s (x14 over 42s)  podchaos-controller  Index with name field:metadata.name does not exist
```
[...]

PR #892, opened 2020-09-09
[update] add client.Reader in Reconciler

fewdan:

solve #878

add client.Reader in Reconciler

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [x] No code

Side effects

Related changes

```release-note
NONE
```


pkg/utils/selector.go
```diff
@@ -120,8 +120,14 @@ func SelectPods(ctx context.Context, c client.Client, selector v1alpha1.Selector
 	if len(selector.FieldSelectors) > 0 {
 		listOptions.FieldSelector = fields.SelectorFromSet(selector.FieldSelectors)
```
yeya24, line 121: Can you please put this line to the line 123? It is not necessary to have two same `if` statement

pkg/apiserver/common/common.go
```diff
@@ -44,16 +44,19 @@ type Pod struct {
 type Service struct {
 	conf    *config.ChaosDashboardConfig
 	kubeCli client.Client
+	Reader  client.Reader
```
yeya24, line 47: Is it necessary to make this field public? Why not just consistent as other fields

yeya24 (dismissed):
Overall this looks perfect! Thanks for the quick fix.
Another thing is that can you please add some E2E tests to ensure this working as expected with fieldSelector?

fewdan:
@yeya24 I have dealt with the other comments, but the e2e test seems inconvenient to add. I tested it manually before.

pkg/utils/selector.go
```diff
@@ -119,9 +119,13 @@ func SelectPods(ctx context.Context, c client.Client, selector v1alpha1.Selector
...
-	}
-	if err := c.List(ctx, &podList, &listOptions); err != nil {
-		return nil, err
+		if err := r.List(ctx, &podList, &listOptions); err != nil {
+			return nil, err
+		}
+	} else {
+		if err := c.List(ctx, &podList, &listOptions); err != nil {
+			return nil, err
+		}
```
WangXiangUSTC, line 128: how about adding a comment about why in this case use client.Reader and in another case use client.Client?

WangXiangUSTC (commented):
rest LGTM

YangKeao (approved):
LGTM

ti-srebot (approved):

YangKeao:
Passing an argument to every function is really annoying (:crying_cat_face: I need `implicit`). Maybe we should think about the code structure and the way to refine it. But for this PR, it's enough to merge :+1: .

ti-srebot (approved):

PR #938, opened 2020-09-21
Add http chaos abort & delay support with no sidecar design

Andrewmatilde:
Signed-off-by: Andrewmatilde <davis6813585853062@outlook.com>

Add http chaos abort & delay support with no sidecar design

Refactor the iptable_server & use iptable_server to support pod container port abort.
Just add code in tc_server to support delay for pod container port delay.
### Checklists
change the implement of podnetworkchaos iptable part.

Tests

- [ ] Unit test
- [ ] E2E test

Side effects

Related changes


controllers/httpchaos/iptables/iptables.go
```diff
@@ -0,0 +1,46 @@
...
+
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/chaos-mesh/chaos-mesh/controllers/common"
+	pb "github.com/chaos-mesh/chaos-mesh/pkg/chaosdaemon/pb"
+	"github.com/chaos-mesh/chaos-mesh/pkg/utils"
+)
+
+func SetIptablesChains(ctx context.Context, c client.Client, pod *v1.Pod, chains []*pb.Chain) error {
```
zhouqiang-cl, line 28: Please add a comment for this function

controllers/httpchaos/types.go
```diff
@@ -116,7 +120,61 @@ func (r *Reconciler) applyAllPods(ctx context.Context, pods []v1.Pod, chaos *v1a
...
-	//TODO: The way to connect with sidecar need be discussed & It will work after the sidecar add to the repo.
 	r.Log.Info("Try to inject Http chaos on pod", "namespace", pod.Namespace, "name", pod.Name)
+	err := r.SetIptables(ctx, pod, chaos)
+	if err != nil {
+		return err
+	}
 	return nil
 }
+
+func (r *Reconciler) SetIptables(ctx context.Context, pod *v1.Pod, chaos *v1alpha1.HTTPChaos) error {
```
zhouqiang-cl, line 131: ditto

pkg/chaosdaemon/iptables_server.go
```diff
@@ -42,144 +42,164 @@ func (s *daemonServer) SetIptablesChains(ctx context.Context, req *pb.IptablesCh
...
+			return "", fmt.Errorf("unknown chain direction %d", chain.Direction)
 		}
-	} else if chain.Direction == pb.Chain_OUTPUT {
-		iptables.ensureRule(&iptablesChain{
-			Name: "CHAOS-OUTPUT",
-		}, "-A CHAOS-OUTPUT -j "+chain.Name)
-		if err != nil {
-			return err
+		if len(chain.Ipsets) > 1 {
+			rule += fmt.Sprintf("-m set --match-set %s %%s -w 5 ", matchPart)
```
WangXiangUSTC, line 156: seems lack some argument

controllers/httpchaos/iptables/iptables.go
Andrewmatilde: Thanks for your review.

pkg/chaosdaemon/pb/chaosdaemon.proto
```diff
@@ -123,13 +123,29 @@ message IptablesChainsRequest {
...
-  string name = 1;
+  string table = 1;
+  enum commandType
+  {
+    NEW = 0;
+    ADD = 1;
+    DELETE = 2;
+  }
+  commandType command =2;
+  string ChainName = 3;
```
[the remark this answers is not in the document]
Andrewmatilde: Thanks for your review.

pkg/chaosdaemon/iptables_server.go
Andrewmatilde: The second placeholder was "%%s" that was just a real percent symbol i use it in the [https://github.com/chaos-mesh/chaos-mesh/pull/938/files/bf54990108dc99f8b4f1a78a4c1fe9cd9439100d#diff-7f7fb0cc6d572e887d964b8933b22145R82]

WangXiangUSTC: ok, how about adding a comment

Andrewmatilde: yes, I agree with you.

pkg/chaosdaemon/iptables_server.go
```diff
@@ -42,144 +42,169 @@ func (s *daemonServer) SetIptablesChains(ctx context.Context, req *pb.IptablesCh
...
+		} else {
+			err := iptables.ensureRule(&iptablesChain{
+				Name: chain.ChainName,
+			}, rule)
+			if err != nil {
+				log.Error(err, "error while add iptables chains")
+				return err
+			}
+		}
+	case pb.Chain_DELETE:
```
YangKeao, line 99: The design of `SetIptablesChains` is purely functional, which means it describes all chains (added by chaos-mesh) on a pod, but not an operation on the IPTables.

YangKeao:
If you tried to modify the IPTables rules or IPSets on a pod, it's really suggested to do it through modifying related `podnetworkchaos` (though it's not well documented). However, it doesn't support to set `dport` as a filter now, so you may need to extend it (rather than rewrite it :crying_cat_face: )

Why not extending `NetworkChaos` to add delay/partition on a specific port, as it's not related to HTTP essentially.

Andrewmatilde:
Sorry, I had a little confuse about the iptables_server. I think it was rewritten just for the podnetworkchaos but named as an operation on the IPTables, and I need to use the string match mode to handle some of the HTTP header in httpchaos. So I modify the iptables_server. 
Another reason is i don't want the user to set the ipset or ip information by their self, but I think it was difficult to turn the raw code into a no ipset version ,so I turn the code into a operation version.

YangKeao:
Sorry, I don't know what you mean :sob: . The `podnetworkchaos` is not created by the user directly. It is created by `NetworkChaos`. You can regard it as a tool to set iptables or ipsets on a pod, so if your code want to set iptables or ipsets, please create or modify the `podnetworkchaos` corresponding to the pod.

`podnetworkchaos` is created to make it easier to handle the situation when multiple chaos want to set iptables rules and ipsets on a same pod. You can read the implementation of `NetworkChaos`, which doesn't touch iptables and ipset directly but modifies the `podnetworkchaos`.

If the `/controllers/networkchaos/podnetworkmanager` is also needed by this PR, you could migrate it to another place (such as `/pkg/podnetworkmanager`).

If you need any help in understanding how `NetworkChaos` works, feel free to contact me (through slack) or leave comment here.

Andrewmatilde:
Sorry,I misunderstand it. Thanks for your reply .

PR #940, opened 2020-09-22
Encapculate controllers

YangKeao:

There are tons of duplicate codes in `controllers/`, which makes it hard to manage these controllers or create new chaos.

Setting up a router to register reconciler automatically and route these requests to endpoints (which implements `Apply` and `Recover`)

### Checklist

Tests

- [x] E2E test

Related changes

### Upgrade

As I have modified the name from `IoChaos` to `IOChaos`, then all users may need to reinstall the CRD (by `kubectl delete -f` and `kubectl apply -f`).


cmd/chaos-builder/main.go
```diff
@@ -0,0 +1,132 @@
...
+			return err
+		}
+		if info.IsDir() {
+			return nil
+		}
+
+		fset := token.NewFileSet()
+		file, err := parser.ParseFile(fset, path, nil, parser.ParseComments)
+		if err != nil {
+			log.Error(err, "fail to parse file")
```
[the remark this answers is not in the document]
YangKeao: Fixed

pkg/router/route_table.go
```diff
@@ -0,0 +1,94 @@
...
+	entry.Endpoints = append(entry.Endpoints, routeEndpoint{
+		RouteFunc:   routeFunc,
+		NewEndpoint: newEndpoint,
+	})
+}
+
+// SetupWithManager setups reconciler with manager
+func SetupWithManager(mgr ctrl.Manager) error {
+	for typ, end := range routeTable {
+		log.Info("setup reconciler with manager", "type", typ, "endpoint", end)
```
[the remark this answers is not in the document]
YangKeao: Fixed. I use `end.Name` to replace `end`

api/v1alpha1/podnetworkchaos_webhook.go
```diff
@@ -18,7 +18,7 @@ import (
 
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
-	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
```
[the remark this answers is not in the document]
YangKeao: only an alias, but imporing`"sigs.k8s.io/controller-runtime/pkg/runtime/log"` directly has been deprecated.

WangXiangUSTC (approved):
LGTM

cmd/chaos-builder/main.go
```diff
@@ -0,0 +1,133 @@
...
+var (
+	log = zap.Logger(true)
+)
+
+type metadata struct {
+	Type string
+}
+
+func main() {
+	generatedCode := `// Copyright 2020 Chaos Mesh Authors.
```
cwen0, line 38: use `const ` to define this.

cmd/chaos-builder/main.go
```diff
@@ -0,0 +1,133 @@
...
+
+		cmap := ast.NewCommentMap(fset, file, file.Comments)
+
+	out:
+		for node, commentGroups := range cmap {
+			for _, commentGroup := range commentGroups {
+				var err error
+				for _, comment := range commentGroup.List {
+					if strings.Contains(comment.Text, "+chaos-mesh:base") {
+						log.Info("build")
```
cwen0, line 89: Add the more detailed log or delete it

cmd/chaos-builder/main.go
```diff
@@ -0,0 +1,133 @@
...
+						baseDecl, ok := node.(*ast.GenDecl)
+						if !ok {
+							err = errors.Errorf("node is not a *ast.GenDecl")
+							log.Error(err, "fail to get type")
+							continue
```
cwen0, line 94: Why not return the error here?

cmd/chaos-builder/init.go
```diff
@@ -0,0 +1,46 @@
...
+		Chaos:     &{{.Type}}{},
+		ChaosList: &{{.Type}}List{},
+	})
+`
+
+func generateInit(name string) string {
+	tmpl, err := template.New("ini").Parse(initTemplate)
+	if err != nil {
+		log.Error(err, "fail to build template")
+		return ""
```
cwen0, line 33: Why ignore error here?

cmd/chaos-builder/main.go
```diff
@@ -0,0 +1,133 @@
...
+						}
+
+						if baseDecl.Tok != token.TYPE {
+							err = errors.Errorf("node.Tok is not token.TYPE")
+							log.Error(err, "fail to get type")
+							continue
```
cwen0, line 100: ditto

cmd/chaos-builder/main.go
```diff
@@ -0,0 +1,133 @@
...
+						baseType, ok := baseDecl.Specs[0].(*ast.TypeSpec)
+						if !ok {
+							err = errors.Errorf("node is not a *ast.TypeSpec")
+							log.Error(err, "fail to get type")
+							continue
+						}
+
+						generatedCode += generateImpl(baseType.Name.Name)
+						initImpl += generateInit(baseType.Name.Name)
+						break out
```
cwen0, line 112: If we define two chaos types in one file, what will happen?

cmd/chaos-builder/main.go
YangKeao: I use `codeHeader` as a const to replace this. `generatedCode` is not const.

cmd/chaos-builder/main.go
YangKeao: I add a line number and column number here.

cmd/chaos-builder/main.go
YangKeao: I agree with you that the error should be aggressive as a develop-time tool.

cmd/chaos-builder/init.go
YangKeao: Oops. I should panic here as it's unreachable (which should be guaranteed by the developer of this code generator)

cmd/chaos-builder/main.go
YangKeao: I replaced this `break` with `continue` so that we can define two chaos types in one file.

cmd/chaos-builder/init.go
```diff
@@ -0,0 +1,46 @@
...
+		panic(err)
+	}
+
+	buf := new(bytes.Buffer)
+	err = tmpl.Execute(buf, &metadata{
+		Type: name,
+	})
+	if err != nil {
+		log.Error(err, "fail to execute template")
+		return ""
```
cwen0, line 42: panic here?

YangKeao: Fixed

cwen0 (approved):
LGTM

WangXiangUSTC (approved):
LGTM

PR #991, opened 2020-09-28
add a check to avoid injectin NetworkChaos on a pod with `hostNetwork: true`

### LINKED ISSUE #990: Check whether pod is in `hostNetwork` before applying NetworkChaos
Applying network chaos on a `hostNetwork` pod is dangerous. We should avoid it through checking before applying.

YangKeao:
Fix #990


fewdan (approved):
LGTM

Gallardot (approved):

ti-srebot (approved):

Issue #666, opened 2020-06-30
Kind config in kind-cluster-build.sh can't run when the local docker registry is running


namco1992:
## Bug Report

**What version of Kubernetes are you using?**

1.15.6 which is the default version specified in the script.

**What did you do?**

```bash
hack/kind-cluster-build.sh
```

**What did you expect to see?**
A new kind cluster should be started.

**What did you see instead?**
```bash
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.15.6) 🖼 
 ✗ Preparing nodes 📦 📦 📦 📦  
ERROR: failed to create cluster: docker run error: command "docker run --hostname kind-control-plane --name kind-control-plane --label io.x-k8s.kind.role=control-plane --privileged --security-opt seccomp=unconfined --security-opt apparmor=unconfined --tmpfs /tmp --tmpfs /run --volume /var --volume /lib/modules:/lib/modules:ro --detach --tty --label io.x-k8s.kind.cluster=kind --net kind --restart=on-failure:1 --publish=127.0.0.1:5000:5000/TCP --publish=127.0.0.1:57565:6443/TCP kindest/node:v1.15.6" failed with error: exit status 125
Command Output: e1e494ee25ecade2a3c919223ccd96305a3db28cef9e8554f0bdeeaf4cccd76f
docker: Error response from daemon: driver failed programming external connectivity on endpoint kind-control-plane (239ac0c526190f879dd8fd564f69335b39ae6f29b1e1120c5b4356d86ddc748a): Error starting userland proxy: listen tcp 127.0.0.1:5000: bind: address already in use.
```

In short, the port 5000 specified in the `extraPortMappings` is conflicted with the local docker registry, it's the same in `install.sh` `install_kubernetes_by_kind` func. However, the [development guide](https://chaos-mesh.org/docs/development_guides/develop_a_new_chaos) assumes that we run a local docker registry.

I would like to suggest fixing it by the approach that kind recommended: https://kind.sigs.k8s.io/docs/user/local-registry/ I referred to this doc and set up the local env successfully. I can create a PR if it's acceptable.

cwen0:
@namco1992 This a good improvement. Looking forward to your PR!

namco1992:
@cwen0 Thanks! I will work on this.

zhouqiang-cl:
666

namco1992:
Not stale, will work on it soon, sorry.

PR #996, opened 2020-09-29
Start local docker registry for building kind cluster if it doesn't exist

namco1992:
Fixed https://github.com/chaos-mesh/chaos-mesh/issues/666.

1. Start local docker registry for kind if it doesn't exist
2. Bump the API version to support `containerdConfigPatches`

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code


hack/kind-cluster-build.sh
```diff
@@ -94,6 +94,16 @@ KUBECTL_BIN=${OUTPUT_BIN}/kubectl
 HELM_BIN=${OUTPUT_BIN}/helm
 KIND_BIN=${OUTPUT_BIN}/kind
 
+# create registry container unless it already exists
+reg_name='registry'
```
namco1992, line 98: I'm not sure if we want to accept arguments for setting the registry name and port. Any thoughts?

WangXiangUSTC (commented):
LGTM

fewdan (dismissed):
LGTM

ti-srebot (dismissed):

hack/kind-cluster-build.sh
cwen0: @namco1992 I think it might be better to use parameters here. For example, the registry does not use `5000`

namco1992: @cwen0 sure I will work on it.

hack/kind-cluster-build.sh
```diff
@@ -150,6 +159,9 @@ ${KIND_BIN} create cluster --config ${configFile} --image kindest/node:${k8sVers
 ${KIND_BIN} get kubeconfig --name=${clusterName} > ${kubeconfigPath}
 export KUBECONFIG=${kubeconfigPath}
 
+echo "connect the local docker registry to the cluster network"
+docker network connect "kind" "${reg_name}"
```
Yiyiyimu, line 163: Delete kind cluster would not remove the connection, and restart kind would comes with the error:
```
Error response from daemon: endpoint with name registry already exists in network kind
``` 

I think we need to see if it's already connected before this step, or check the error output

namco1992: @Yiyiyimu thanks for spotting this! I updated my PR to check the err and exit code.

zhouqiang-cl:
@fewdan @WangXiangUSTC @cwen0 @YangKeao  PTAL

cwen0 (approved):
LGTM

ti-srebot (approved):

PR #788, opened 2020-08-05
refractor network chaos to support multiple network chaos on the same pod

YangKeao:
### What problem does this PR solve?

Part of #718. This PR brings a lot of new features to NetworkChaos:

1. Support multiple network chaos on the same pod.

2. Support target selector for bandwidth limit

### What is changed and how does it work?

A new resource `PodNetworkChaos` is created to record the relation between `Iptables`, `IPSet` and `Tc` rules and the chaos definitions.

The implementation of the network partition is quite straightforward. If multiple network partition chaoses are set on the same pod, 
all these ips/cidrs will be blocked.

For bandwidth limit and netem chaos, the final traffic control tree will be like:

```
                +--------+
                |NETEM/TBF
                +---+----+
                    |
                    |
                +---+----+
                |NETEM/TBF
                +---+----+
                    |
                    |
                +---+----+
                |NETEM/TBF
                +---+----+
                    |
                    |
                +---v----+
                |PRIO    |
                +---+----+
                    |
    +-Filter 1------------Default---+
    |               |               |
    |           Filter 2            |
+---+----+      +--------+      +---+----+
|NETEM/TBF      |NETEM/TBF      |SFQ     |
+--------+      +--------+      +--------+
```

All netem/bandwidth chaos without a filter will be piped at the beginning of qdisc (and linked to root qdisc). The netem/bandwidth chaos with a filter will be set on one band of PRIO qdisc. 

**Limitation**: Only one netem/tbf qdisc with a filter will be applied, so that if you are applying two netem/tbf chaos and their targets overlapped, only one of them takes effect (for the pod in the intersection of their targets).

### Check List

Tests

- E2E test
- Manual test

Code changes

- Has Go code change

PR #1014, opened 2020-10-02
Encapsulate Recover of controllers

### LINKED ISSUE #961: Encapsulate multi-pod chaos
Many kinds of chaos are sharing the same `Recover` logic. A typical `Recover` function is like below:

```
func (r *Reconciler) Recover(ctx context.Context, req ctrl.Request, obj v1alpha1.InnerObject) error {

	podchaos, ok := obj.(*v1alpha1.PodChaos)
	if !ok {
		err := errors.New("chaos is not PodChaos")
		r.Log.Error(err, "chaos is not PodChaos", "chaos", obj)
		return err
	}

	if err := r.cleanFinalizersAndRecover(ctx, podchaos); err != nil {
		return err
	}

	r.Event(podchaos, v1.EventTypeNormal, utils.EventChaosRecovered, "")
	return nil
}

func (r *Reconciler) cleanFinalizersAndRecover(ctx context.Context, podchaos *v1alpha1.PodChaos) error {
	var result error

	for _, key := range podchaos.Finalizers {
		ns, name, err := cache.SplitMetaNamespaceKey(key)
		if err != nil {
			result = multierror.Append(result, err)
			continue
		}

		var pod v1.Pod
		err = r.Client.Get(ctx, types.NamespacedName{
			Namespace: ns,
			Name:      name,
		}, &pod)

		if err != nil {
			if !k8serror.IsNotFound(err) {
				result = multierror.Append(result, err)
				continue
			}

			r.Log.Info("Pod not found", "namespace", ns, "name", name)
			podchaos.Finalizers = utils.RemoveFromFinalizer(podchaos.Finalizers, key)
			continue
		}

		err = r.recoverPod(ctx, &pod, podchaos)
		if err != nil {
			result = multierror.Append(result, err)
			continue
		}

		podchaos.Finalizers = utils.RemoveFromFinalizer(podchaos.Finalizers, key)
	}

	if podchaos.Annotations[common.AnnotationCleanFinalizer] == common.AnnotationCleanFinalizerForced {
		r.Log.Info("Force cleanup all finalizers", "chaos", podchaos)
		podchaos.Finalizers = podchaos.Finalizers[:0]
		return nil
	}

	return result
}
```

The `Recover` and `cleanFinalizersAndRecover` functions for other reconcilers (such as StressChaos and TimeChaos) even have the same AST :crying_cat_face: . We should encapsulate them into a base struct to avoid the code clone.

Copy & Paste is a shame for a project and programmer :facepalm: . Let's remedy the embarrassing situation.

Yiyiyimu:

fix #961, currently only encapsulate chaos w/o managers.

Add a common `CleanFinalizersAndRecover` which implemented for a new struct.

### Checklist

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

```release-note
NONE
```


Yiyiyimu:
It would be a bit hard for encapsulation for controllers with managers, like `IoChaos` and `NetworkChaos`. They are quite similar but still there is some chaos-specific logic, for example #788 and #892 is made for networkChaos only.

Actually I'm not so sure if it would be a good idea to encapsulate this part. 
Not like `api` which every chaos type has the same API, encapsulating code here might make it hard to implement a change of certain chaos type, like the PR I listed. But I'm not sure about this inference, maybe certain change could work on different types but right now only one type is changed, so encapsulation would make it effective. @YangKeao Could you give me some suggestions?

YangKeao:
Maybe I should also try to make a `subresourcemanager` to replace `podiochaosmanager` and `podnetworkchaosmanager`... But these changes are blocked until #940 get merged. The code generator in #940 could provide a more convinient way to add getter and setter, and obviously I need `SetName`, `SetNamespace` in `subresourcemanager` :cry:  .

Yiyiyimu:
> Maybe I should also try to make a `subresourcemanager` to replace `podiochaosmanager` and `podnetworkchaosmanager`... But these changes are blocked until #940 get merged. The code generator in #940 could provide a more convinient way to add getter and setter, and obviously I need `SetName`, `SetNamespace` in `subresourcemanager` 😢 .

Sure I think we could wait until that time to update this part

Yiyiyimu:
Hi @YangKeao is the current implementation what we want

Yiyiyimu:
Hi @YangKeao PTAL

YangKeao:
Actually, I don't think generating all codes through a generator is a good idea. Only small parts of them (such as some getter or setter) should be done by code generation.

We can analyze the problem like this:

To implement `Recover`, we require an endpoint with `cleanFinalizersAndRecover` and `Event` methods. To implement `cleanFinalizersAndRecover` for an endpoint, we need a getter and setter for the `Finalizers`, a `Client`, and the most important part: `recoverPod`.

Only some of these requirements should be generated through the code generator (e.g. the `Finalizers` getter and setter for the custom resource).

Yiyiyimu:
Yeah it's actually also weird for me, to "generate" this large block of code that contains some logic. But since we could easily get/set finalizer so just only generate this part could not remove much duplicated code.

I have another idea (still a bit weird) that to build a common `cleanFinalizersAndRecover`, in which pass the endpoint as an argument (don't know if it is plausible), it could be like 
```Go
func (e *endpoint) Recover(ctx context.Context, req ctrl.Request, chaos v1alpha1.InnerObject) error {
	stressChaos, ok := chaos.(*v1alpha1.StressChaos)
	finalizers, err := cleanFinalizersAndRecover(ctx, e, chaos, stressChaos.Finalizers, stressChaos.Annotations[common.AnnotationCleanFinalizer])
	stressChaos.Finalizers = finalizers
}
// the common one
func cleanFinalizersAndRecover(ctx context.Context, e *endpoint, chaos v1alpha1.InnerObject, finalizers []string, forceClean string) error {
	...
	err = e.recoverPod(ctx, &pod, chaos)
	...
}
// do another type assertion in recoverPod
func (e *endpoint) recoverPod(ctx context.Context, pod *v1.Pod, chaos *v1alpha1.StressChaos) error {
	stressChaos, ok := chaos.(*v1alpha1.StressChaos)
	...
}
```

YangKeao:
@Yiyiyimu We can ignore the `*manager` part temporarily and encapsulate others first.

A straightforward plan could be defining a trait `RecoverPod` (maybe you can pick a better name), which contains one `RecoverPod` function.

And automatically implement `cleanFinalizersAndRecover` for all these traits. However, as the golang is not designed as a good language, you cannot `impl<T: IRecoverPod> ICleanFinalizers for T`. But you can construct a new `RecoverDelegate` struct whose only field is a `RecoverPod` interface. Than implement `Recover` and `cleanFinalizersAndRecover` for this `RecoverDelegate` struct.

Yiyiyimu:
Thanks for the suggestions and learned a lot from interface/trait!! And yeah the way implemented in Rust is way more elegant than in Go 😢

Yiyiyimu:
@cwen0 it seems the test goes stable now. PTAL

Yiyiyimu:
Hi @cwen0 @YangKeao could we review this pr at a faster speed? There are so many times files got conflicts.

cwen0:
> Hi @cwen0 @YangKeao could we review this pr at a faster speed? There are so many times files got conflicts.

Sorry for our late reply. I will review this pr ASAP.

controllers/networkchaos/partition/types.go
```diff
@@ -253,32 +253,30 @@ func (e *endpoint) SetChains(ctx context.Context, pods []v1.Pod, chains []v1alph
...
-		e.Log.Error(err, "cleanFinalizersAndRecover failed")
 		return err
 	}
 	e.Event(networkchaos, v1.EventTypeNormal, events.ChaosRecovered, "")
 
 	return nil
 }
 
-func (e *endpoint) cleanFinalizersAndRecover(ctx context.Context, networkchaos *v1alpha1.NetworkChaos) error {
+func (e *endpoint) cleanFinalizersAndRecover(ctx context.Context, chaos *v1alpha1.NetworkChaos) error {
```
cwen0, line 273: Here is also the repeated logic. Can we improve this logic?

cwen0 (commented):
Rest LGTM

controllers/networkchaos/partition/types.go
Yiyiyimu: The current implementation makes use of the same variables in 'recover' of different chaos. But for `networkchaos` and `iochaos` which use its own `chaosmanager`, things could not be solved in this way. @YangKeao [suggested](https://github.com/chaos-mesh/chaos-mesh/pull/1014#issuecomment-747188298) to leave this part away for now, until a common [`subresourcemanager`](https://github.com/chaos-mesh/chaos-mesh/pull/1014#issuecomment-706519681) is implemented

controllers/dnschaos/types.go
```diff
@@ -107,71 +110,32 @@ func (r *endpoint) Recover(ctx context.Context, req ctrl.Request, chaos v1alpha1
...
+		return err
 	}
+	r.Log.Info("Cancel DNS chaos to DNS service", "ip", service.Spec.ClusterIP)
 
-	return result
-}
-
-func (r *endpoint) recoverPod(ctx context.Context, pod *v1.Pod) error {
-	r.Log.Info("Try to recover pod", "namespace", pod.Namespace, "name", pod.Name)
+	r.cancelDNSServerRules(service.Spec.ClusterIP, config.ControllerCfg.DNSServicePort, chaos.Name)
```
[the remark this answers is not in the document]
Yiyiyimu: Fixed. Thanks!

cwen0 (approved):
LGTM

cwen0:
@YangKeao PTAL

YangKeao (approved):
LGTM

Issue #501, opened 2020-05-12
Support setting only duration field and not setting cron


cwen0:
## Feature Request

**Is your feature request related to a problem? Please describe:**

At present, we both require to set `cron` and `duration` fields when to define chaos experiment, but sometimes, we also just want to set `duration` and not setting `cron`. For example, we just want to inject chaos action once and last a specific time.

yeya24:
Help wanted!

qhh0205:
+1

STRRL:
Hi @cwen0 , I realize that `cron` and `at` can cover most of scenario, less is more.
So in my new design, there is no any complex things, only adding support for `at`. It will add a new field for `SchedulerSpec` like `At`, and modify two-phase controller for supporting only-one-shot.

Any problem about this design? 
I will make a PR in next days.

qhh0205:
any update? @STRRL @cwen0

STRRL:
@qhh0205 

Just a few days ago, I have a more clearly solution, could avoid making changes on two-phase controller.

I will implement `at` feature via the common Reconciler.

I need more time, please keeping your expectance. :)

STRRL:
I think we'd better implement this by the workflow. So I will close it now.

YangKeao:
Now (after release-2.0). Every `*Chaos` has only duration without cron 😺

PR #1016, opened 2020-10-03
feature: pause for certain duration

### LINKED ISSUE #544: Support pause chaos for a certain duration
## Feature Request

Now Chaos Mesh supports pausing chaos by annotation. However, the pause state is continued forever unless the user removes that annotation.

It would be good to support pausing a certain time duration. After that duration, the chaos resumes. The command should be like:

```shell
kubectl annotate networkchaos web-show-network-delay experiment.chaos-mesh.org/pauseTime=1h
```

Can refer to the implement of [experiment.chaos-mesh.org/pause](https://github.com/chaos-mesh/chaos-mesh/blob/5d409d77ae86c3896b15e2809c6b977999368479/api/v1alpha1/common_types.go#L26)

Yiyiyimu:

fix #544

In twophase:
1. Use `cron.ParseStandard` to parse and get the resume time after pause for certain time
2. Use `NextStart` and `NextRecover` to set the next cycle after pause

### Checklist

Todo list:
- [x] add e2e test
- [ ] do manual test
- [ ] add doc

Tests

- [ ] Unit test
- [x] E2E test
- [ ] No code

```release-note
support pause for certain duration
```


Yiyiyimu:
I think we could also implement `at` (as discussed in #501) to implement one-time chaos injection, with similar logic

api/v1alpha1/httpchaos_types.go
```diff
@@ -134,12 +134,12 @@ func (in *HTTPChaos) IsDeleted() bool {
 	return !in.DeletionTimestamp.IsZero()
 }
 
-// IsPaused returns whether this resource has been paused
-func (in *HTTPChaos) IsPaused() bool {
-	if in.Annotations == nil || in.Annotations[PauseAnnotationKey] != "true" {
-		return false
+// GetPause returns whether this resource has been paused
```
yeya24, line 137: If the method is renamed to `GetPause`, then this comment should be updated.

api/v1alpha1/httpchaos_types.go
```diff
@@ -134,12 +134,12 @@ func (in *HTTPChaos) IsDeleted() bool {
...
+func (in *HTTPChaos) GetPause() string {
+	if in.Annotations == nil || in.Annotations[PauseAnnotationKey] == "" {
```
yeya24, line 139: We can simplify the logic here.

``` go
func (in *HTTPChaos) GetPause() string {
	if in.Annotations == nil  {
        }
        return in.Annotations[PauseAnnotationKey]
```

controllers/twophase/types.go
```diff
@@ -97,7 +97,7 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 		status.Experiment.Phase = v1alpha1.ExperimentPhaseFinished
 		status.FailedMessage = emptyString
-	} else if chaos.IsPaused() {
+	} else if chaos.GetPause() != "" {
```
yeya24, line 100: It is better to have a check to ensure the format is valid.

api/v1alpha1/httpchaos_types.go
Yiyiyimu: Thank you for the suggestion. I thought it would be better to sacrifice code cleanness to improve the readability, like `if true` means it should not be paused, and vice versa. Don't know if that's needed.

api/v1alpha1/httpchaos_types.go
Yiyiyimu: thx! fixed.

controllers/twophase/types.go
Yiyiyimu: thx! fixed

api/v1alpha1/httpchaos_types.go
yeya24: I didn't get your point. Why do u think this way doesn't have readability :(? In your context, the getPause method is to return the value of the pause annotation. So that should be super clear and readable

yeya24: I didn't get your point. Why do u think this way doesn't have readability :(? In your context, the getPause method is to return the value of the pause annotation. So that should be super clear and readable. Let's change.

Yiyiyimu: Thank you for the suggestion. That makes sense to me and I'll make change. 
I'm still new to this so I might have some bad coding habit :(

cwen0:
@Yiyiyimu Can you fix these conflicts first?

Yiyiyimu:
> @Yiyiyimu Can you fix these conflicts first?

fixed

cmd/chaos-builder/impl.go
```diff
@@ -26,12 +26,17 @@ func (in *{{.Type}}) IsDeleted() bool {
 	return !in.DeletionTimestamp.IsZero()
 }
 
-// IsPaused returns whether this resource has been paused
-func (in *{{.Type}}) IsPaused() bool {
-	if in.Annotations == nil || in.Annotations[PauseAnnotationKey] != "true" {
-		return false
+// GetPause returns the annotation when the chaos needs to be paused
```
cwen0, line 29: How to cancel pause action?

controllers/twophase/types.go
```diff
@@ -92,15 +92,50 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+			if chaos.GetPause() == "true" {
+				err = r.Recover(ctx, req, chaos)
+				if err != nil {
+					r.Log.Error(err, "failed to pause chaos")
+					updateFailedMessage(ctx, r, chaos, err.Error())
+					return ctrl.Result{Requeue: true}, err
+				}
+			} else {
+				// Pause time is set
+				var pauseSpec v1alpha1.SchedulerSpec
```
cwen0, line 108: I think `SchedulerSpec` is not necessary here, because we can only use `time.Now().Add(duration)` to  implement this.

controllers/twophase/types.go
```diff
@@ -92,15 +92,50 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	} else if chaos.GetPause() != "" {
 		if status.Experiment.Phase == v1alpha1.ExperimentPhaseRunning {
 			r.Log.Info("Pausing")
 
-			err = r.Recover(ctx, req, chaos)
-			if err != nil {
-				r.Log.Error(err, "failed to pause chaos")
-				updateFailedMessage(ctx, r, chaos, err.Error())
-				return ctrl.Result{Requeue: true}, err
+			if chaos.GetPause() == "true" {
```
cwen0, line 99: `If chaos.GetPause() == "false“`?

controllers/twophase/types.go
```diff
@@ -92,15 +92,50 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+				// Pause time is set
+				var pauseSpec v1alpha1.SchedulerSpec
+				pauseSpec.Cron = "every " + chaos.GetPause()
+				nextStart, err := utils.NextTime(pauseSpec, time.Now())
+				if err != nil {
+					r.Log.Error(err, "failed to get the next start time after pause, check the format of pause input")
+					return ctrl.Result{}, err
+				}
+
+				err = r.Recover(ctx, req, chaos)
```
cwen0, line 116: If `Recover` is necessary，we can run this function before P99.

controllers/twophase/types.go
```diff
@@ -92,15 +92,50 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+				if err != nil {
+					r.Log.Error(err, "failed to pause chaos")
+					updateFailedMessage(ctx, r, chaos, err.Error())
+					return ctrl.Result{Requeue: true}, err
+				}
+
+				if nextStart.Before(chaos.GetNextRecover()) {
```
cwen0, line 123: Why do we need to check `nextStart.Before(chaos.GetNextRecover()) ` here?

controllers/twophase/types.go
```diff
@@ -92,15 +92,50 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+					chaos.SetNextStart(*nextStart)
+				} else {
+					cronGap := getCronGap(ctx, r, chaos)
```
cwen0, line 126: I didn't understand this logic here. Why we can't use the following code directly? 
```
chaos.SetNextStart(nextStart)
chaos.SetNextRecover(nextStart.Add(duration)
```

cmd/chaos-builder/impl.go
Yiyiyimu: Set `in.Annotations[PauseAnnotationKey] = ""`, as `SetPause` do. 
The current implementation is a bit counterintuitive.. Added a line of comment for the explanation. Thx~

controllers/twophase/types.go
Yiyiyimu: Thanks! Didn't know time could parse string to duration by then. fixed~

controllers/twophase/types.go
Yiyiyimu: So we still have the original pause&unpause, that if we set pause to `true`, we use the original method. Otherwise when we set pause to some duration, we use the new one.

controllers/twophase/types.go
Yiyiyimu: Thx! fixed

controllers/twophase/types.go
Yiyiyimu: Refactored this part. See below.

controllers/twophase/types.go
Yiyiyimu: We want to keep in the same pattern as cron told us. 
So if the resume time should be in waiting state, it would change status from pausing to waiting, and wait for the next start, rather than start immediately.

Yiyiyimu:
Hi @cwen0 I think it's ready to be reviewed, or we could wait till #1221 is merged to review this feature.

Yiyiyimu:
Hi @cwen0 I think it's ready to be reviewed 😁

BTW, the new two-phase scheduler really simply the logic of this feature a lot! 🚀 🚀

cmd/chaos-builder/impl.go
```diff
@@ -35,12 +35,18 @@ func (in *{{.Type}}) IsDeleted() bool {
...
+func (in *{{.Type}}) GetPause() string {
+	if in.Annotations == nil {
+		return ""
 	}
-	return true
+	return in.Annotations[PauseAnnotationKey]
+}
+
+// SetPause set the pausetime of annotation. Empty meams not paused
+func (in *{{.Type}}) SetPause(s string) {
```
cwen0, line 48: Why not add a single function to recover pause action?

Yiyiyimu: Good point! Fixed thanks!

controllers/twophase/types.go
```diff
@@ -73,13 +80,23 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		targetPhase = v1alpha1.ExperimentPhaseRunning
 	}
 
-	if chaos.IsPaused() {
+	if chaos.GetPause() != "" {
 		targetPhase = v1alpha1.ExperimentPhasePaused
+		// pause for certain duration, only set autoResume once
+		if chaos.GetPause() != "true" && status.Experiment.Phase != v1alpha1.ExperimentPhasePaused {
```
cwen0, line 86: Why not add a single annotation to represent the pause time?

cmd/chaos-builder/impl.go
```diff
@@ -35,12 +35,18 @@ func (in *{{.Type}}) IsDeleted() bool {
...
+	in.Annotations[PauseAnnotationKey] = s
```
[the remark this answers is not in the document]
Yiyiyimu: Currently it is changed to
```
// RecoverPause set the pausetime to empty, to stop pause
func (in *{{.Type}}) RecoverPause() {
	in.Annotations[PauseAnnotationKey] = ""
}
```
so the only case to call this function is when the auto-pause resume, and at this time `in.Annotations[PauseAnnotationKey]` would be always be some kind of `duration`. I think `in.Annotations` could not be `nil` here.

cmd/chaos-builder/impl.go
```diff
@@ -35,12 +35,18 @@ func (in *{{.Type}}) IsDeleted() bool {
...
+// return empty when annotations is not set, or pause key is set to empty
+func (in *{{.Type}}) GetPause() string {
+	if in.Annotations == nil {
+		return ""
 	}
-	return true
+	return in.Annotations[PauseAnnotationKey]
+}
+
+// RecoverPause set the pausetime to empty, to stop pause
```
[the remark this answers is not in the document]
Yiyiyimu: Thx!

controllers/twophase/types.go
Yiyiyimu: The reason I use the same annotation is that [as doc recommends](https://chaos-mesh.org/docs/user_guides/run_chaos_experiment#resume-a-chaos-experiment), we set annotation to nil to resume pause. And I think that's the same pattern we could also use here, from duration to nil.

If it would make you or maybe users confused, I could add another annotation like `pauseDuration`, compared to the original `pause`.

Yiyiyimu:
@cwen0 it seems the test goes stable now. PTAL

controllers/twophase/types.go
cwen0: I think to add another annotation like `pauseDuration` is better.

cwen0:
@Yiyiyimu Can you fix this CI?

Yiyiyimu:
Sorry for the late reply. Will fix it this weekend

cwen0:
@Yiyiyimu Can you fix this CI?

STRRL:

PR #1023, opened 2020-10-09
upgrade toda

YangKeao:

Don't override attributes for `lookup` operation.

Override `ino` or other attributes for `lookup` operation may cause an error in many situations, because kernel will use this `ino` to open a file. Overriding the result of `getattr` is enough, because the user can only get ino through `stat` syscall, which will lead to a `getattr` operation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -30,7 +30,7 @@ RUN apk add --no-cache curl tar
 
 WORKDIR /bin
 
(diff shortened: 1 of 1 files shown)


fewdan (approved):
LGTM

zhouqiang-cl (approved):
LGTM

ti-srebot (approved):

PR #1029, opened 2020-10-10
feat: use YAML format as experiment description

g1eny0ung:

This PR aims to use YAML format as an experiment detail in the dashboard.

#### Motivation:

Usually, we can create or update a chaos experiment by run `kubectl apply -f`. But the dashboard isn't, we reorder some spec values into different areas, especially in the `update` editor. The difficulty is The user will face a very huge JSON structure, but our documentation doesn't explain this (because we use inside the API). And also, to handle different chaos, our update codes add much more conditions and patches which are hard to maintain.

So to solve this problem it's better to make the experiment detail structure consistent with the CRD structure.

- JSON editor in the dashboard detail page to YAML editor
- Update `/experiments/detail/:uid` to return the same structure as the chaos YAML
- Update `/experiments/update` to receive the same structure as the chaos YAML
- Also update archives related codes
- Fix `DeleteIncompleteExperiments`, call it after fx providers already initialized.
- Some bugs fixes

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


fewdan:
Please resolve conflicting files.

fewdan (dismissed):
LGTM

pkg/store/dbstore/store.go
```diff
@@ -26,7 +26,7 @@ import (
 )
 
 var (
-	log = ctrl.Log.WithName("store")
+	log = ctrl.Log.WithName("store -> dbstore")
```
cwen0, line 29: why use `store -> dbstore` here?

g1eny0ung: Because I think `store -> dbstore` can be more clearly than `store` when debugging because the `dbstore` package is under the `store`.

pkg/store/experiment/experiment.go
```diff
@@ -26,190 +25,115 @@ import (
...
-// DeleteIncompleteExperiments implement core.ExperimentStore interface.
-func (e *experimentStore) DeleteIncompleteExperiments(_ context.Context) error {
-	return e.db.Where("finish_time IS NULL").Unscoped().
-		Delete(core.Event{}).Error
-}
-
-func (e *experimentStore) List(_ context.Context, kind, ns, name string) ([]*core.ArchiveExperiment, error) {
-	archives := make([]*core.ArchiveExperiment, 0)
-	query, args := constructQueryArgs(kind, ns, name, "")
+// ListMeta implement the core.ExperimentStore.ListMeta method.
```
cwen0, line 50: s/implement/implements/g

pkg/store/experiment/experiment.go
```diff
@@ -26,190 +25,115 @@ import (
...
 
-func (e *experimentStore) ListMeta(_ context.Context, kind, ns, name string) ([]*core.ArchiveExperimentMeta, error) {
-	archives := make([]*core.ArchiveExperimentMeta, 0)
-	query, args := constructQueryArgs(kind, ns, name, "")
-
-	db := e.db.Table("archive_experiments")
-	if len(args) > 0 {
-		db = db.Where(query, args)
-	}
+// FindByUID implement the core.ExperimentStore.FindByUID method.
```
cwen0, line 63: ditto

pkg/store/experiment/experiment.go
```diff
@@ -26,190 +25,115 @@ import (
...
+
 	nowTime := time.Now()
-	for _, exp := range expList {
-		if exp.Archived == false {
-			continue
-		}
+	for _, exp := range experiments {
 		if exp.FinishTime.Add(ttl).Before(nowTime) {
-			if err := e.db.Table("archive_experiments").Unscoped().Delete(*exp).Error; err != nil {
+			if err := e.db.Table("experiments").Unscoped().Delete(*exp).Error; err != nil {
```
cwen0, line 114: keep `archive_experiments` as the name of table

cwen0 (approved):
LGTM

zhouqiang-cl (approved):
LGTM

PR #1030, opened 2020-10-12
add integration test action

WangXiangUSTC:

1. Now we don't have tests for script files just like `install.sh`
2. The current E2E tests are complex to write

1. add an action for the integration test, so that we can test some script file like `install.sh`
2. add a test framework for test, and then we can use the shell script to write test cases
3. write a test case for network delay chaos as a demo

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


zhouqiang-cl (approved):
LGTM

fewdan (approved):
LGTM

Issue #688, opened 2020-07-10
NetworkChaos with target does not inject latency only between source and target pod.


nunziox:
[Not a contribution]

## Bug Report
**What version of Kubernetes are you using?**
v1.16.9-gke.6

**What did you do?**
On my test-bench project I have:
- An Nginx pod (ProxyPod) running on a k8s-cluster-A
- A test pod (TestPod) running on a k8s-cluster-A
- A ServiceA running on a VM outside a k8s-cluster-A
- A ServiceB running on a VM outside a k8s-cluster-A
I would like to inject latency only between TestPod -> ServiceA.
TestPod reach ServiceA via a /ServiceA endpoint  exposed by ProxyPod
TestPod also logs RTT to ServiceA and ServiceB which queries each second on a loop.
TestPod queries ServiceB directly (not via the proxy pod)
I used a NetworkChaos CRD specifying the target section to inject latency only between source and target pod. SourcePod was set to TestPod and target was set to ProxyPod. Both pods were correctly discovered and not errors where shown on logs.
I have set direction:"to" and mode:"all"

**What did you expect to see?**
TestPod RTT to ServiceA increased

**What did you see instead?**
TestPod RTT to ServiceA and ServiceB increased

YangKeao:
Sorry for the late reply. As I cannot reproduce this problem on my local cluster, I need more information to locate the problem. :cry: But now there isn't any simple way to get them. If you are willing to help us locate the problem, please follow the steps below:

1. Find the node which the test pod running on by `kubectl get pods  -o wide|grep PODNAME`, the 7th column is the node.

2. Find the chaos-daemon corresponding to this node by `kubectl get pods -n chaos-testing -o wide|grep NODENAME`.

3. Find the pid of test pod. There are several ways to do this. You can find docker container id of the pod with `kubectl`, and then use `docker` cli to get its pid. You can also read the related logs of corresponding chaos-daemon and it contains pid:

For example, the log 

```
2020-07-13T11:34:28.015Z	INFO	chaos-daemon-server	Delete qdisc on PID	{"pid": 5537}
```

means your test pod's pid is 5537

4. Use the following commands to get ipset, tc qdisc rules of the container:

```
kubectl exec CHAOS_DAEMON_POD_NAME -n chaos-testing -- /usr/bin/nsenter -n/proc/PID/ns/net -- ipset list
[...]

PR #1031, opened 2020-10-12
use iptables to classify packets

YangKeao:

#688 The default kernel of GKE is not built with `CONFIG_NET_EMATCH` and `CONFIG_NET_EMATCH_IPSET` so we cannot match ipset as a filter in tc.

This PR uses `iptables -A TC-TABLES-0 -m set --match-set A dst -j CLASSIFY --set-class 3:4 -w 5` to replace it.

### Checklist

Tests

- [x] E2E test

I have run the e2e test on GKE, and it works well.


cwen0 (approved):
LGTM

WangXiangUSTC (approved):
LGTM

PR #1054, opened 2020-10-15
upgrade toda

YangKeao:

Upgrade `toda`, so that the attribution override will work for inode.

### Checklist

Tests

I have tested it with `make example` and `make example-inject` in toda.


fewdan (approved):
LGTM

cwen0 (approved):
LGTM

Issue #925, opened 2020-09-18
I can't see  latency after I apply network-delay.yaml


XGameCloud:
## Question
Env:
  OFF-LINE cluster
  docker: 18.06
  k8s: 1.15.1
  chaos-mesh: v0.9.1
Description:
  I follow the Doc(v0.9.1). At User Guide -> Run Chaos Experiment -> Step 3. I can't see 10ms latency after I use 'kubectl apply -f network-delay.yaml'.
  Here are some info about the whole system.
-------------------------------------------------------------
$ kubectl get nodes
NAME       STATUS   ROLES    AGE   VERSION
xcloud37   Ready    <none>   13h   v1.15.1
xcloud47   Ready    master   23h   v1.15.1
xcloud48   Ready    <none>   23h   v1.15.1
xcloud49   Ready    <none>   20h   v1.15.1
xcloud50   Ready    <none>   20h   v1.15.1
------------------------------------------------------------
$ kubectl get pods -n chaos-testing -o wide
NAME                                        READY   STATUS    RESTARTS   AGE   IP            NODE       NOMINATED NODE   READINESS GATES
chaos-controller-manager-65cd5bfb8b-vh59l   1/1     Running   0          18h   10.244.1.12   xcloud48   <none>           <none>
chaos-daemon-cxbtg                          1/1     Running   0          18h   10.244.1.11   xcloud48   <none>           <none>
chaos-daemon-kh7qz                          1/1     Running   0          13h   10.244.4.2    xcloud37   <none>           <none>
chaos-daemon-knpk8                          1/1     Running   10         18h   10.244.2.10   xcloud49   <none>           <none>
chaos-daemon-tcmc8                          1/1     Running   0          18h   10.244.3.7    xcloud50   <none>           <none>
chaos-dashboard-59bcc758fc-spqg8            1/1     Running   0          14h   10.244.3.8    xcloud50   <none>           <none>
----------------------------------------------------------
$ kubectl get pods -o wide --show-labels
NAME                       READY   STATUS    RESTARTS   AGE   IP            NODE       NOMINATED NODE   READINESS GATES   LABELS
web-show-b4cf67576-qljnw   1/1     Running   0          42m   10.244.2.13   xcloud49   <none>           <none>            app=web-show,pod-template-h
----------------------------------------------------------
$ kubectl get NetworkChaos
NAME                     AGE
web-show-network-delay   17m
--------------------------------------------------------
[...]

Issue #824, opened 2020-08-19
networkchaos error:  connect: connection refused


alexec:
I'm getting the following error:

```
chaos-controller-manager-79f46bc85f-sth84 chaos-mesh 2020-08-19T18:25:25.141Z	ERROR	controllers.NetworkChaos	error while creating podnetworkchaos	{"reconciler": "networkchaos", "action": "netem", "error": "admission webhook \"vpodnetworkchaos.kb.io\" denied the request: rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing dial tcp 172.19.0.2:31767: connect: connection refused\""}
```

```yaml
apiVersion: chaos-mesh.org/v1alpha1
kind: NetworkChaos
metadata:
  name: web-show-network-delay
  namespace: chaos-testing
spec:
  action: delay # the specific chaos action to inject
  mode: all # the mode to run chaos action; supported modes are one/all/fixed/fixed-percent/random-max-percent
  selector: # pods where to inject chaos actions
    namespaces:
      - argo
    labelSelectors:
      "app": "argo-server"  # the label of the pod for chaos injection
  delay:
    latency: "10s"
  duration: "30s" # duration for the injected chaos experiment
  scheduler: # scheduler rules for the running time of the chaos experiments about pods.
    cron: "@every 60s"
```

Installed on K3S as follows:

```bash
 curl -sSL https://mirrors.chaos-mesh.org/latest/install.sh | bash -s -- --runtime containerd
```

alexec:
Ahhh
```
[...]

Issue #755, opened 2020-07-30
The stress-burn-cpu seems not working


del-zhenwu:
## Question

Before asking a question, make sure you have:

- Reviewed relevant Kubernetes information: Google your error messages and look at K8s docs.
- Searched open and closed [GitHub issues](https://github.com/chaos-mesh/chaos-mesh/issues?utf8=%E2%9C%93&q=is%3Aissue)
- Read the documentation:
  - [Chaos Mesh Readme](https://github.com/chaos-mesh/chaos-mesh)
  - [Chaos Mesh Doc](https://github.com/chaos-mesh/chaos-mesh/tree/master/doc)

I'm testing StressChaos with the example yaml, it run successfully, but I found the CPU usage was not changed factually
(using `kubectl watch pod *` )
, what's the possible reason for this.
```
2020-07-30T04:00:23.009Z	INFO	chaos-daemon-server	running command	{"command": "nsenter -p/proc/11347/ns/pid -- stress-ng --cpu 2 --cpu-load 100"}
2020-07-30T04:00:23.009Z	INFO	chaos-daemon-server	Start process successfully
2020-07-30T04:00:53.001Z	INFO	chaos-daemon-server	Canceling stressors	{"request": "instance:\"10355\" startTime:1596081622000 "}
2020-07-30T04:00:53.005Z	INFO	chaos-daemon-server	killing children for nsenter	{"pid": 10356}
2020-07-30T04:00:53.005Z	INFO	chaos-daemon-server	Successfully canceled stressors
2020-07-30T04:00:53.006Z	INFO	chaos-daemon-server	Stressors cancelled	{"request": "scope:POD target:\"docker://3759c19a59c455bcedac07ba8d692fd3b129a2bdcb565d6bda06c7ce808dbb13\" stressors:\" --cpu 2 --cpu-load 100\" "}
2020-07-30T04:00:58.007Z	INFO	chaos-daemon-server	Executing stressors	{"request": "scope:POD target:\"docker://3759c19a59c455bcedac07ba8d692fd3b129a2bdcb565d6bda06c7ce808dbb13\" stressors:\" --cpu 2 --cpu-load 100\" "}
2020-07-30T04:00:58.010Z	INFO	chaos-daemon-server	running command	{"command": "nsenter -p/proc/11347/ns/pid -- stress-ng --cpu 2 --cpu-load 100"}
2020-07-30T04:00:58.011Z	INFO	chaos-daemon-server	Start process successfully
[...]

Issue #743, opened 2020-07-27
Create networkchaos failed with the reason `i/o timeout`


del-zhenwu:
## Bug Report

**What version of Kubernetes are you using?**

**What did you do?**

podchaos can run successfully, but it failed when running networkchaos tests, here is the logs displayed on controller pods:
```
2020-07-27T09:22:51.826Z	INFO	util	Creating client to chaos-daemon	{"node": "test-worker9"}
2020-07-27T09:23:11.811Z	ERROR	controllers.NetworkChaos	flush pod ipset error	{"reconciler": "networkchaos", "action": "partition", "error": "rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing dial tcp 192.168.1.116:31767: i/o timeout\""}
github.com/go-logr/zapr.(*zapLogger).Error
	/go/pkg/mod/github.com/go-logr/zapr@v0.1.0/zapr.go:128
github.com/chaos-mesh/chaos-mesh/controllers/networkchaos/partition.(*Reconciler).Apply
	/src/controllers/networkchaos/partition/types.go:143
github.com/chaos-mesh/chaos-mesh/controllers/twophase.applyAction
	/src/controllers/twophase/types.go:197
github.com/chaos-mesh/chaos-mesh/controllers/twophase.(*Reconciler).Reconcile
	/src/controllers/twophase/types.go:158
github.com/chaos-mesh/chaos-mesh/controllers/networkchaos.(*Reconciler).scheduleNetworkChaos
	/src/controllers/networkchaos/types.go:91
github.com/chaos-mesh/chaos-mesh/controllers/networkchaos.(*Reconciler).Reconcile
	/src/controllers/networkchaos/types.go:51
github.com/chaos-mesh/chaos-mesh/controllers.(*NetworkChaosReconciler).Reconcile
	/src/controllers/networkchaos_controller.go:56
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:256
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.4.0/pkg/internal/controller/controller.go:232
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
[...]

PR #1074, opened 2020-10-18
feat: chaosctl for print debug info

### LINKED ISSUE #694: Debug information collector for chaos-mesh
## Feature Request

Nowadays it's really hard to collect debug information for some chaos. For example, the user should follow below four steps to collect debug information #688 :

> 1. Find the node which the test pod running on by kubectl get pods -o wide|grep PODNAME, the 7th column is the node.
> 2. Find the chaos-daemon corresponding to this node by kubectl get pods -n chaos-testing -o wide|grep NODENAME.
> 3. Find the pid of test pod. There are several ways to do this. You can find docker container id of the pod with kubectl, and then use docker cli to get its pid. You can also read the related logs of corresponding chaos-daemon and it contains pid
> 4. Get ipset, tc qdisc rules of the container

And it's also hard to write a script to collect the debug information, as there are several nodes affected by one chaos, and some information cannot be collected from CLI (the cli of `containerd` is really feature-poor :sob: ).

So we need a process to help us collect related debug information. This process can be shipped as a standalone executable file or embedded in controller-manager image. It can be accessed by running directly (with enough priviledge) or `kubectl exec`. 

Here is a list of information we needed for different kinds of chaos:

1. NetworkChaos. iptables, ipsets and tc qdisc/filter configs and related `podnetworkchaos`.

2. StressChaos: cgroup configs for stress-ng and target process/container.

3. IoChaos: mounts information

This list may become richer with the progress of development, so the extensibility should be considered carefully.

Yiyiyimu:

### What's this PR did
fix #694 

### Checklist
**SUPPORT CHAOS**
- [x] NetworkChaos
- [x] StressChaos
- [x] IoChaos

**DEBUG**
- [x] compare with expected (a demo for networkChaos)
- [x] move to ctrl.client from shell
- [x] chaos applied on multiple pods (larger radius): use utils.selectPods
- [x] replace extractFromJson with type assert (as how controllers do)
- [x] create daemon client locally to get pid more robustly
- [x] add test case
- [x] add backup for pod that not has `ls`, `ps` and `cat`
- [x] add feature to print log from all chaos-mesh components as `chaosctl logs`

**CTL**
- [x] add color to output title
- [x] use format "chaosctl debug xxxchaos"
- [x] support chaos in certain namespace
- [x] use kubectl compatible auto-completion as `chaosctl debug xxxchaos CHAOSNAME -n NAMESPACE`


Yiyiyimu:
Should open in personal repo

Yiyiyimu:
It seems that make check on the local environment could not fix the current issue in Github Action

pkg/debug/networkchaos/networkchaos.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+		if tailNum == "" {
+			return fmt.Errorf("could not found networkchaos related logs")
+		}
+	}
+	line := strings.Split(string(out), "\n")[0]
+	nsenterPath := regexp.MustCompile("(?:-n/proc/)(.*)(?:/ns/net)").FindStringSubmatch(line)[0]
+
+	// print out debug info
+	cmd := fmt.Sprintf("kubectl exec %s -n %s -- /usr/bin/nsenter %s -- ipset list", p.ChaosDaemonPodName, p.ChaosDaemonPodNamespace, nsenterPath)
```
YangKeao, line 64: Add a grpc call to collect information, rather than running `kubectl`.

You can use `github.com/chaos-mesh/chaos-mesh/pkg/utils.NewChaosDaemonClient` to create a grpc client to chaos-daemon on related node. There are a lot of examples in `controllers/podnetworkchaos/...`

YangKeao: It seems that there are a lot of `kubectl exec` in this PR :angry: . Try to organize them into some grpc calls and get information through gRPC.

pkg/debug/common/common.go
```diff
@@ -0,0 +1,174 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("run command '%s' failed with: %s", cmd, err.Error())
+	}
+
+	failedMessage := regexp.MustCompile("(?:Failed Message:\\s*)(.*)").FindStringSubmatch(string(out))
+	if len(failedMessage) != 0 {
+		return nil, fmt.Errorf("chaos failed with: %s", failedMessage)
+	}
+	podHier := []string{"Status", "Experiment", "Pod Records", "Name"}
+	podName, err := ExtractFromYaml(string(out), podHier)
```
YangKeao, line 131: Oops. Please use kubernetes client to get pod or any needed resources :crying_cat_face:, but not through `kubectl` commands.

YangKeao (changes requested):
It seems that you have misunderstood programming on kubernetes. I will talk with you in detail :cry:

pkg/chaosctl/cmd/completion.go
```diff
@@ -0,0 +1,76 @@
...
+`,
+	DisableFlagsInUseLine: true,
+	ValidArgs:             []string{"bash", "zsh", "fish", "powershell"},
+	Args:                  cobra.ExactValidArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		switch args[0] {
+		case "bash":
+			cmd.Root().GenBashCompletion(os.Stdout)
+		case "zsh":
+			cmd.Root().GenZshCompletion(os.Stdout)
```
yeya24, line 65: Cobra's support for `zsh` completion is very limited and usually doesn't work well. It would be good if we can reuse the code from `kubectl`. https://github.com/kubernetes/kubectl/blob/master/pkg/cmd/completion/completion.go#L145

yeya24: Don't be afraid of the hard coding scripts :). The change should be easy. Just rename `kubectl` to `chaosctl` would be fine.

pkg/debug/common/common.go
Yiyiyimu: Hi @YangKeao I just rewrite the shell-wise part with controller-runtime/client-go. Welcome review~

pkg/chaosctl/cmd/completion.go
Yiyiyimu: Hi @yeya24 Actually cobra's support for zsh just got updated after https://github.com/spf13/cobra/pull/1070, and I think kubectl community is also working on this update and migration on https://github.com/kubernetes/kubectl/issues/882.

pkg/chaosctl/cmd/debug.go
```diff
@@ -0,0 +1,130 @@
...
+
+	"github.com/chaos-mesh/chaos-mesh/pkg/debug/iochaos"
+	"github.com/chaos-mesh/chaos-mesh/pkg/debug/networkchaos"
+	"github.com/chaos-mesh/chaos-mesh/pkg/debug/stresschaos"
+)
+
+type DebugOptions struct {
+	ChaosName string
+	Namespace string
+}
```
yeya24, line 34: I feel like the k8s client should be part of this method and we can pass it to the debug method.

pkg/chaosctl/cmd/completion.go
Yiyiyimu: A screenshot for this in zsh:
```
 ~  chaosctl debug networkchaos -n kube-
kube-node-lease  kube-public      kube-system
```

pkg/chaosctl/cmd/debug.go
```diff
@@ -0,0 +1,130 @@
...
+			return err
+		}
+	default:
+		return fmt.Errorf("chaos not supported")
+	}
+	return nil
+}
+
+func flagCompletion(cmd *cobra.Command) error {
+	k8sConfig, err := config.GetConfig()
```
yeya24, line 100: Would be good to switch to https://github.com/kubernetes/cli-runtime and use its factory to initialize the k8s config and client.

This supports more kubectl options like `--context`, `--server` by default.

pkg/chaosctl/cmd/completion.go
Yiyiyimu: It just we need to generate clientset for chaos CRDs to support CRUD(I think controller-gen could do this), to make auto-complete for chaos work

yeya24: Cool! That's perfect. Thanks for letting me know.

pkg/chaosctl/cmd/completion.go
```diff
@@ -0,0 +1,76 @@
...
+		case "fish":
+			cmd.Root().GenFishCompletion(os.Stdout, true)
+		case "powershell":
+			cmd.Root().GenPowerShellCompletion(os.Stdout)
+		}
```
yeya24, line 70: So have you tested `fish` and `powershell` support in Cobra? If they are well supported already then it is fine.

pkg/debug/common/common.go
```diff
@@ -15,51 +15,75 @@ package common
...
+	ChaosDaemonNamespace string
 }
 
-func ExtractFromYaml(yamlStr string, str []string) (interface{}, error) {
+func init() {
+	_ = v1alpha1.AddToScheme(scheme)
+	_ = clientgoscheme.AddToScheme(scheme)
+}
+
+func UpperCaseChaos(str string) string {
```
yeya24, line 60: Any reason to make it public?

Yiyiyimu: My bad, thx!

pkg/chaosctl/cmd/completion.go
Yiyiyimu: I just test for `fish` which works fine. 
Since I don't have k8s on windows, I test `gh`(github cli) which use [the current version of cobra](https://github.com/cli/cli/blob/4a8ab49cf1d23c76d77a7cb69537b2bc66769afa/pkg/cmd/completion/completion.go) and it works fine. Good to go~

pkg/debug/networkchaos/networkchaos.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	// get nsenter path from log
+	var nsenterPathList []string
+	for _, tailNum := range []int64{100, 1000, 10000, -1} {
+		log, err := common.GetLog(p.ChaosDaemonName, p.ChaosDaemonNamespace, tailNum)
+		if err != nil {
+			return fmt.Errorf("get log failed with: %s", err.Error())
+		}
+		nsenterPathList = regexp.MustCompile("(?:-n/proc/)(.*)(?:/ns/net)").FindStringSubmatch(log)
```
YangKeao, line 52: I don't think you should get this information from log, which is obviously unstable.

You can get this information from `chaos-daemon`. It can lookup PID through the container id, and the namespace path is `/proc/$PID/ns/...`

However, communicating with `chaos-daemon` could be not that easy for a command line tool (outside the cluster). A possible solution is to place this binary in controller-manager image, and execute it. 

Or setup a `kubectl port-forward` (or directly import the go pkg and call it from your code, if it's simple enough.) in background to communicate with the `chaos-daemon`.

pkg/debug/common/common.go
```diff
@@ -0,0 +1,317 @@
...
+	parts := regexp.MustCompile("(.*)(chaos)").FindStringSubmatch(str)
+	return strings.Title(parts[1]) + strings.Title(parts[2])
+}
+
+// ExtractFromJson extract certain item from given string slice
+// String means the key of yaml
+// Number means the order in a slice
+func ExtractFromJson(chaos runtime.Object, str []string) (interface{}, error) {
```
YangKeao, line 68: Do you need something like `reflect` :rofl: ?

pkg/chaosctl/cmd/debug.go
Yiyiyimu: Thx I create the clientset (one for controller-runtime and another for go-client, since I need them both) in one place and send them everywhere as a parameter. 
https://github.com/chaos-mesh/chaos-mesh/blob/26eaa3a19746eb9698b61627e7a540ffc8d4b8ee/pkg/debug/common/common.go#L55-L58
https://github.com/chaos-mesh/chaos-mesh/blob/26eaa3a19746eb9698b61627e7a540ffc8d4b8ee/pkg/debug/common/common.go#L203-L208
It's just a bit ugly ... do you have some suggestions on this

pkg/chaosctl/cmd/debug.go
Yiyiyimu: Sure I'll try to only use one type of client

pkg/debug/common/common.go
Yiyiyimu: Thx for the reminder! That's what I need when I was using `kubectl` for dealing with different types of return values. I just need to return string for now.

pkg/debug/networkchaos/networkchaos.go
Yiyiyimu: Thanks for the suggestions! I'll look into it

pkg/debug/common/common.go
YangKeao: I mean... `reflect` could be a better way to get a string through a path (like '.Spec.Action') of struct (than serialize and deserialize into map).

If it's needed (for example, this deep field is common for a lot of chaos), you could add it to `chaos-builder` code generator. It's much safer than both `reflect` and `ExtractFromJson`

Yiyiyimu: Sorry I couldn't find how to use `reflect` to deal with this job... But I still managed to remove `ExtractFromJson`, by adding a type assertion part in each `Debug` function for different chaos, just like how controllers do. And after that I could get a string through a path.

BTW, it seems that I could not use getter in this circumstance. Since I could not type assert `runtime.Object` (which is what I get from `AllKinds`) to a common chaos type like `InnerObject`, I still have to do type assertion to specific chaos to get what I need. Is there someplace I misunderstood?

pkg/debug/common/common.go
```diff
@@ -0,0 +1,245 @@
...
+
+	if phase == "Waiting" {
+		waitTime := nextStart.Sub(time.Now())
+		fmt.Printf("Waiting for chaos to start, in %s\n", waitTime)
+		time.Sleep(waitTime)
+	}
+
+	// TODO: failed and maybe not appropirate to create manager here, to get client.Reader
+	// So could not parse fieldSelector for now
+	pods, err := utils.SelectPods(ctx, c, nil, selector)
```
Yiyiyimu, line 195: Failed and maybe not appropirate to create a manager here, to get client.Reader. So could not parse fieldSelector for now. Any suggestions?

pkg/debug/common/common.go
YangKeao: > I could not type assert `runtime.Object` (which is what I get from AllKinds) to a common chaos type like `InnerObject`

Why you cannot? In `pkg/router`, the `Reconcile` method for `Reconciler` asserts `runtime.Object` into `v1alpha1.InnerSchedulerObject`. If the object do have these methods, I think the type conversion is available.

Yiyiyimu: My bad, I used type assertion in the wrong way for it.

pkg/debug/networkchaos/networkchaos.go
Yiyiyimu: fixed. 
Just found out there is already an implementation of port-forward for [e2e test](https://github.com/chaos-mesh/chaos-mesh/blob/master/test/e2e/util/portforward/portforward.go), so the work is easier than expected.

Should I move this `portforward` folder to pkg/utils, since it's not only used by e2e anymore?

pkg/debug/networkchaos/networkchaos.go
```diff
@@ -0,0 +1,165 @@
...
+	cmd = fmt.Sprintf("/usr/bin/nsenter %s -- iptables --list", nsenterPath)
+	out, err = cm.Exec(daemonName, daemonNamespace, cmd, c.K8sClient)
+	if err != nil {
+		return fmt.Errorf("cmd.Run() failed with: %s", err.Error())
+	}
+	cm.Print("3. [iptables list]", 1, cm.ColorCyan)
+	cm.Print(string(out), 1, "")
+
+	cmd = fmt.Sprintf("/usr/bin/nsenter %s -- iptables --list", nsenterPath)
+	out, err = cm.Exec(daemonName, daemonNamespace, cmd, c.K8sClient)
```
STRRL, line 143: Seems the output of `iptables` hasn't be used.

pkg/debug/common/common.go
```diff
@@ -0,0 +1,266 @@
...
+// Exec executes certain command and returns the result
+// runtime-controller only support CRUD， use client-go client
+func Exec(pod string, ns string, cmd string, c *kubernetes.Clientset) (string, error) {
+	req := c.CoreV1().RESTClient().Post().
+		Resource("pods").
+		Name(pod).
+		Namespace(ns).
+		SubResource("exec")
+
+	req.VersionedParams(&v1.PodExecOptions{
```
STRRL, line 110: `Exec` will fail with multi-container pod. We need specify a container name.

pkg/debug/iochaos/iochaos.go
```diff
@@ -0,0 +1,73 @@
...
+	return nil
+}
+
+func debugEachPod(ctx context.Context, pod v1.Pod, daemon v1.Pod, chaos *v1alpha1.IoChaos, c *cm.ClientSet) error {
+	daemonName := daemon.GetObjectMeta().GetName()
+	daemonNamespace := daemon.GetObjectMeta().GetNamespace()
+
+	// print out debug info
+	cmd := fmt.Sprintf("ls /proc/1/fd -al")
+	out, err := cm.Exec(daemonName, daemonNamespace, cmd, c.K8sClient)
```
STRRL, line 57: Why execute `ls /proc/1/fd -al` in chaos-daemon.

It there any reference? 🙈

pkg/utils/grpc.go
```diff
@@ -32,19 +32,23 @@ const DefaultRPCTimeout = 60 * time.Second
 var RPCTimeout = DefaultRPCTimeout
 
 // CreateGrpcConnection create a grpc connection with given port
-func CreateGrpcConnection(ctx context.Context, c client.Client, pod *v1.Pod, port int) (*grpc.ClientConn, error) {
-	nodeName := pod.Spec.NodeName
-	log.Info("Creating client to chaos-daemon", "node", nodeName)
+func CreateGrpcConnection(ctx context.Context, c client.Client, pod *v1.Pod, port int, address string) (*grpc.ClientConn, error) {
```
STRRL, line 35: `pod` and `address` will not used at same time. `pod` has some extra logic to find out actual (node) address.

How about refactor this methods? Writing another “overloading” method, like
`CreateGrpcConnection(ctx,c,pod,port)` and `CreateGrpcConnectionWithAddress(ctx,c,address,port)`

STRRL (changes requested):
`chaoctl` is powerful!  Feel free to take any conversation with us. 😁

pkg/debug/common/common.go
```diff
@@ -0,0 +1,266 @@
...
+		return nil, fmt.Errorf("failed to get chaos %s: %s", chaosName, err.Error())
+	}
+	return chaos, nil
+}
+
+// GetChaosList returns chaos list limited by input
+func GetChaosList(ctx context.Context, chaosType string, chaosName string, ns string, c client.Client) ([]string, error) {
+	chaosType = upperCaseChaos(strings.ToLower(chaosType))
+	allKinds := v1alpha1.AllKinds()
+	chaosListIntf := allKinds[chaosType].ChaosList
```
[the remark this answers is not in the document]
Yiyiyimu: Interface.... it seems a weird abbreviation. Do you have some other suggestions, maybe use the full name

pkg/debug/common/common.go
```diff
@@ -0,0 +1,266 @@
...
+	// get podName
+	failedMessage := status.FailedMessage
+	if failedMessage != "" {
+		return nil, nil, fmt.Errorf("chaos failed with: %s", failedMessage)
+	}
+
+	phase := status.Experiment.Phase
+	nextStart := status.Scheduler.NextStart
+
+	if phase == "Waiting" {
```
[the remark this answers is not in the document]
Yiyiyimu: thx! fixed

pkg/chaosctl/cmd/root.go
```diff
@@ -0,0 +1,49 @@
...
+// Execute adds all child commands to the root command and sets flags appropriately.
+// This is called by main.main(). It only needs to happen once to the rootCmd.
+func Execute() {
+	if err := rootCmd.Execute(); err != nil {
+		fmt.Println(err)
+		os.Exit(1)
+	}
+}
+
+func init() {
```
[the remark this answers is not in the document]
Yiyiyimu: thx! fixed

pkg/chaosctl/main.go
```diff
@@ -0,0 +1,20 @@
+// Copyright 2020 Chaos Mesh Authors.
```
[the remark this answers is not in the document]
Yiyiyimu: Ah didn't know that before thanks!!

pkg/debug/networkchaos/networkchaos.go
Yiyiyimu: Ah my bad fixed thx!

pkg/utils/grpc.go
Yiyiyimu: thx for the suggestion! fixed

pkg/debug/common/common.go
Yiyiyimu: Thx for the suggestions! It seems still uses `client` from `controller-runtime` in the links. Is there someplace I misunderstood?

pkg/debug/iochaos/iochaos.go
YangKeao: @STRRL get all file descripters and their link target is an important debug information for `iochaos`

pkg/debug/common/common.go
STRRL: Ouch, I misremembered about that.  It seems chaos-dashboard does not using what I say.

TLDR, you could use `utils.SelectPods(ctx, c, nil, selector)` without any issue.

When we get `client.Client` from `Manager`(It's a struct of controller-manager), like what `chaos-dashboard` dose in [here](https://github.com/chaos-mesh/chaos-mesh/blob/6af3d9f70b760a50867425ff1be4d01ca3041c81/pkg/collector/server.go#L95), this client will contains a cache(index), this causing problem with fieldSelector.
When you dig into the constructor of `Manager` you could find out that `apiReader` is actually the normal client, `client` is a combination of `apiReader` and a `cache`, [ref](https://github.com/kubernetes-sigs/controller-runtime/blob/114431a4df156146338f5bb190af0a0664144b24/pkg/manager/manager.go#L380). And you also could find out that `client.Client` is a superset for `client.Reader`.

So the way you create `client.Client` will make a client without cache, it will NOT cause issues about fieldSelector. 🎉

pkg/debug/common/common.go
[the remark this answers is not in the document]
STRRL: I think using the full name is much better. 🤪

Makefile
```diff
@@ -134,6 +134,10 @@ binary: chaosdaemon manager chaosfs chaos-dashboard bin/pause bin/suicide
 watchmaker:
 	$(CGOENV) go build -ldflags '$(LDFLAGS)' -o bin/watchmaker ./cmd/watchmaker/...
 
+# Build chaosctl
+chaosctl:
+	go install ./cmd/chaosctl
```
YangKeao, line 139: `$(GO) build -ldflags '$(LDFLAGS)' -o bin/chaosctl ./cmd/chaosctl/*.go`, generate the executable files in `bin` folder.

pkg/debug/common/common.go
```diff
@@ -0,0 +1,273 @@
...
+	e2econfig "github.com/chaos-mesh/chaos-mesh/test/e2e/config"
+	"github.com/chaos-mesh/chaos-mesh/test/e2e/util/portforward"
+)
+
+var (
+	ColorReset = "\033[0m"
+	ColorRed   = "\033[31m"
+	ColorGreen = "\033[32m"
+	ColorCyan  = "\033[36m"
+	ColorBlue  = "\033[34m"
```
YangKeao, line 48: It would be better to use a package to print with color than printing ANSI escape code manually.

(`https://github.com/fatih/color` google gives me this one :smile_cat:  )

pkg/debug/stresschaos/stresschaos.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+	var cpuMountType string
+	if regexp.MustCompile("(cpu,cpuacct)").MatchString(string(out)) {
+		cpuMountType = "cpu,cpuacct"
+	} else {
+		cpuMountType = "cpu"
+	}
+
+	cmd = fmt.Sprintf("ps")
+	out, err = cm.Exec(podName, podNamespace, cmd, c.KubeCli)
```
YangKeao, line 81: When running a command in target pod, keep in mind that the target container could be really simple, which means there is no `cat`, no `ps`...

e.g. The widely used `debian:slim` base image doesn't have `ps` command (which is provided by `procps` package)

YangKeao: This could be a really difficult problem (especially for file related message collecting). For `ps` or other command which doesn't need to enter the mount namespace, you could run `nsenter` in `chaos-daemon` as an alternative.

However, for commands which need to enter the mount namespace (like `ls` and `cat`), things get much more complicated. A possible solution is to write a `c` or `rs` program to `setns` dynamically and read the folder or file.

pkg/debug/common/common.go
Yiyiyimu: That's it. I think I understand for now. Thank you for the explanation!! 

Actually there is a tiny issue, since if we need to select with fieldSelectors, current version would forcibly use `client.Reader`.
https://github.com/chaos-mesh/chaos-mesh/blob/6afb841b02df35a03bbca84419e70cb95eb8c8e1/pkg/utils/selector.go#L140-L150

Considering [another PR](https://github.com/chaos-mesh/chaos-mesh/pull/1112/files#diff-8a6b3b234279ff422c0e1c449141a18d54d29f47d383ab2d549a2a0788503ac8R95) encounter the same problem, we could add another case - when `reader` is `nil`, still `list` from `client`

pkg/debug/common/common.go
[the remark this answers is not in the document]
Yiyiyimu: sure thx! 😆

pkg/debug/common/common.go
STRRL: How about using `utils.SelectPods(ctx, c, c, selector)`, as you known `client.Client` is a superset for `client.Reader`.

Makefile
STRRL: BTW, @Yiyiyimu please remove the binary file `pkg/chaosctl/chaosctl `.

pkg/debug/common/common.go
Yiyiyimu: thx for pointing it out!

I noticed only podChaos (the container is killed so nothing we couldn't exec I guess) and [timeChaos](https://chaos-mesh.org/docs/chaos_experiments/timechaos_experiment#configuration-file) would specify the container to be injected with `containerName`. So for the rest, is using the `Containers[0]` is enough for this?

Makefile
Yiyiyimu: thx! fixed

pkg/debug/stresschaos/stresschaos.go
Yiyiyimu: Thanks for the remind! I'll look into it

pkg/debug/common/common.go
Yiyiyimu: Ah yeah that's a good idea! thx🎉

pkg/debug/stresschaos/stresschaos.go
Colstuwjx: just a reminder: `nsenter` also support running with `-m` option to enter the target process mount ns, and I suggest make an individual `nsenter` pkg, just like [kubectl-debug did](https://github.com/aylei/kubectl-debug/blob/1ab75d9b50c38a797c8634de01d48930e8147a23/pkg/nsenter/nsenter.go).

Yiyiyimu: @Colstuwjx Great to know that, thanks!

YangKeao: @Colstuwjx `pkg/bpm` package has included the function to generate `nsenter` arguments, and it's widely used in `chaos-daemon` package :smile_cat: 

Although `nsenter` support changing mount namespace, `ls` and `cat` may not exist in the target mnt namespace, which is the problem :disappointed:

zhouqiang-cl:
@Yiyiyimu Could you resolve conflicts, thank you

Yiyiyimu:
> @Yiyiyimu Could you resolve conflicts, thank you

fixed. thx~

pkg/debug/stresschaos/stresschaos.go
Yiyiyimu: Hi @YangKeao I think I fix this problem

pkg/debug/common/common.go
Yiyiyimu: fixed. thx!

pkg/debug/iochaos/iochaos.go
Yiyiyimu: My bad it should be executed in the target pod. 
Fixed thx!

Yiyiyimu:
Notice in several issues (#743, #824, #925, #1134) when `rpc connection error` happens, maintainers would ask users to check if the connection between controller manager and daemon is working, so I also add this feature in https://github.com/chaos-mesh/chaos-mesh/pull/1074/commits/46fb709af6f95dbbfbe5f41dc4f9b717f0ec082f.

STRRL:
Hi @Yiyiyimu , I want make a new feature about debug tool: fetch logs from components, just operation like `minikube logs`.

A sub-command `logs` will fetch logs from all pods like `chaos-daemon-xxx`, `chaos-controller-manager`, `chaos-dashboard`, and print it to stdout.

Would you help us to implement this feature? : P

Yiyiyimu:
Hi @STRRL Glad finally someone wants a new feature 🤣 No problem!

For options of `minikube logs`, I guess `length` and `node` is what we need. Do we need some other options?
```
  -f, --follow        Show only the most recent journal entries, and continuously print new entries as they are appended to the journal.
  -n, --length int    Number of lines back to go within the log (default 60)
      --node string   The node to get logs from. Defaults to the primary control plane.
      --problems      Show only log entries which point to known problems
```

STRRL:
I think we do not need these options now. 🤔
What I need is just fetching ALL LOGS easily. 🙈

Yiyiyimu:
Hi @STRRL feature supported. Welcome give it a try~

BTW I still support these two flags. For `length(tail)`, it would be a mess to print all logs but there might be cases that 60 lines are not enough. For `nodes`... ~idk maybe people would deploy chaos mesh on multiple nodes and only would like to show log from one?~ that's not possible user could directly get log from certain chaos-daemon on the node. Idk maybe I should remove this flag

STRRL:
As you already implement it, let's keep it for a while. : P

Yiyiyimu:
> As you already implement it, let's keep it for a while. : P

Thx sure 😄

STRRL:
Hi @Yiyiyimu , I saw you accomplish all items in the checklist, is this PR ready to review?  😁

If the answer is YES, could you resolve the conflicts with master branch? : P

Yiyiyimu:
Hi @STRRL conflict fixed. Thx for reviewing~

pkg/chaosctl/common/exec.go
```diff
@@ -0,0 +1,124 @@
...
+			}
+			err = fmt.Errorf("%s\nnsenter also failed with: %s", err.Error(), errNs.Error())
+		}
+		return "", err
+	}
+
+	return out, nil
+}
+
+func exec(ctx context.Context, pod v1.Pod, daemon v1.Pod, cmd string, c *kubernetes.Clientset) (string, error) {
```
YangKeao, line 52: Is the `daemon` a redundant argument here?

pkg/chaosctl/common/exec.go
```diff
@@ -0,0 +1,124 @@
...
+	// See grpc/grpc-go#3918 for detail. Could be resolved in the future
+	grpclog.SetLoggerV2(grpclog.NewLoggerV2(ioutil.Discard, ioutil.Discard, ioutil.Discard))
+	pid, err := GetPidFromPod(ctx, pod, daemon)
+	if err != nil {
+		return "", err
+	}
+	switch cmdSubSlice[0] {
+	case "ps":
+		nsenterPath := "-p/proc/" + strconv.Itoa(pid) + "/ns/pid"
+		nsCmd := fmt.Sprintf("mount -t proc proc /proc && %s && umount proc", cmd)
```
YangKeao, line 109: It's a bad idea to run these commands (as they have side effects).

Actually, as I remembered,  `chaos-daemon` has mouted the `procfs`...

pkg/chaosctl/common/exec.go
```diff
@@ -0,0 +1,124 @@
...
+		newCmd := fmt.Sprintf("/usr/bin/nsenter %s -- /bin/bash -c '%s'", nsenterPath, nsCmd)
+		return exec(ctx, daemon, daemon, newCmd, c)
+	case "cat", "ls":
+		// we need to enter mount namespace to get file related infomation
+		// but enter mnt ns would prevent us to access `cat`/`ls` in daemon
+		// so use `nsexec` to achieve using nsenter and cat together
+		if len(cmdSubSlice) < 2 {
+			return "", fmt.Errorf("%s should have one argument at least", cmdSubSlice[0])
+		}
+		newCmd := fmt.Sprintf("/usr/local/bin/nsexec %s %s", strconv.Itoa(pid), cmd)
```
YangKeao, line 119: Rather than judging through the command, it would be better to add an argument to the function.

Or add a `MntNs` function for `ProcessBuilder` in `pkg/bpm` and use it to build the command string.

pkg/chaosctl/common/exec.go
```diff
@@ -0,0 +1,124 @@
...
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/remotecommand"
+	kubectlscheme "k8s.io/kubectl/pkg/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client/config"
+)
+
+// Exec executes certain command and returns the result
+// runtime-controller only support CRUD， use client-go client
+func Exec(ctx context.Context, pod v1.Pod, daemon v1.Pod, cmd string, c *kubernetes.Clientset) (string, error) {
```
YangKeao, line 34: I don't think we should try to run the command in the target pod firstly. Running all commands from the daemon is easier to understand.

And the same command could have different behavior (especially for busybox districution like alpine), so running commands in `chaos-daemon` directly could be much more reliable.

STRRL (changes requested):
Expect comment on the codes, I think we could consider re-designing about `pkg/chaosctl/common/exec.go`.

Basically, we need two different way for executing a command:
1. execute a command in target container, target pod
2. execute a command somewhere else (usually in a chaos-daemon pod that sits on the same node for the target pod), then change the namespace into the target pod.

When we implement the second way, we could reuse the method which implements the first way, with `nsenter` or `nsexec`; And the caller should call the different method explicitly. (Although the first way is not commonly used, I think we should still expose it.)

So I prefer like this:  
exec.go
```golang
func Exec(ctx context.Context, target *v1.Pod, cmd []string, c *kuberentes.Clientset) (string, error){
    // execute command with streaming client
}

// use any other name you like, I'm really not good at name thing. 
func ExecBypass(ctx context.Context, target *v1.Pod, cmd []string, c *kuberentes.Clientset) (string, error){
    // find out the target chaos-daemon pod
    // modify the commands with nsenter or nsexec
    // execute in chaos-daemon by calling Exec()
}
```

What do you think about it? @Yiyiyimu  @YangKeao

pkg/chaosctl/debug/stresschaos/stresschaos.go
```diff
@@ -0,0 +1,167 @@
...
+		result.Pods = append(result.Pods, podResult)
+		if err != nil {
+			return fmt.Errorf("for %s: %s", podName, err.Error())
+		}
+	}
+	return nil
+}
+
+func debugEachPod(ctx context.Context, pod v1.Pod, daemon v1.Pod, chaos *v1alpha1.StressChaos, c *cm.ClientSet, result *cm.PodResult) error {
+	// cpu or memory chaos
```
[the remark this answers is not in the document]
Yiyiyimu: Thanks for the reminder! Which one do you think we should do to deal with this, exhaust what `stress-ng` provide(if there's not that much), add this single case, or just print that chaos in `stressngStressors` is not supported

pkg/chaosctl/common/exec.go
Yiyiyimu: My bad. Thx!

pkg/chaosctl/common/exec.go
Yiyiyimu: You are right, it would interfere with other processes between mount and umount. Here I do this to mount target pod `procfs` on chaos-daemon, so that `ps` would show information of target pod, not chaos-daemon[[Reference](https://lwn.net/Articles/531419/)].

As the article says, I could mount the child fs to somewhere else and use chroot to access it, but it somehow failed and I havem't found out why.

pkg/chaosctl/common/exec.go
Yiyiyimu: Yes you're right. 
If I take it right, as @STRRL said, we need to change the order of these two methods: firstly go through chaos-daemon, if that's not working then directly execute in target pod. Or we just remove the backup plan, so the errors from going through `chaos-damon` would not be covered and we could know it and fix it.

STRRL: I think we don't need the "backup plan".

All commands should already be considered about "use chaos-daemon" or "run directly" before it executes.

Yiyiyimu: @STRRL Sure that would be better. Thanks!

It just which kind of command should be "run directly". Or it's just for future consideration, that for now everything could go through chaos-daemon

pkg/utils/selector.go
```diff
@@ -160,13 +162,10 @@ func SelectPods(ctx context.Context, c client.Client, r client.Reader, selector
...
-				err := c.Get(ctx, types.NamespacedName{
-					Name: nodename,
-				}, &node)
-				if err == nil {
-					nodes = append(nodes, node)
-					continue
+				if err := c.Get(ctx, types.NamespacedName{Name: nodename}, &node); err != nil {
+					return nil, err
 				}
+				nodes = append(nodes, node)
```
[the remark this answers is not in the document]
Yiyiyimu: If we want to keep "best-effort", we still need to return error when no nodes were selected out after the loop, or we will miss the error that we set that wrong `Nodes`

pkg/chaosctl/cmd/logs.go
```diff
@@ -0,0 +1,118 @@
...
+  chaosctl logs -t 100 -n NODENAME`,
+		Run: func(cmd *cobra.Command, args []string) {
+			if err := o.Run(args, c); err != nil {
+				log.Fatal(err)
+			}
+		},
+		ValidArgsFunction: noCompletions,
+	}
+
+	logsCmd.Flags().Int64VarP(&o.tail, "tail", "t", 60, "Number of lines of recent log")
```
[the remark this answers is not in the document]
Yiyiyimu: Sure fixed🙈

Dockerfile
```diff
@@ -31,6 +31,7 @@ RUN apk add --no-cache curl tar
 WORKDIR /bin
 
 RUN curl -L https://github.com/chaos-mesh/toda/releases/download/v0.1.8/toda-linux-amd64.tar.gz | tar -xz
+RUN curl -L https://github.com/Yiyiyimu/nscat/releases/download/v0.1.1/nsexec-linx-amd64.tar.gz | tar -xz
```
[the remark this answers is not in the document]
Yiyiyimu: My bad. Thanks!

pkg/chaosctl/common/exec.go
Yiyiyimu: My bad logic here is not correct. I was trying to filter out commands `cat`/`ls` that didn't have arguments, which is meaningless. Still thank you for the idea, I'll add an argument parsing tool for `nsexec` later.

pkg/utils/selector.go
```diff
@@ -606,3 +605,83 @@ func RandomFixedIndexes(start, end, count uint) []uint {
...
+	var nodeObjects []runtime.Object
+	var nodes []v1.Node
+
+	for i := 0; i < n; i++ {
+		node := newNode(fmt.Sprintf("%s%d", namePrefix, i), label)
+		nodeObjects = append(nodeObjects, &node)
+		nodes = append(nodes, node)
+	}
+	return nodeObjects, nodes
+}
```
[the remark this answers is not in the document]
Yiyiyimu: Hi @STRRL we need to use them in both `pkg/chaosctl/common/` and `pkg/utils/`, so I could not place it in `_test.go`. Maybe I should create another file like `generate.go` in `pkg/utils` to place them

pkg/utils/grpc.go
```diff
@@ -33,18 +33,31 @@ var RPCTimeout = DefaultRPCTimeout
...
+			Name: nodeName,
+		}, &node)
+		if err != nil {
+			return nil, err
+		}
+		address = node.Status.Addresses[0].Address
 	}
 
-	conn, err := grpc.Dial(fmt.Sprintf("%s:%d", node.Status.Addresses[0].Address, port),
+	conn, err := grpc.Dial(fmt.Sprintf("%s:%d", address, port),
```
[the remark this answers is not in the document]
Yiyiyimu: Cool that would be better! Thx!

pkg/chaosctl/debug/stresschaos/stresschaos.go
```diff
@@ -0,0 +1,167 @@
...
+		return fmt.Errorf("Could not find stress-ng, StressChaos failed")
+	}
+	stressngPid := strings.Fields(stressngLine[0])[0]
+
+	cmd = fmt.Sprintf("cat /proc/1/cgroup")
+	out, err = cm.Exec(ctx, pod, daemon, cmd, c.KubeCli)
+	if err != nil {
+		return fmt.Errorf("run command '%s' failed with: %s", cmd, err.Error())
+	}
+	result.Items = append(result.Items, cm.ItemResult{Name: "/proc/1/cgroup", Value: string(out)})
```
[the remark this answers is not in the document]
Yiyiyimu: Hi @STRRL could you give me some reference about the time when PID 1 is not our target. Sorry I couldn't find those myself 😿 

For this problem, print `/proc/cgroups` could work well, but for the later one, which needs to list `/proc/$PID/fd`, is there any method that we could get to fetch the `init` process when the pid is not 1.

pkg/chaosctl/common/exec.go
STRRL: It's just for future consideration. XD

pkg/chaosctl/debug/stresschaos/stresschaos.go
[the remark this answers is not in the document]
STRRL: > Hi @STRRL could you give me some reference about the time when PID 1 is not our target.

You could take a look at this repo: [hello-kubernetes](https://github.com/paulbouwer/hello-kubernetes), it's a node web application. `npm` is the PID 1 process, and `node` is a child process.

>For this problem, print /proc/cgroups could work well, but for the later one, which needs to list /proc/$PID/fd, is there any method that we could get to fetch the init process when the pid is not 1.

How about printing fd information for each process?

pkg/chaosctl/common/exec.go
Yiyiyimu: I'm watching the [implementation of IoChaos](https://mp.weixin.qq.com/s/Noo0pEN3LqMyl8PJT5uvEQ), it seems the same methods could be used on this problem

pkg/chaosctl/debug/stresschaos/stresschaos.go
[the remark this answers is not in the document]
Yiyiyimu: Thank you @STRRL for the example! Learned a lot!
> How about printing fd information for each process?

I'm not so sure, it could be a mess when the target pod is complicated. But if maintainers could easily filter out what they need, I'm ok with this way.
cc @YangKeao what do you think

pkg/chaosctl/common/common.go
```diff
@@ -0,0 +1,352 @@
...
+		}
+	}
+	return nil
+}
+
+func checkConnForCtrlAndDaemon(ctx context.Context, daemons []v1.Pod, c *ClientSet) error {
+	ctrlSelector := v1alpha1.SelectorSpec{
+		LabelSelectors: map[string]string{"app.kubernetes.io/component": "controller-manager"},
+	}
+	ctrlMgrs, err := utils.SelectPods(ctx, c.CtrlCli, c.CtrlCli, ctrlSelector)
```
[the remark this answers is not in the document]
Yiyiyimu: Thanks! fixed

pkg/utils/selector.go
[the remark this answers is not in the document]
STRRL: Here is a similar situation to this. You could take a look on [controllers/test/types.go](https://github.com/chaos-mesh/chaos-mesh/blob/26fe5523a8ff9fdc3d934c43e5a0cd8ec50aec48/controllers/test/types.go#L128) and [controllers/podchaos/containerkill/containerkill_suite_test.go](https://github.com/chaos-mesh/chaos-mesh/blob/26fe5523a8ff9fdc3d934c43e5a0cd8ec50aec48/controllers/podchaos/containerkill/containerkill_suite_test.go#L62). 

I think you could even reuse it.

pkg/chaosctl/debug/stresschaos/stresschaos.go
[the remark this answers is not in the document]
Yiyiyimu: Sorry I made a mistake here. We need to compare cgroup of "the application" with `stress-ng`, as shown in [the original post](https://github.com/chaos-mesh/chaos-mesh/issues/755#issuecomment-667006237). So we still need to fetch the target pid.

pkg/chaosctl/common/exec.go
Yiyiyimu: Thanks again for the refactor suggestion! 
Since currently we need to execute commands in both chaos-mesh components (mostly daemon but one for controller-manager) and target pods, so the commands executed in chaos-mesh components could directly use the first direct one, which is easy to reuse when we need 😆

pkg/utils/chaosdaemon.go
```diff
@@ -43,14 +43,30 @@ func (c *GrpcChaosDaemonClient) Close() error {
...
 func NewChaosDaemonClient(ctx context.Context, c client.Client, pod *v1.Pod, port int) (ChaosDaemonClientInterface, error) {
+	return newChaosDaemonClient(ctx, c, pod, port, "")
+}
+
+// NewChaosDaemonClientLocally would create ChaosDaemonClient in localhost
+func NewChaosDaemonClientLocally(port int) (ChaosDaemonClientInterface, error) {
+	return newChaosDaemonClient(nil, nil, nil, port, "localhost")
+}
+
+func newChaosDaemonClient(ctx context.Context, c client.Client, pod *v1.Pod, port int, address string) (ChaosDaemonClientInterface, error) {
```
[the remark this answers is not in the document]
Yiyiyimu: Thanks! fixed

pkg/chaosctl/cmd/debug.go
```diff
@@ -0,0 +1,200 @@
...
+		var chaosResult cm.ChaosResult
+		chaosResult.Name = chaosNameList[i]
+
+		var err error
+		switch chaosType {
+		case "networkchaos":
+			err = networkchaos.Debug(ctx, chaos, c, &chaosResult)
+		case "stresschaos":
+			err = stresschaos.Debug(ctx, chaos, c, &chaosResult)
+		case "iochaos":
```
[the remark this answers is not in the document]
Yiyiyimu: Thanks for the suggestions! First time to know "magic values" 🙈  

Don't if it's ideal but I just predeclared them at the front, like `const networkChaos = "networkchaos"`. It's a bit redundant to use `iota` here, so it's not literally enum. Would that be acceptable?

pkg/utils/selector.go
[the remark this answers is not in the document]
Yiyiyimu: Yeah I noticed there is an almost same `GenerateNPods` function before. I think I could fix it in another PR after this PR got merged.

pkg/chaosctl/cmd/debug.go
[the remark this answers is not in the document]
STRRL: Well using Enums instead of directly using string is just to avoid dumb errors like mistyping. 😃

So when you use this enum, both string or `itoa` is acceptable for me. And you should expose the Enums for other packages could use it (if needed).

pkg/chaosctl/debug/stresschaos/stresschaos.go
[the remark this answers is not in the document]
STRRL: > I'm not so sure, it could be a mess when the target pod is complicated.

"Mess" is not a big issue, it just a thing that needs improvement with UX. 

Imagination: When you profiling a production environment:

- your tools dump a lot of information, you will finally find the reason by your own eyeball.
- your tools print not enough information, then you request your "System Administrator" for SSH permission for next profiling. ¯\\_(ツ)_/¯

Then talking about the UX, `grep` is the CLI tool that we always use. So a "grep-friendly" tool is our favorite. For example:

```text
# bad 🙁
[StressChaos for pod-xxx-xxxxx]
    cgroups:
        cpu: xxx
        cpuacct: xxx
        memory: xxx

# good 😆
StressChaos for pod-xxx-xxxxx cgroups cpu: xxx
StressChaos for pod-xxx-xxxxx cgroups cpuacct: xxx
StressChaos for pod-xxx-xxxxx cgroups memory: xxx
```

If you interest in it, we could create an issue and make discussions about this. And We could improve UX as another PR.

> Sorry I made a mistake here. We need to compare cgroup of "the application" with stress-ng, as shown in the original post. So we still need to fetch the target pid.

I still suggest print the all information. 🤪

pkg/chaosctl/debug/iochaos/iochaos.go
```diff
@@ -0,0 +1,74 @@
...
+		if err != nil {
+			return fmt.Errorf("for %s: %s", podName, err.Error())
+		}
+	}
+	return nil
+}
+
+func debugEachPod(ctx context.Context, pod v1.Pod, daemon v1.Pod, chaos *v1alpha1.IoChaos, c *cm.ClientSet, result *cm.PodResult) error {
+	// print out debug info
+	cmd := fmt.Sprintf("ls -l /proc/1/fd")
```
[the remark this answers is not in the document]
YangKeao: Yes. The `IOChaos` will try to inject for all the process in the pid namespace.

pkg/chaosctl/debug/stresschaos/stresschaos.go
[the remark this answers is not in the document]
Yiyiyimu: Thank you I never thought about the way "grep-friendly" before! It's not hard to change I'll commit about this later.

> I still suggest print the all information.

What if we set a flag for this, so users could explicitly set their application PID. Then the problem would turn to by default do we print all or print 1.

pkg/chaosctl/cmd/debug.go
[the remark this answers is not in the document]
Yiyiyimu: Got it. Thanks for the explanation!!

pkg/chaosctl/debug/stresschaos/stresschaos.go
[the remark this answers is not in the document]
STRRL: > Considering this tool is mainly for users to provide debug information, so IDE or just browser is what we use to filter out information. In this case I think a better hierarchy might be more intuitive?

It's not so emergency, I suggest make this in another PR. : )

As you already refactored the codes: from " just printing it" to "using structs to describe information then "rendering" it to stdout". So you can easily change the way of rendering.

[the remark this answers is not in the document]
Yiyiyimu: Sure thx I'll note it down for later.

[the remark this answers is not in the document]
Yiyiyimu: Hi @STRRL print all info fixed PTAL

pkg/chaosctl/debug/stresschaos/stresschaos.go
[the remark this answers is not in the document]
Yiyiyimu: Currently return "not-supported" error for this situation

Yiyiyimu:
@YangKeao @STRRL Thanks for the review! PTAL

pkg/chaosctl/debug/stresschaos/stresschaos.go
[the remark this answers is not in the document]
YangKeao: Get `cgroup` related information is also valuable for other stress chaos.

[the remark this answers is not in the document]
Yiyiyimu: Thanks! Fixed

pkg/chaosctl/common/exec.go
```diff
@@ -0,0 +1,104 @@
...
+	switch cmdSubSlice[0] {
+	case "ps":
+		nsenterPath := "-t " + strconv.Itoa(pid) + " -p -m"
+		newCmd := fmt.Sprintf("/usr/bin/nsenter %s -- /bin/sh -c '%s'", nsenterPath, cmd)
+		return Exec(ctx, daemon, newCmd, c)
+	case "cat", "ls":
+		// we need to enter mount namespace to get file related infomation
+		// but enter mnt ns would prevent us to access `cat`/`ls` in daemon
+		// so use `nsexec` to achieve using nsenter and cat together
+		newCmd := fmt.Sprintf("/usr/local/bin/nsexec %s %s", strconv.Itoa(pid), cmd)
```
YangKeao, line 99: I came up with a better idea to implement a more general `nsexec` 🍻 . You can find the source code and documents at https://github.com/YangKeao/nsexec, and it will be brought to `chaos-daemon` in https://github.com/chaos-mesh/chaos-mesh/pull/1260/ .

I don't know whether it's a big project to migrate to [YangKeao/nsexec](https://github.com/YangKeao/nsexec).

pkg/utils/selector.go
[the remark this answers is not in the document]
YangKeao: It seems that we don't need to keep `best-effort`. `fail-fast` is great 🍻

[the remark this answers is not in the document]
YangKeao: It seems that we don't need to keep best-effort. fail-fast is great 🍻 @cwen0 What's your opinion?

pkg/chaosctl/common/exec.go
Yiyiyimu: Cool I test it and it works fine. I'll commit the change till #1260 got merged.

pkg/chaosctl/common/exec.go
```diff
@@ -0,0 +1,104 @@
...
+	grpclog.SetLoggerV2(grpclog.NewLoggerV2(ioutil.Discard, ioutil.Discard, ioutil.Discard))
+	pid, err := GetPidFromPod(ctx, pod, daemon)
+	if err != nil {
+		return "", err
+	}
+	switch cmdSubSlice[0] {
+	case "ps":
+		nsenterPath := "-t " + strconv.Itoa(pid) + " -p -m"
+		newCmd := fmt.Sprintf("/usr/bin/nsenter %s -- /bin/sh -c '%s'", nsenterPath, cmd)
+		return Exec(ctx, daemon, newCmd, c)
```
STRRL, line 94: It does not make sense.
It finally accesses the "process" and "mount" namespace of the target container. 
It can not resolve "the docker image of the target container does NOT have `ps`". 🤔

Fetching cgroups with debugging StressChaos is also a similar situation: you need to access /proc, so you need to change mnt namespace, but there might no binary file(ls) in target mnt namespace.

I think this problem could only resolve by https://github.com/YangKeao/nsexec. 😰

STRRL: Oh just for resolving for "ps", it could use your `nsexec ls /proc` as another solution. 🤔

Yiyiyimu: You are ABSOLUTELY right. I don't know why I made this mistake.. Thanks for pointing it out! Keao's `nsexec` could solve it pretty well~

pkg/chaosctl/common/exec.go
Yiyiyimu: @YangKeao Updated to #1260, PTAL

Yiyiyimu:
Hi @YangKeao @STRRL PTAL

pkg/portforward/portforward.go
```diff
@@ -49,9 +49,10 @@ type PortForward interface {
 // portForwarder implements PortForward interface
 type portForwarder struct {
 	genericclioptions.RESTClientGetter
-	ctx    context.Context
-	config *rest.Config
-	client kubernetes.Interface
+	ctx       context.Context
+	config    *rest.Config
+	client    kubernetes.Interface
+	enableLog bool
```
STRRL, line 55: I think we do not need `enableLog`, more logs is not a bad thing.

Yiyiyimu: Hi @STRRL the thing is somewhat like this. By default, everytime we run `portforward`, one line of log would be print to show the pod and the port. I don't know if it would be helpful for debugging e2e test but I still keep it there.
https://github.com/chaos-mesh/chaos-mesh/blob/478d00d01bb0f9fb08a1085428a7da8c8f9df4e8/test/e2e/util/portforward/portforward.go#L80-L85

Considering we would need the same function for `chaosctl`, it could be kind of misleading for users if several lines of this kind of log print out. Since there is no something like a switch to disable `klog` output, I add this flag to disable the output when needed.

STRRL: Oh, you are right. 🙈

STRRL (approved):
LGTM
 
Thanks for your contribution @Yiyiyimu!

pkg/portforward/portforward.go
Yiyiyimu: Cool 😁

YangKeao (approved):
LGTM

It finally arrives! (Maybe we need a document later 😺 ) 🍻

zhouqiang-cl:
Is there have readme about this tool? @Yiyiyimu

Issue #1091, opened 2020-10-22
chaos-daemon support run with non-root user


WangXiangUSTC:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Now, chaos-daemon runs with root user as default, the permissions are too high that some users want to limit it.

**Describe the feature you'd like:**

chaos daemon support run with a non-root user, and add a config for it, user can choose run with root or non-root user.

WangXiangUSTC:
this issue is frozen because the emand is not clear

PR #1094, opened 2020-10-23
chaos-daemon: support run as non-root user

WangXiangUSTC:

fix issue https://github.com/chaos-mesh/chaos-mesh/issues/1091

Now, chaos-daemon runs with root user as default, the permissions are too high that some users want to limit it.

chaos daemon support run with a non-root user, and add a config for it, user can choose run with root or non-root user.
And command will run with `sudo`.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [x] Manual test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


YangKeao:
Is there any difference between run as root or run as normal user in a container (about security)? . For example, you can simply rewrite `/etc/sudoers` in your image and allow anyone to run `sudo` without password.

And I don't think the newly installed `sudo` will be recorded by any security audit software :( .

PR #1104, opened 2020-10-27
upgrade toda to 0.1.8

YangKeao:

1. implement `fsyncdir`, so that TiKV will not crash under heavy load.

2. use a customized [glibc](https://github.com/chaos-mesh/toda-glibc) to work with Local Thread Storage.

3. the opened file will not be affected by unlinking 

### Checklist

Tests

- [ ] E2E test


zhouqiang-cl:
😃

fewdan (dismissed):
LGTM

YangKeao:
I have reproduced this e2e failure on centos:

```
[2020-10-30 05:37:41.220425 +00:00] INFO [toda] src/main.rs:68: parse injector configs
[2020-10-30 05:37:41.220488 +00:00] INFO [toda] src/main.rs:70: inject with config [Latency(LatencyConfig { filter: FilterConfig { path: Some("/var/run/data/*"), methods: None, percent: 100 }, latency: 10ms })]
thread 'main' panicked at 'failed to spawn thread: Os { code: 22, kind: InvalidInput, message: "Invalid argument" }', /rustc/16957bd4d3a5377263f76ed74c572aad8e4b7e59/src/libstd/thread/mod.rs:619:29
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

It seems that the `clone` syscall has returned an error.

YangKeao:
The former error is because in early version linux kernel, process in different pid namespaces cannot share sig handlers.

```c
/*
* If the new process will be in a different pid or user namespace
* do not allow it to share a thread group or signal handlers or
* parent with the forking task.
*/
if (clone_flags & CLONE_SIGHAND) {
        if ((clone_flags & (CLONE_NEWUSER | CLONE_NEWPID)) ||
             (task_active_pid_ns(current) !=
                         current->nsproxy->pid_ns))
              return ERR_PTR(-EINVAL);
}
```

YangKeao:
> The former error is because in early version linux kernel, process in different pid namespaces cannot share sig handlers.
> 
> ```c
> /*
> * If the new process will be in a different pid or user namespace
> * do not allow it to share a thread group or signal handlers or
> * parent with the forking task.
> */
> if (clone_flags & CLONE_SIGHAND) {
>         if ((clone_flags & (CLONE_NEWUSER | CLONE_NEWPID)) ||
>              (task_active_pid_ns(current) !=
>                          current->nsproxy->pid_ns))
>               return ERR_PTR(-EINVAL);
> }
> ```

[This commit](https://github.com/torvalds/linux/commit/faf00da544045fdc1454f3b9e6d7f65c841de302) in kernel removes this limitation for higher version linux.

cwen0 (approved):
LGTM

fewdan (approved):
LGTM

cwen0:
@YangKeao PTAL

PR #1109, opened 2020-10-28
Remove privileged and add some capabilities

### LINKED ISSUE #1101: chaos daemon run with privileged = false
## Feature Request

now we will run chaos daemon with `privileged=true`, all capabilities will be added when privileged is true. Some users think we should limit the authority.

set `privileged=false`, and add some capabilities we really need.

YangKeao:

Fix #1101 

Chaos Daemon now works with `privileged: false`. In replace, it creates a `fuse` device and add it to cgroup device list manually.


cwen0 (approved):
LGTM

WangXiangUSTC (approved):
LGTM

Issue #1107, opened 2020-10-28
chaos-dashboard could not play well with namespace scoped deployment


STRRL:
## Bug Report

After https://github.com/chaos-mesh/chaos-mesh/pull/971, chaos-controller-manager could work with rbac restriction.

Sin this PR did not contain any modification about chaos-dashboard, chaos-dashboard is unusable with namespace scoped mode. 

Chaos dashboard prints many logs about "forbidden" about resources.

https://pastebin.ubuntu.com/p/GwbzjZTZq8/

We should add a flag on chaos-dashboard like what we do on chaos-controller-manager.

Issue #1072, opened 2020-10-17
Make pkg doesn't reference controller gloabl variables (e.g. common.ControllerCfg )


Colstuwjx:
## Feature Request

**Is your feature request related to a problem? Please describe:**

It's not a good idea to reference controller(a.k.a business code) variables from pkg(a.k.a common libraries), it may introduces cycle imports, and I did find one, the common.ControllerCfg, it has been used by [pkg/utils/selector.go](https://github.com/chaos-mesh/chaos-mesh/blob/212da2d5095666bde793a9b4d636abfc38496f0f/pkg/utils/selector.go#L88) and [pkg/router/reconciler.go](https://github.com/chaos-mesh/chaos-mesh/blob/cf9e1ca11670df27b2b8995b0f4942cf96142330/pkg/router/reconciler.go#L46).

**Describe the feature you'd like:**

Remove  controller references from pkg, since `pkg` should be a common library without business code.

**Describe alternatives you've considered:**

None.

**Teachability, Documentation, Adoption, Migration Strategy:**

We need to refactor all related codes and testcases, it would introduce break changes since we need to tweak the function arguments, but I think we should do this, and setup the rules for avoiding do such reference anymore.

Colstuwjx:
BTW, there is a [twophase reference](https://github.com/chaos-mesh/chaos-mesh/blob/cf9e1ca11670df27b2b8995b0f4942cf96142330/pkg/router/reconciler.go#L87) which is also a potential cycle import point.

YangKeao:
Yes. I have faced a lot of cycle import problems while developing #940 :sob: 

I felt that the lack of `namespace` is the source of cycle import, because we have to split things into package (which cannot cycle import) in go but cannot split them into `namespace` (which is safe to cycle import).

I don't know whether creating such a lot small packages is a good practice, as it will really bring a lot of trouble :facepalm: .

Colstuwjx:
TBH, I'm not clear understand with your so-called `namespace`, if it means package namespace, I suggest we setup some code regulations for it. From now, we add packages under the `pkg` folder as needed, and of course, it may introduce cycle import risks.

There are some suggestions, in my opinion:

1. common libraries could put into `utils` or sth like `lib`, make it as a purge lib;
2. common logic encapsulate, like pkg/router, put them into an sub-directory, e.g. core/router;
3. api related code, put them into sth like `api`, on root directory or under `pkg`

[...]

PR #1110, opened 2020-10-28
make chaos dashboard be compliable with namespace scoped mode

STRRL:

Fix  https://github.com/chaos-mesh/chaos-mesh/issues/1107, chaos-dashboard is uncompliable  with namespace scoped mode.

- Add ENV to chaos-dashboard in helm deployment yaml.
- Adding new options while creating `Manager`, only watching ONE specific namespace.
- Change some `List` action while user query something from k8s api(detail in Side effects).
> appending at 2020-10-29
- change `/api/common/namespace` to `/api/common/chaos-available-namespaces`, means only return namespaces which could inject chaos (chaos experiment explosion scope)
- remove permission about `list` on `namespaces` with namespace scoped mode

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

It changes `/api/experiments` behavior under namespace scoped mode: origin parameter `namespace` will be replaced by `targetNamespace`.

Related changes

```release-note
NONE
```

### DIFF
--- a/helm/chaos-mesh/templates/chaos-dashboard-deployment.yaml
+++ b/helm/chaos-mesh/templates/chaos-dashboard-deployment.yaml
@@ -48,6 +48,18 @@ spec:
             - name: TZ
               value: {{ .Values.timezone | default "UTC" }}
             {{- end }}
+            - name: TARGET_NAMESPACE
+              value: {{ .Values.controllerManager.targetNamespace }}
+            - name: CLUSTER_SCOPED
+              value: "{{ .Values.clusterScoped }}"
+            {{- if .Values.controllerManager.allowedNamespaces }}
+            - name: ALLOWED_NAMESPACES
+              value: {{ .Values.controllerManager.allowedNamespaces }}
+            {{- end }}
+            {{- if .Values.controllerManager.ignoredNamespaces }}
+            - name: IGNORED_NAMESPACES
+              value: {{ .Values.controllerManager.ignoredNamespaces }}
+            {{- end }}
           volumeMounts:
             - name: storage-volume
               mountPath: {{ .Values.dashboard.persistentVolume.mountPath }}
--- a/install.sh
+++ b/install.sh
@@ -1192,6 +1192,10 @@ spec:
               value: "2333"
             - name: TZ
               value: ${timezone}
+            - name: TARGET_NAMESPACE
+              value: chaos-testing
+            - name: CLUSTER_SCOPED
+              value: "true"
           volumeMounts:
             - name: storage-volume
               mountPath: /data
--- a/install.sh
+++ b/install.sh
@@ -951,7 +951,6 @@ metadata:
 rules:
   - apiGroups: [ "" ]
     resources:
-      - namespaces
       - nodes
     verbs: [ "get", "list", "watch" ]
 ---
--- a/install.sh
+++ b/install.sh
@@ -952,6 +952,7 @@ rules:
   - apiGroups: [ "" ]
     resources:
       - nodes
+      - namespaces
     verbs: [ "get", "list", "watch" ]
 ---
 # Source: chaos-mesh/templates/controller-manager-rbac.yaml
--- a/pkg/apiserver/common/common.go
+++ b/pkg/apiserver/common/common.go
@@ -65,7 +65,7 @@ func Register(r *gin.RouterGroup, s *Service) {
 	endpoint := r.Group("/common")
 
 	endpoint.POST("/pods", s.listPods)
-	endpoint.GET("/namespaces", s.getNamespaces)
+	endpoint.GET("/namespaces", s.getChaosAvailableNamespaces)
 	endpoint.GET("/kinds", s.getKinds)
 	endpoint.GET("/labels", s.getLabels)
 	endpoint.GET("/annotations", s.getAnnotations)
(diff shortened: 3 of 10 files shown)


helm/chaos-mesh/templates/chaos-dashboard-deployment.yaml
```diff
@@ -48,6 +48,18 @@ spec:
...
+            - name: CLUSTER_SCOPED
+              value: "{{ .Values.clusterScoped }}"
+            {{- if .Values.controllerManager.allowedNamespaces }}
+            - name: ALLOWED_NAMESPACES
+              value: {{ .Values.controllerManager.allowedNamespaces }}
+            {{- end }}
+            {{- if .Values.controllerManager.ignoredNamespaces }}
+            - name: IGNORED_NAMESPACES
+              value: {{ .Values.controllerManager.ignoredNamespaces }}
+            {{- end }}
```
WangXiangUSTC, line 62: seems that `ALLOWED_NAMESPACES` and `IGNORED_NAMESPACES` is useless

WangXiangUSTC (commented):
rest LGTM

helm/chaos-mesh/templates/chaos-dashboard-deployment.yaml
STRRL: These ENV will  inject into `common.ControllerCfg` which was dependent in `pkg/utils/selector.go`.

Many API in `pkg/apiserver/common/common.go` use `pkgutils.SelectPods` which located in `pkg/utils/selector.go`. For keeping same action, I think we should set these ENVs.

This is a bad-smell  which already pointed by https://github.com/chaos-mesh/chaos-mesh/issues/1072 😥.

WangXiangUSTC (dismissed):
LGTM

STRRL:
PTAL @cwen0

cwen0:
@STRRL  Does this configuration affect the [common API](https://github.com/chaos-mesh/chaos-mesh/blob/master/pkg/apiserver/common/common.go)?

STRRL:
> @STRRL Does this configuration affect the [common API](https://github.com/chaos-mesh/chaos-mesh/blob/master/pkg/apiserver/common/common.go)?

Yes, kind of fixing a bug.

There are two ENV `ALLOWED_NAMESPACES` and `IGNORED_NAMESPACES ` are NOT set in old deployment yaml file. These ENVs are needed by `pkgutils.SelectPods()` in common API.

cwen0:
> > @STRRL Does this configuration affect the [common API](https://github.com/chaos-mesh/chaos-mesh/blob/master/pkg/apiserver/common/common.go)?
> 
> Yes, kind of fixing a bug.
> 
> There are two ENV `ALLOWED_NAMESPACES` and `IGNORED_NAMESPACES ` are NOT set in old deployment yaml file. These ENVs are needed by `pkgutils.SelectPods()` in common API.

If we set namespace scope, `common.getNamespaces` may return a error. Can we avoid this behavior?

STRRL:
> If we set namespace scope, `common.getNamespaces` may return a error. Can we avoid this behavior?

Currently service account dose hold the permission about `list` with `namespace` in cluster level, [link](https://github.com/chaos-mesh/chaos-mesh/blob/4cde3cafcee1158752176f5b035389edb9f2e6cb/helm/chaos-mesh/templates/controller-manager-rbac.yaml#L46). So I don't think it will throw an error.

STRRL:
More changes:
- change `/api/common/namespace` to `/api/common/chaos-available-namespaces`, means only return namespaces which could inject chaos (chaos experiment explosion scope)
- remove permission about `list` on `namespaces` with namespace scoped mode

PTAL @cwen0  @WangXiangUSTC @g1eny0ung

pkg/config/dashboard.go
```diff
@@ -29,6 +29,11 @@ type ChaosDashboardConfig struct {
 	EnableLeaderElection bool   `envconfig:"ENABLE_LEADER_ELECTION"`
 	Database             *DatabaseConfig
 	PersistTTL           *PersistTTLConfig
+	// ClusterScoped means control Chaos Object in cluster level(all namespace),
+	ClusterScoped bool `envconfig:"CLUSTER_SCOPED" default:"true"`
+	// TargetNamespace is the target namespace to injecting chaos.
+	// It only works with ClusterScoped is false;
+	TargetNamespace string `envconfig:"TARGET_NAMESPACE" default:""`
```
g1eny0ung, line 36: Should we set the `TARGET_NAMESPACE` to `chaos-testing`? I noticed that the installation script defaults to chaos-testing.

install.sh
```diff
@@ -951,7 +951,6 @@ metadata:
 rules:
   - apiGroups: [ "" ]
     resources:
-      - namespaces
```
WangXiangUSTC, line 954: when `CLUSTER_SCOPED` is true, still need the privileged to list namespace

pkg/config/dashboard.go
STRRL: It's configured by `controllerManager.targetNamespace` in `helm/chaos-mesh/values.yaml`, [link](https://github.com/chaos-mesh/chaos-mesh/blob/master/helm/chaos-mesh/values.yaml#L43).

We have a bash script that rendering `instasll.sh` from helm files.

install.sh
STRRL: I will fix it. 🙈

pkg/config/dashboard.go
g1eny0ung: Understood, thx for the explanation.~

pkg/apiserver/common/common.go
```diff
@@ -65,7 +65,7 @@ func Register(r *gin.RouterGroup, s *Service) {
 	endpoint := r.Group("/common")
 
 	endpoint.POST("/pods", s.listPods)
-	endpoint.GET("/namespaces", s.getNamespaces)
+	endpoint.GET("/chaos-available-namespaces", s.getChaosAvailableNamespaces)
```
WangXiangUSTC, line 68: how about keep the `/namespaces` and add the new API `/chaos-available-namespaces`

STRRL: Great! I am working on this.

STRRL: Hi @WangXiangUSTC, it finished in 4a056c7 , please take a look.

WangXiangUSTC (approved):
LGTM

WangXiangUSTC:
Does this PR need cherry pick to 1.0?

STRRL:
> Does this PR need cherry pick to 1.0?

Yes it is.

g1eny0ung (approved):
LGTM

Issue #1095, opened 2020-10-23
Dashboard Pod crashes after deployment of StressChaos/TimeChaos Experiments


FischlerA:
## Bug Report

**What version of Kubernetes are you using?**
AWS EKS 1.18

**What did you do?**
I installed Chaos-Mesh using the[ helm Chart ]( https://github.com/chaos-mesh/chaos-mesh/tree/master/helm/chaos-mesh 
) for the installation. After successful installation of the ChaosMesh everything is working fine when deploying e.g. PodChaos Experiments. 
As soon as i deploy either TimeChaos or StressChaos Experiments the Dashboard Pod crashes.

```
2020-10-23T07:57:55.747Z        DEBUG   controller-runtime.controller   Successfully Reconciled {"controller": "timechaos", "request": "confluentplatform/time-shift-example"}
2020-10-23T07:57:55.767Z        DEBUG   controller-runtime.controller   Successfully Reconciled {"controller": "timechaos", "request": "confluentplatform/time-shift-example"}
2020-10-23T07:58:05.021Z        DEBUG   controller-runtime.controller   Successfully Reconciled {"controller": "timechaos", "request": "confluentplatform/time-shift-example"}
panic: reflect: call of reflect.Value.IsZero on zero Value

goroutine 627 [running]:
reflect.Value.IsZero(0x0, 0x0, 0x0, 0x1bde45c)
        /usr/local/go/src/reflect/value.go:1118 +0x4b6
github.com/chaos-mesh/chaos-mesh/api/v1alpha1.(*TimeChaos).GetChaos(0xc0020db340, 0xc0020db340)
        /src/api/v1alpha1/zz_generated.chaosmesh.go:925 +0x204
github.com/chaos-mesh/chaos-mesh/api/v1alpha1.(*TimeChaosList).ListChaos(0xc0004369a0, 0x2113940, 0xc002084900, 0x20dc880)
        /src/api/v1alpha1/zz_generated.chaosmesh.go:955 +0x149
github.com/chaos-mesh/chaos-mesh/pkg/apiserver/experiment.(*Service).state.func1(0x445736, 0x1eef340)
        /src/pkg/apiserver/experiment/experiment.go:959 +0x118
golang.org/x/sync/errgroup.(*Group).Go.func1(0xc001805f20, 0xc0020849c0)
[...]

PR #1111, opened 2020-10-29
Add test for generated codes

YangKeao:

To avoid problem like #1095 , we need to add some tests to the generated code. These tests could only avoid panic but don't check the logic.


cmd/chaos-builder/main.go
```diff
@@ -48,20 +48,29 @@ const codeHeader = `// Copyright 2020 Chaos Mesh Authors.
...
 import (
 	"reflect"
 	"time"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
-
 `
+	generatedTest := codeHeader + `
+import (
```
cwen0, line 63: Can we define these imports as a const and avoid to define these in main function? such as `const testImport = `xxxx`

cmd/chaos-builder/main.go
```diff
@@ -131,5 +141,12 @@ func init() {
 	}
 	fmt.Fprint(file, generatedCode)
 
+	generatedTest += testInit
+	file, err = os.Create("./api/v1alpha1/zz_generated.chaosmesh_test.go")
+	if err != nil {
+		log.Error(err, "fail to create file")
```
WangXiangUSTC, line 147: how about exit with 1?

cmd/chaos-builder/test.go
```diff
@@ -0,0 +1,169 @@
...
+
+const testInit = `
+func init() {
+	faker.AddProvider("ioMethods", func(v reflect.Value) (interface{}, error) {
+		return []IoMethod{LookUp}, nil
+	})
+}
+`
+
+const testTemplate = `
```
WangXiangUSTC, line 29: how about defining it after `testInit`?

YangKeao: Sorry, I don't understand this comment.

WangXiangUSTC: 🤕 just ignore it

cmd/chaos-builder/main.go
YangKeao: Fixed.

cmd/chaos-builder/main.go
YangKeao: Fixed

WangXiangUSTC (dismissed):
LGTM

cwen0 (approved):
LGTM

cwen0 (approved):

g1eny0ung (approved):

Issue #1092, opened 2020-10-22
dashboard support authority management


WangXiangUSTC:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Chaos can be created and deleted by anyone who has access to the dashboard, and the authority of dashboard is very high, it means the Dashboard is a big security hazard.

**Describe the feature you'd like:**

The dashboard should support authority management, login by service account and token created by k8s. And the user can only do the chaos limited by the role bind with the service account.

torblerone:
Is there already a rough date for the feature? I would like to use Chaos Mesh in the company I work for, but access control is a **critical feature** for cross-team use and distribution of authorizations, e.g. for defining experiments on certain namespaces for specific users.

The granularity of the authorizations could optimally be mapped to the various activities that can be used with Chaos Mesh, especially in connection with Kubernetes (see also k8s Best Practices for RBAC with admin / edit / view).

WangXiangUSTC:
You can look the proposal for details https://github.com/chaos-mesh/rfcs/blob/main/text/2020-10-22-authn-and-authz-on-chaos-dashboard. We plan to implement this feature next month.

WangXiangUSTC:
You can comment in this issue if have any question @torblerone

torblerone:
Many Thanks @WangXiangUSTC ! It is nice to see that you are thinking and developing in this direction too. As I said before, it would be great if you could distribute role permissions in a very fine-grained manner. I'm thinking of permissions for certain types of chaos, namespaces, maybe even services. But that shouldn't downplay your state of affairs, you have orientated yourself very well to CN when it comes to a role concept! Thanks for that!

WangXiangUSTC:
The permissions depend on what can do with the service account, the dashboard will not manage the authority. 
This feature is in developing, you can see https://github.com/chaos-mesh/chaos-mesh/pull/1112.

WangXiangUSTC:
solve this issue via https://github.com/chaos-mesh/chaos-mesh/pull/1188

PR #1112, opened 2020-10-29
dashboard: use k8s client with token

WangXiangUSTC:

https://github.com/chaos-mesh/chaos-mesh/issues/1092

Chaos can be created and deleted by anyone who has access to the dashboard, and the authority of the dashboard is very high, it means the Dashboard is a big security hazard.

The dashboard should use the k8s client with a token to visit the k8s cluster, and the token is provided by the user in the dashboard frontend. Users can only do chaos which is allowed by the token generated by service account.

1. add a k8s client pool implement by LRU, and it will save several clients in the cache.
2. HTTP API will extract the token and get k8s client from the client pool.

### Checklist

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


pkg/clientpool/client.go
```diff
@@ -0,0 +1,93 @@
...
+
+// K8sClients is an object of Clients
+var K8sClients *Clients
+
+// Clients is the client pool of k8s client
+type Clients struct {
+	sync.RWMutex
+
+	localConfig *rest.Config
+	clients     *lru.Cache
```
STRRL, line 35: Hi @WangXiangUSTC , I think we don't need cache the clients.

As implementation in [kubernetes-dashboard](https://github.com/kubernetes/dashboard/blob/master/src/app/backend/client/manager.go#L415), it will create new clients for each requests.

WangXiangUSTC: maybe several users will use the chaos dashboard at the same time, and will create the client for each request. I think the cache can improve the performance.

pkg/apiserver/common/common.go
```diff
@@ -80,14 +74,20 @@ func Register(r *gin.RouterGroup, s *Service) {
...
+
 	exp := &core.SelectorInfo{}
 	if err := c.ShouldBindJSON(exp); err != nil {
 		c.Status(http.StatusBadRequest)
 		_ = c.Error(utils.ErrInvalidRequest.WrapWithNoMessage(err))
 		return
 	}
 	ctx := context.TODO()
-	filteredPods, err := pkgutils.SelectPods(ctx, s.kubeCli, s.reader, exp.ParseSelector())
+	filteredPods, err := pkgutils.SelectPods(ctx, kubeCli, nil, exp.ParseSelector())
```
WangXiangUSTC, line 90: use nil doesn't matter, because the kubeCli will not read from cache

WangXiangUSTC:
just ignore the conflicts now, I will merge this pr to branch dashboard_token when getting two LGTM

pkg/apiserver/common/common.go
Yiyiyimu: we could do `SelectPods(ctx, kubeCli, kubeCli, exp.ParseSelector())`, or when the selector is `fieldSelector` it could not successfully do `List`
https://github.com/chaos-mesh/chaos-mesh/blob/6afb841b02df35a03bbca84419e70cb95eb8c8e1/pkg/utils/selector.go#L140-L150

STRRL: Well, we could not use `SelectPods(ctx, kubeCli, kubeCli, selector)` here, because the implementation of `kubeCli` actually is a `client.DelegatingClient`. `DelegatingClient` contains a cache and require an index(not exist if we manually create one for fieldSelector) while querying with fieldSelector. This issue is considered in https://github.com/chaos-mesh/chaos-mesh/issues/878#issuecomment-688841414.

The reason why you could use this in `chaosctl` is your client is a pure `client.Client` without cache.

Yiyiyimu: My bad. Didn't noticed the change in `SelectPods`. Again thanks for the explanation!

WangXiangUSTC: thank you all for review, I will update this PR after the test passed in branch `dashboard_token_test `, and then welcome to review this pr again. 😃

WangXiangUSTC:
close this pr because we have a new pr https://github.com/chaos-mesh/chaos-mesh/pull/1188 to instead, it contains the codes of front end

Issue #770, opened 2020-07-31
Support DNS Chaos


zhouqiang-cl:
## Feature Request

**Is your feature request related to a problem? Please describe:**

We can support DNS unavailability. It is useful for internet apps, especially when you create your own DNS server

zhouqiang-cl:
There maybe many fault on DNS

- [ ] return wrong IP
- [ ] delay return
- [ ] not resolve domain
- [ ] can not access DNS

PR #1139, opened 2020-11-04
add e2e test for DNS chaos

WangXiangUSTC:

https://github.com/chaos-mesh/chaos-mesh/issues/770

add e2e test for DNS chaos

### Checklist

Tests

- [ ] Unit test
- [x] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


test/config.go
```diff
@@ -40,5 +41,6 @@ func NewDefaultConfig() *Config {
 		E2EImage:          "localhost:5000/pingcap/e2e-helper:latest",
 		ChaosFSImage:      "localhost:5000/pingcap/chaos-fs:latest",
 		ChaosScriptsImage: "localhost:5000/pingcap/chaos-scripts:latest",
+		ChaosDNSImage:     "pingcap/chaosdns:latest",
```
YangKeao, line 44: Though the default value of them takes no effect (as they will be set by the default value of flags), keep a consistency is more tidy.

YangKeao (commented):
Rest LGTM

test/config.go
WangXiangUSTC: do you mean the image should be `localhost:5000/pingcap/chaosdns:latest`?

WangXiangUSTC: done

WangXiangUSTC:
@YangKeao PTAL again

ui/src/slices/experiments.ts
```diff
@@ -19,7 +19,10 @@ export const getStateofExperiments = createAsyncThunk(
   async () => (await api.experiments.state()).data
 )
 
-export const getNamespaces = createAsyncThunk('common/chaos-available-namespaces', async () => (await api.common.chaosAvailableNamespaces()).data)
+export const getNamespaces = createAsyncThunk(
+  'common/chaos-available-namespaces',
+  async () => (await api.common.chaosAvailableNamespaces()).data
+)
```
STRRL, line 25: It seems there is no change in this file.

Could we revert it?

STRRL (changes requested):
rest LGTM

ui/src/slices/experiments.ts
WangXiangUSTC: this should be format by tools in vscode automatically, do you think the update is better? @g1eny0ung

g1eny0ung: Yep. All codes used by the front will be formatted automatically before committing. These changes can be safely ignored.

 And this change should be revert because the latest code is already formatted.

g1eny0ung: Oh, my mistake! I looked at the `dashboard_token_test`  branch before. It's no problem to format this part.

WangXiangUSTC: OK，@ STRRL PTAL again

STRRL (approved):
LGTM

YangKeao (approved):
LGTM

PR #1162, opened 2020-11-09
[add] get event by id

fewdan:
Now the front-end search events need to get all the events, and then filter them according to the ID.
Add an interface to get the event according to the ID, which can omit the step of filtering.

Add an interface which can get event by id.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


pkg/apiserver/event/event.go
```diff
@@ -57,6 +59,7 @@ func Register(r *gin.RouterGroup, s *Service) {
 	// TODO: add more api handlers
 	endpoint.GET("", s.listEvents)
 	endpoint.GET("/dry", s.listDryEvents)
+	endpoint.GET("/get", s.getEvent)
```
g1eny0ung, line 62: I recommend following the restful API, use the format of `/events/id` to get the single event.

g1eny0ung, line 62: I recommend following the restful API, use the format of /events/id to get a single event.

fewdan: ok, i will update it.

g1eny0ung (approved):
LGTM

pkg/apiserver/event/event.go
g1eny0ung: Due to the `/dry` route will conflict with `/:id` and it's no possible to change the `/dry`(Because of compatibility) for now. We will still use `/get` to get a single event.

cwen0: how about `/events/get/:id` ?

pkg/apiserver/event/event.go
```diff
@@ -136,3 +139,42 @@ func (s *Service) listDryEvents(c *gin.Context) {
...
+		c.Status(http.StatusBadRequest)
+		_ = c.Error(utils.ErrInvalidRequest.New("the format of id is wrong"))
+		return
+	}
+
+	event, err := s.event.Find(context.Background(), uint(id))
+	if err != nil {
+		if !gorm.IsRecordNotFoundError(err) {
+			c.Status(http.StatusInternalServerError)
+			_ = c.Error(utils.ErrInternalServer.NewWithNoMessage())
```
cwen0, line 171: Why not returning this error?

pkg/apiserver/event/event.go
cwen0: or `events/detail/:id`?

pkg/apiserver/event/event.go
fewdan: This is consistent with "/archives/detail".
If we want to modify it, I will also change "/archives/detail"?

cwen0 (approved):
LGTM

PR #1169, opened 2020-11-12
cmd/chaosfs: Refact code to fix some bugs.

johncming:

Fix some bugs:
1. Printing the version after checking the flag may result in a failure to print the version.
2. The channel returned by `ctrl.SetupSignalHandler` does not pass a signal type.
3. `pf.Remove()` may lead to a race(remove twice).
4. `exec.CommandContext` uses the same context in retry operation may lead to unexpected behavior (context.DeadlineExceeded).

Fix bugs above.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [X] No code

> There should be tests, but I haven't found them yet. The pr state is set to suspending before the tests are found.

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/cmd/chaosfs/main.go
+++ b/cmd/chaosfs/main.go
@@ -20,17 +20,21 @@ import (
 	"math/rand"
 	"os"
 	"os/exec"
+	"os/signal"
+	"sync"
+	"syscall"
 	"time"
-
-	"github.com/ethercflow/hookfs/hookfs"
 
 	"github.com/chaos-mesh/chaos-mesh/pkg/chaosfs"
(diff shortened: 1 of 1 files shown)


YangKeao:
Sorry, I forgot to remove `chaosfs` from Chaos Mesh 😿 . It's no longer used by IOChaos. The new implementation of the file system has been moved to [toda/hookfs](https://github.com/chaos-mesh/toda/blob/master/src/hookfs/mod.rs)

cwen0:
@johncming We also have the same issue on other components, you can try to fix it.

STRRL:
As chaosfs was deprecated, this PR should be closed.

Issue #886, opened 2020-09-09
Refactor selector


STRRL:
## Feature Request

**Is your feature request related to a problem? Please describe:**

No.

**Describe the feature you'd like:**

Currently, we use `SelectorSpec` to describe which objects we want.
It was impelemented in `pkg/utils/selector.go`, as a util function.

It could NOT handle more complex logic in the future.

We need refactor it, designing a more efficient selector.

**Describe alternatives you've considered:**

**Teachability, Documentation, Adoption, Migration Strategy:**

There might be another problem: 

It will change to a "selector instance" , rather than  a util function. We could using a global singleton instance for a migration solution. 

But we should redesign more things(like many many "Reconciler", even a liitle abuse to this words) to make the dependency tree clean. Might using a DI framework at the right time.

Yisaer:
@Gallardot Thanks for your participation for this issue. As refactoring design need to be discussed first, if you have any ideas, please share the design in the issue or document before starting to code.

Gallardot:
```go
type FilterContext struct {
	ctx context.Context
	client client.Client
	selector v1alpha1.SelectorSpec
	err error
	pods []v1.Pod
	done bool
}

type Filter interface {
	process(selectorContext *FilterContext)
}

type NamespaceFilter struct {
}
func (filter *NamespaceFilter) process(selectorContext *FilterContext)  {
	// TODO: do namespace filter
}

type PodsFilter struct {
}
func (filter *PodsFilter) process(selectorContext *FilterContext)  {
	// TODO: do pods filter
}

type FieldsFilter struct {
}

func (filter *FieldsFilter) process(selectorContext *FilterContext)  {
	// TODO: do fields filter
}

type LabelFilter struct {
}

func (filter *LabelFilter) process(selectorContext *FilterContext)  {
	// TODO: do fields filter
}

func NewFilters() []Filter{
	ret := make([]Filter,0)
	// TODO: add more filters
	ret = append(ret, new(PodsFilter))
	ret = append(ret, new(NamespaceFilter))
	ret = append(ret, new(FieldsFilter))
	ret = append(ret, new(LabelFilter))
	return ret
}

var filters = NewFilters()

// Refactor SelectPods
func SelectPods(ctx context.Context, c client.Client, selector v1alpha1.SelectorSpec) ([]v1.Pod, error) {
	filterCxt := FilterContext{
		ctx: ctx,
		client:c,
		selector:selector,
	}
	for _,filter := range filters{
[...]

PR #1175, opened 2020-11-13
chore: remove common.ControllerCfg reference from utils pkg

### LINKED ISSUE #1072: Make pkg doesn't reference controller gloabl variables (e.g. common.ControllerCfg )

Colstuwjx:
Signed-off-by: colstuwjx <colstuwjx@gmail.com>

Fixed #1072 .

Make pkg/utils doesn't use common.ControllerCfg reference anymore, in order to achieve this, we must pass the controllerCfg in the upstream function calls.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


Colstuwjx:
Ping @STRRL @YangKeao @cwen0 PTAL.

STRRL (changes requested):
Hi @Colstuwjx , thanks a lot!

I find here still a reference to `common.Controller.Cfg` at pkg/router/reconciler.go:46. Could you help us to refactor it? 🥺

Colstuwjx:
> I find here still a reference to common.Controller.Cfg at pkg/router/reconciler.go:46. Could you help us to refactor it? 🥺

Unfortunately, it's not easy to do this. The `Reconcile` method is limited to one parameter, `Reconcile(req ctrl.Request)`, hence, we could not add those cfg as function parameters.

Any suggestions? @STRRL

STRRL:
We could add a parameter to `NewReconciler()`, and make `ClusterScoped` and `TargetNamespace` as fields under struct `Reconciler`. (IMO it should be a state of `Reconciler`)

After refactoring, only `controller` and `cmd/controller-manager` need import `common`; it's a prerequisite for cleanup unnecessary global variable and `init()`. 😁

Colstuwjx:
> We could add a parameter to NewReconciler(), and make ClusterScoped and TargetNamespace as fields under struct Reconciler. (IMO it should be a state of Reconciler)

Sounds a good idea, but there is still an issue [here](https://github.com/chaos-mesh/chaos-mesh/blob/cf9e1ca/pkg/router/route_table.go#L75), if we changed the `NewReconciler` function parameter, how to handle with the `SetupWithManager`?

We've been setup router manager by calling `err = router.SetupWithManager(mgr)` in `cmd/controller-manager`, should we also pass through the `ClusterScoped ` and `TargetNamespace ` as `SetupManager` paramters?

STRRL (approved):
LGTM

pkg/utils/selector.go
```diff
@@ -78,20 +77,20 @@ func SelectAndFilterPods(ctx context.Context, c client.Client, r client.Reader,
 // SelectPods returns the list of pods that are available for pod chaos action.
 // It returns all pods that match the configured label, annotation and namespace selectors.
 // If pods are specifically specified by `selector.Pods`, it just returns the selector.Pods.
-func SelectPods(ctx context.Context, c client.Client, r client.Reader, selector v1alpha1.SelectorSpec) ([]v1.Pod, error) {
+func SelectPods(ctx context.Context, c client.Client, r client.Reader, selector v1alpha1.SelectorSpec, clusterScoped bool, targetNamespace string, allowedNamespaces, ignoredNamespaces string) ([]v1.Pod, error) {
```
Gallardot, line 80: It's kind of ugly, should we rethink how we do management configuration? Maybe it's time for a round of refactoring?

STRRL: Yes. I think we should build a `selector` thing, that contains clients, configurations as its state.

I rememeber we have a talk about this on https://github.com/chaos-mesh/chaos-mesh/issues/886.

Now we hae rfc repo, how about creating a RFC about the selector?

Colstuwjx: Yes, from my view, I think we'd better to call `SelectAndFilterPods`(it will call `SelectPods`) from the common reconciler layer, rather than each chaos endpoint's `Apply` method, therefore, we could consume the controllerCfg and don't need to pass the controller cfg into the util methods.

But it would need a lot of refactoring, so I wonder if we do this job in this PR, or put them into another PR.

fewdan (approved):
LGTM

pkg/utils/selector.go
Yiyiyimu: BTW, current implementation would cause lint to produce waring:
```
linting
~/go/bin/revive -formatter friendly -config revive.toml $(go list ./... | grep -vE "chaos-mesh/test|pkg/ptrace|zz_generated|vendor")
  ⚠  https://revive.run/r#flag-parameter  parameter 'clusterScoped' seems to be a control flag, avoid control coupling
 ~/go/src/github.com/chaos-mesh/pkg/utils/selector.go:80:16

  ⚠  https://revive.run/r#flag-parameter  parameter 'clusterScoped' seems to be a control flag, avoid control coupling
  ~/go/src/github.com/chaos-mesh/pkg/utils/selector.go:80:16

  ⚠  https://revive.run/r#flag-parameter  parameter 'clusterScoped' seems to be a control flag, avoid control coupling
  ~/go/src/github.com/chaos-mesh/pkg/utils/selector.go:80:16

⚠ 3 problems (0 errors, 3 warnings)

Warnings:
  3  flag-parameter
```

Colstuwjx: Yes, a workaround for this warning should be that passing the cfg as the `SelectPods` parameter, but I think the best solution should be that refactoring the `SelectAndFilterPods ` and calling it from the common reconciler layer.

PR #1181, opened 2020-11-17
Add microk8s support for chaos-mesh install.sh script

YangKeao:

The default socket path for microk8s is located at `/var/snap/microk8s/common/run/containerd.sock`, so the users cannot setup a microk8s cluster with default configuration (like k3s).

In this PR, `install.sh` with `--microk8s` argument will set `socketPath` to `/var/snap/microk8s/common/run/containerd.sock`


fewdan (approved):
LGTM

cwen0 (approved):
LGTM

PR #1189, opened 2020-11-19
[add]add unit tests for events

fewdan:

add unit tests for events

add unit tests for events

### Checklist

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


pkg/apiserver/event/event_test.go
```diff
@@ -0,0 +1,349 @@
...
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(rr.Body.Bytes()).Should(Equal(responseBody))
+		})
+
+		It("test err", func() {
+			rr := httptest.NewRecorder()
+			request, err := http.NewRequest(http.MethodGet, "/api/events?uid=err", nil)
+			Expect(err).ShouldNot(HaveOccurred())
+			router.ServeHTTP(rr, request)
+			Expect(err).ShouldNot(HaveOccurred())
```
WangXiangUSTC, line 236: this line can be deleted

pkg/apiserver/event/event_test.go
```diff
@@ -0,0 +1,349 @@
...
+					Duration:     "1m",
+					Pods:         nil,
+					ExperimentID: "testUID",
+				},
+			}
+			rr := httptest.NewRecorder()
+			request, err := http.NewRequest(http.MethodGet, "/api/events/dry?kind=testKind", nil)
+			Expect(err).ShouldNot(HaveOccurred())
+			router.ServeHTTP(rr, request)
+			Expect(err).ShouldNot(HaveOccurred())
```
WangXiangUSTC, line 264: ditto

pkg/apiserver/event/event_test.go
```diff
@@ -0,0 +1,349 @@
...
+			rr := httptest.NewRecorder()
+			request, err := http.NewRequest(http.MethodGet, "/api/events/get?id=1", nil)
+			Expect(err).ShouldNot(HaveOccurred())
+			router.ServeHTTP(rr, request)
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(rr.Code).Should(Equal(http.StatusInternalServerError))
+			Expect(err).ShouldNot(HaveOccurred())
+		})
+
+		It("other err", func() {
```
WangXiangUSTC, line 339: what is the difference between "other err" and the "not found" above?

pkg/apiserver/event/event_test.go
WangXiangUSTC: and have many same problems in other case

pkg/apiserver/event/event_test.go
fewdan: In API processing, if the error is "Gorm recordnotfounderror", the returned results are different, so there are two cases.

pkg/apiserver/event/event_test.go
```diff
@@ -0,0 +1,335 @@
...
+		}
+		router = gin.Default()
+		r := router.Group("/api")
+		endpoint := r.Group("/events")
+		endpoint.GET("", s.listEvents)
+		endpoint.GET("/dry", s.listDryEvents)
+		endpoint.GET("/get", s.getEvent)
+	})
+
+	AfterEach(func() {
```
g1eny0ung, line 188: If there don't a need for this hook, I think it's better to remove it.

pkg/apiserver/event/event_test.go
```diff
@@ -0,0 +1,335 @@
...
+			Expect(rr.Code).Should(Equal(http.StatusOK))
+			responseBody, err := json.Marshal(response)
+			Expect(err).ShouldNot(HaveOccurred())
+			Expect(rr.Body.Bytes()).Should(Equal(responseBody))
+		})
+
+		It("test err", func() {
+			rr := httptest.NewRecorder()
+			request, err := http.NewRequest(http.MethodGet, "/api/events/dry?kind=err", nil)
+			Expect(err).ShouldNot(HaveOccurred())
```
g1eny0ung, line 270: Shoud be `Expect(err).Should(HaveOccurred())`?

g1eny0ung (changes requested):

pkg/apiserver/event/event_test.go
fewdan: This is the err returned by "NewRequest" , which should be nil.

pkg/apiserver/event/event_test.go
fewdan: image

In the other "*_ test.go" , the same method is used, so I moved here directly.

fewdan: It should be "Add any teardown steps that needs to be executed after each test" here, let me change it.

g1eny0ung: Ok. 👌🏽

pkg/apiserver/event/event_test.go
g1eny0ung: 🤔 Know about it. But it seems that all tested requests returned `nil`. Is it necessary to make this step of assertion?

fewdan: OK, I'll take these out.

WangXiangUSTC (approved):
LGTM

g1eny0ung (approved):
LGTM

PR #1193, opened 2020-11-20
chaos daemon: add new function to create daemon server for the physic node

WangXiangUSTC:

chaos-daemon implement almost the same chaos as chaos-mesh, and the code is redundant, we can just import chaos mesh as pkg in chaos-daemon. And we need to do some adapt in chaos-mesh.

add new container client for the physic node so that chaos-daemon can use code in chaos-mesh, you can see https://github.com/chaos-mesh/chaos-daemon/pull/3

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


pkg/bpm/bpm.go
```diff
@@ -237,6 +237,11 @@ func GetNsPath(pid uint32, typ NsType) string {
 
 // SetNS sets the namespace of the process
 func (b *ProcessBuilder) SetNS(pid uint32, typ NsType) *ProcessBuilder {
+	// NOTE: pid is 0 means don't need to enter namespace for this process.
+	if pid == 0 {
```
YangKeao, line 241: I prefer judging outside this function and don't run `SetNS` at all...

WangXiangUSTC: I agree with you, but need to add a lot of judgement, looks ugly 😨 
Do you have any idea about how to do judgment?

STRRL: I think we would never get pid of 0 from user's process. So here is must some logic that explicitly tell you "use pid 0 as this parameter".

WangXiangUSTC: OK, I remove the `nsOptions` from ProcessBuilder and use it when build bec4459
PTAL again

cwen0 (dismissed):
LGTM

WangXiangUSTC:
@YangKeao @STRRL PTAL again

pkg/bpm/bpm.go
YangKeao: It's even worse than the original judgement (at least it doesn't break the `Builder` pattern)... 

You could use `SetNSOpt` directly to pass an option slice. If the slice is empty, then nothing would be done.

WangXiangUSTC:
thanks for the suggestion of @STRRL, now this pr only add a new function 😆 , PTAL again @YangKeao @STRRL

YangKeao (approved):
LGTM

STRRL (approved):
LGTM

PR #1215, opened 2020-11-24
Fix networkchaos/iochaos recover slow

YangKeao:

In `networkchaos`, all `Commit` shares the same `manager`, which will lead to a lot of duplicated commits. For example: if you have 100 pods, 5050 update requests will be sent to Kubernetes, which is really a big problem.

This PR will fix this bug.

BTW: who can teach me how to write elegant codes without tagged enum, without generic, without tuple, and without namespace 👿. The lack of these features has bothered me a lot while developing Chaos Mesh.


controllers/iochaos/podiochaosmanager/podiochaosmanager.go
```diff
@@ -60,8 +70,16 @@ func (m *PodIoManager) WithInit(key types.NamespacedName) *PodIoTransaction {
...
 
+// KeyErrorTuple is a tuple (Key, Err)
+type KeyErrorTuple struct {
+	Key types.NamespacedName
+	Err error
+}
+
 // Commit will update all modifications to the cluster
-func (m *PodIoManager) Commit(ctx context.Context) error {
+func (m *PodIoManager) Commit(ctx context.Context) chan KeyErrorTuple {
```
STRRL, line 80: What's the advantage about using a channel here?

 I think return a `[]error` will be better, it still make sure method `Commit` will pending until all operation finished.

cwen0: If we use chan here, we can add a buffer for it.

YangKeao: @STRRL The way to create a `[]KeyErrorTuple` is just by creating a channel and send the `KeyErrorTuple` into this channel, and the receiver could append these tuples to the slice.

Then why not return the channel directly?

controllers/iochaos/podiochaosmanager/podiochaosmanager.go
```diff
@@ -60,8 +70,16 @@ func (m *PodIoManager) WithInit(key types.NamespacedName) *PodIoTransaction {
 	return t
 }
 
+// KeyErrorTuple is a tuple (Key, Err)
+type KeyErrorTuple struct {
```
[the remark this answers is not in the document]
YangKeao: Good idea.

Actually, it's not a tuple. It's a tagged union, which could be an error with a key or a successful response with a key. The `tuple` is to simulate `tagged union` and the `struct` is used to simulate `tuple`... 

I agree with you that I should name it `CommitResponse` because `tuple` is also hard to understand here.

controllers/iochaos/podiochaosmanager/podiochaosmanager.go
STRRL: For keeping it simple and stupid.

There are 2 reasons that I considered:
- return `[]KeyErrorTuple` means `Commit()` is a synchronized method, we do not need to think about anything about multi-thread. Once channel is used, user should think about "Should I close it?", "Are there any other consumers also read elements from this channel?". Also exposing a channel means here is a possibility that receiver codes could close it(which is dangerous, sending item into a closed channel and closing a closed channel will make panic), and the scope of parallelism is escape from `Commit()` to its caller,  it's a bad-smell about parallelism pattern.
- Using `[]KeyErrorTuple` will not make perform decrease. A for-loop still perform as single-thread.

YangKeao: @STRRL I prefer to make it a read-only channel

@cwen0 Do you have any suggestions on the buffer size? I think blocking at the end of execution is not a problem (as the rest part is all about calculation, without IO operation)

controllers/iochaos/podiochaosmanager/podiochaosmanager.go
```diff
@@ -60,8 +70,16 @@ func (m *PodIoManager) WithInit(key types.NamespacedName) *PodIoTransaction {
...
+	Key types.NamespacedName
```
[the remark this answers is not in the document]
YangKeao: Why should I convert it to string here?

controllers/iochaos/podiochaosmanager/podiochaosmanager.go
STRRL: `Commit(ctx) <-chan` works like "async-commit", since our use-case is quite sample, I think it is acceptable.

PS: If we have many around things need call this, I still suggest provide a "SyncCommit()" 🤪

controllers/networkchaos/trafficcontrol/types.go
```diff
@@ -115,37 +135,40 @@ func (r *endpoint) Apply(ctx context.Context, req ctrl.Request, chaos v1alpha1.I
...
-		return err
-	}
+	ch := m.Commit(ctx)
 
 	networkchaos.Status.Experiment.PodRecords = make([]v1alpha1.PodStatus, 0, len(pods))
-	for index, pod := range pods {
+	for keyErrorTuple := range ch {
+		err := keyErrorTuple.Err
+		if err != nil {
+			// if pod is not found or not running, don't print error log and wait next time.
```
Colstuwjx, line 144: should we also print out one line info or warning log if the pod is not found or not running?

controllers/iochaos/podiochaosmanager/podiochaosmanager.go
Colstuwjx: TBH, I also get confused when I see the `Commit` return is a channel, as @STRRL said, return channel means that it's an async operation, and I think we don't need to return the channel to outside caller, XXXManager calls `m.Commit` and even doesn't have the requirements for async communication, e.g. abort the job.

See the [errgroup test example](https://go.googlesource.com/sync/+/master/errgroup/errgroup_test.go#65), it's similar to this case, and I suggest follow the standard code style.

controllers/networkchaos/trafficcontrol/types.go
YangKeao: I agree with you 👍

STRRL: There are some logs will be printed in the method `Commit()`.

controllers/iochaos/podiochaosmanager/podiochaosmanager.go
YangKeao: @Colstuwjx Nice example! I haven't realized that I can modify a slice if the capacity is big enough 😿 , so no need for a channel.

controllers/networkchaos/trafficcontrol/types.go
Colstuwjx: Yes, It seems that we handled the error twice, I think we'd better do it once, see the [logging best practise from dave cheney](https://dave.cheney.net/2015/11/05/lets-talk-about-logging).

controllers/iochaos/podiochaosmanager/podiochaosmanager.go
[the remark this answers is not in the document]
cwen0: I prefer using a string instead of a struct as a key.

controllers/iochaos/podiochaosmanager/podiochaosmanager.go
YangKeao: Fixed

controllers/iochaos/podiochaosmanager/podiochaosmanager.go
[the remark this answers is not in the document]
YangKeao: Preference is not a strong enough argument :smiling_imp:

controllers/networkchaos/trafficcontrol/types.go
YangKeao: @Colstuwjx Fixed, I have removed redundant logs.

However, nearly all errors are finally "handled" by retrying globally... Maybe we should use `warn` or other levels to log these errors, but it could be a big project to refine error handling :crying_cat_face:

STRRL (approved):
LGTM

cwen0 (approved):
LGTM

Issue #1207, opened 2020-11-23
scheduler: nextRecover shouldn't be  later than nextStart


WangXiangUSTC:
## Bug Report

**What version of Kubernetes are you using?**

git hash 058fb44c997a9fedd18d94f07911565b7b0013e9

**What did you do?**

create a chaos test with `"scheduler": {"cron":"*/1 * * * * "}, "duration": "30s"`

**What did you expect to see?**

run chaos success

**What did you see instead?**

get errors in controller-manager's log:

```log
2020-11-23T10:42:44.962Z        ERROR   controllers.podchaos    Then recover can never be reached.      {
"reconciler": "podchaos", "resource name": "busybox/test3", "scheduler": {"cron":"*/1 * * * * "}, "durati
on": "30s"}
```

WangXiangUSTC:
I add some log, the `now` is `2020-11-23T10:42:44.990Z`, and the next start is `2020-11-23T10:43:00.000Z`, the `next recover` is `2020-11-23T10:43:14.990Z`, so `next start` < `next recover` and return this error

PR #1221, opened 2020-11-25
Easier twophase controller

YangKeao:

#1207 #1214 Make the `twophase` scheduler easier to understand.

> I think the twophase reconciler is too complicated. There are three parts:
> 1. The basic twophase scheduler and retry logic
> 2. Support pause
> 3. Support updating scheduler

I created two controllers for two-phase scenario: 

1. Normal controller supports pause, retry, and basic scheduler
2. A `SchedulerUpdater` controller to update `nextStart` when the `scheduler `changes.

The `SchedulerUpdater` controller is quite easy to understand. I will describe the `Normal` controller design:

1. All modifications on chaos resource and apply/recover chaos should be triggered by an intention to turn into another phase.

2. The intention to turn into another phase could be triggered by time, delete or pause.

3. An intention could fail to set the phase because all side effect actions could fail and turn into the `Failed` phase.


controllers/twophase/state_machine.go
```diff
@@ -0,0 +1,173 @@
...
+
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/chaos-mesh/chaos-mesh/api/v1alpha1"
+	"github.com/chaos-mesh/chaos-mesh/pkg/utils"
+)
+
+type chaosStateMachine struct {
```
WangXiangUSTC, line 29: how about adding unit tests for the `chaosStateMachine `

controllers/twophase/state_machine.go
```diff
@@ -0,0 +1,173 @@
...
+
+		now := time.Now()
+		status.Experiment.EndTime = &metav1.Time{
+			Time: now,
+		}
+		status.Experiment.Duration = now.Sub(status.Experiment.StartTime.Time).String()
+
+		// Pause from a running phase should set the nextStart to now
+		// so that the Reconciler will start another cycle of running right
+		// after resume
```
WangXiangUSTC, line 96: if the `cron` is like `30 * * * *`, it means run chaos on the 30th minute every hour,  can't run right after resume

controllers/twophase/state_machine.go
```diff
@@ -0,0 +1,173 @@
...
+				}
+			}
+
+			m.Chaos.SetNextStart(*nextStart)
+			m.Chaos.SetNextRecover(nextRecover)
+
+			status.Experiment.StartTime = &metav1.Time{Time: startTime}
+			status.Experiment.Duration = duration.String()
+		} else {
+			updated = true
```
WangXiangUSTC, line 143: can set `updated = true` after L104 and remove this line and L106

controllers/twophase/state_machine.go
```diff
@@ -0,0 +1,173 @@
...
+
+			status.Experiment.Phase = targetPhase
+		}
+	}
+
+	if status.FailedMessage != emptyString {
+		updated = true
+	}
+	status.FailedMessage = emptyString
```
WangXiangUSTC, line 152: why need to do this?

controllers/twophase/types.go
```diff
@@ -61,206 +59,64 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-				nextTime = chaos.GetNextRecover()
-			}
-			duration := nextTime.Sub(now)
-			r.Log.Info("Requeue request", "after", duration)
+		targetPhase = v1alpha1.ExperimentPhaseFinished
+	}
 
-			return ctrl.Result{RequeueAfter: duration}, nil
-		}
+	r.Log.Info("decide target phase", "target phase", targetPhase)
```
WangXiangUSTC, line 82: how about adding the chaos's name in all the log of this function `Reconcile `?

controllers/twophase/types.go
```diff
@@ -61,206 +59,64 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
 	chaos := _chaos.(v1alpha1.InnerSchedulerObject)
 
-	duration, err := chaos.GetDuration()
-	if err != nil {
-		r.Log.Error(err, "failed to get chaos duration")
-		return ctrl.Result{}, err
-	}
+	status := chaos.GetStatus()
+
+	targetPhase := status.Experiment.Phase
```
cwen0, line 64: If `status.Experiment.Phase`and `nextStartTime` both are empty, what will happen?

controllers/twophase/state_machine.go
YangKeao: @cwen0 There are 6 * 6 = 36 different branches (though most of them are duplicated). I don't think drawing a clear diagram is possible :crying_cat_face: .

YangKeao: `unit test` has been added

controllers/twophase/state_machine.go
YangKeao: Remove the `failedMessage` from the former execution.

controllers/twophase/state_machine.go
```diff
@@ -0,0 +1,173 @@
...
+	Req   ctrl.Request
+	*Reconciler
+}
+
+func unexpectedInto(currentPhase v1alpha1.ExperimentPhase, targetPhase v1alpha1.ExperimentPhase) error {
+	return errors.Errorf("turn from %s into %s is unexpected", currentPhase, targetPhase)
+}
+
+// This method changes the phase of an object and do some side effects
+// There are 6 different phases, so there could be 6 * 6 = 36 branches
```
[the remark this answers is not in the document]
YangKeao: The empty one "" is also a phase, or should I create an `ExperimentPhaseUnitialized` to equal to `""`?

controllers/twophase/types.go
YangKeao: `targetPhase` will be `Running`, as `chaos.GetNextStart().Before(now)` is true.

controllers/twophase/state_machine.go
```diff
@@ -0,0 +1,173 @@
...
+			startTime := time.Now()
+			nextStart, err := utils.NextTime(*m.Chaos.GetScheduler(), startTime)
+			if err != nil {
+				m.Log.Error(err, "failed to get the next start time")
+				return updated, err
+			}
+			nextRecover := startTime.Add(*duration)
+			// if the duration is too long, `nextRecover` could be after `nextStart`
+			// we can jump over a start to make sure `nextRecover` is before `nextStart`
+			for nextRecover.After(*nextStart) {
```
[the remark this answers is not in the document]
YangKeao: ? It will only loop 60 times and get a correct time. How could it block the goroutine?

controllers/twophase/state_machine.go
```diff
@@ -0,0 +1,173 @@
...
+			}
+			status.Experiment.Phase = targetPhase
+
+			startTime := time.Now()
+			nextStart, err := utils.NextTime(*m.Chaos.GetScheduler(), startTime)
+			if err != nil {
+				m.Log.Error(err, "failed to get the next start time")
+				return updated, err
+			}
+			nextRecover := startTime.Add(*duration)
```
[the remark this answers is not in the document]
YangKeao: The risk of renaming schema is too high. I would like to try to do it in another PR.

controllers/twophase/state_machine.go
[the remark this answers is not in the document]
cwen0: Yes, `ExperimentPhaseUnitialized` is OK.

controllers/twophase/state_machine.go
YangKeao: @cwen0 I don't know the exact definition of `pause`. It seems that the `pause` has two different meanings for these two kinds of scheduler.

Maybe we should clarify it first, through an RFC or an issue... I could modify the implementation according to that later (in another PR). Confusing feature will lead to confusing codes, which we always hate.

controllers/twophase/state_machine.go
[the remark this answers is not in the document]
cwen0: Yes, my fault, this example will loop 60 times. If the duration is too long, such as 1 year, and the `corn` is very short, such as 10s. this function will be blocked very long.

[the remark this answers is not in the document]
YangKeao: @cwen0 I can not even binary search here, as the scheduler could only iterate once. Do you have any suggestion?

cwen0 (approved):
LGTM

controllers/twophase/types.go
YangKeao: added in the name

controllers/twophase/state_machine.go
Yiyiyimu: I don't think `pause` has two meanings for two kinds of schedulers. For the current implementation, in both situations, if we resume at the waiting state of the scheduler, we would turn the state from paused to waiting (at "Recovering"), get the next start time from the scheduler  ("Waiting"), wait for the next start time and go into "Starting". 
So it's like, the scheduler keeps the same, and the pause period is dug out of it.

For #1207, I checked [the original implementation](https://github.com/chaos-mesh/chaos-mesh/pull/15/commits/dd87ad8600ebac800fae8f54fc43aa3ef24ed78d). Maybe I missed something, but I think change nextRecover from `now + duration` to `nextStart + duration` could fix it.

controllers/twophase/state_machine.go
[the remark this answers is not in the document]
Yiyiyimu: Could we directly set `nextStart, err = utils.NextTime(*m.Chaos.GetScheduler(), nextRecover)`, since the condition fulfill the current one would make nextStart the first start time after nextRecover

controllers/twophase/state_machine.go
```diff
@@ -0,0 +1,186 @@
...
+
+			counter := 0
+			// if the duration is too long, `nextRecover` could be after `nextStart`
+			// we can jump over a start to make sure `nextRecover` is before `nextStart`
```
Yiyiyimu, line 135: Hi @YangKeao I have another idea to achieve maybe an ideal situation, that for standard cron (* * * * *), we could follow the schedule in this cycle (which we missed the start time), rather than start at the next cycle (or the cycle after the next).

So there are two situation: we entered in **running** period, or **waiting** period. We could calculate the whole cron cycle duration, and get the `waitDuration` from `cycle - duration`, which would be helpful in the following.

1. Running period
When `time.Now().Before(nextStart - waitDuration)`, we could see we are in the running period. Entering state machine, still in this 20 conditions, but nextRecover should be `nextStart - waitTime`. And keep the same for others.

2. Waiting period
The same, when `time.Now().After(nextStart - waitDuration)`. This time we should separate a condition, from `Uninitialized` to `Waiting`, and the only thing to do is set next start. After that it could be the same with the `every` type cron.

STRRL: `cycle` is not a stable value, for example: `5/15 * * * *`, it means execute at each `<any hour>:05` and `<any hour>:15`.

Yiyiyimu: @STRRL Yes you're right I missed this point. Maybe we could implement a `Previous`, just like `Next` but get the previous start time from certain time, which would also need when implementing [pause for duration](https://github.com/chaos-mesh/chaos-mesh/pull/1016/files#diff-04c792dfed57630c1de983400da2e7639c6d5c3a929c0124f7b5c87aa9a19105R121-R124).
Actually someone creates a [PR about this feature](https://github.com/robfig/cron/pull/216) in cron repo, but the maintainer thinks it's outside the scope of that repo so we need to do this ourselves.

Yiyiyimu: > ~`cycle` is not a stable value, for example: `5/15 * * * *`, it means execute at each `<any hour>:05` and `<any hour>:15`.~
> 
> That's not an issue. I still considering other unexpected situations.

Yeah but something like `5/18 * * * *` would still cause an unstable cycle (run at 5, 23, 41, 59 in each hour)

STRRL:
I could not find where the first call of `Chaos.SetNextStart()`;

- When the Chaos Object just creates, its Status is empty, `GetNextStart()` will return an empty `time.Time{}`, and phase is 
`ExperimentPhaseUnitialized`
- And `state machine` only update this field when state turn to `Running`; [controllers/twophase/state_machine.go:150-151](https://github.com/YangKeao/chaos-mesh/blob/5e15b0bdadcd8794261607570776776f24009689/controllers/twophase/state_machine.go#L150-L151)
- The condition to turn into `Running` is `chaos.GetNextStart().Before(now)`[controllers/twophase/types.go:72-74](https://github.com/YangKeao/chaos-mesh/blob/5e15b0bdadcd8794261607570776776f24009689/controllers/twophase/types.go#L72-L74); But I think when the object just commit to Kubernetes, `chaos.GetNextStart()` returns an empty `time.Time{}`, so that the two-phase will not start.

Did I miss something? 😰

controllers/twophase/state_machine.go
YangKeao: @Yiyiyimu I have come up with an idea: if the current phase is `Paused`, we should ignore the `targetPhase`. And loop and iterate forward the `nextStart` and `nextRecover` until at least one of them is bigger than `now`.

The behavior of `@every 5min` is different as it may not apply right after resume. However, it works like erasing a period of time.

Yiyiyimu: @YangKeao Yes it could work for `pause`, but that still could not solve the problem when we just start, since we don't know the last start time.

controllers/twophase/state_machine.go
```diff
@@ -0,0 +1,186 @@
...
+	Req   ctrl.Request
+	*Reconciler
+}
+
+func unexpectedInto(currentPhase v1alpha1.ExperimentPhase, targetPhase v1alpha1.ExperimentPhase) error {
+	return errors.Errorf("turn from %s into %s is unexpected", currentPhase, targetPhase)
+}
+
+// This method changes the phase of an object and do some side effects
+// There are 6 different phases, so there could be 6 * 6 = 36 branches
```
Colstuwjx, line 42: For which state could trans to the target state or not, this method seems still not clear to me.

Look at some well-designed fsm repo, e.g. [this one](https://github.com/looplab/fsm), it will setup fsm rules **explicitly**. In our scenario, `currentPhase` is the fsm state, and `targetPhase` is fsm target state, how about write down these 6*6 branches as a rule map, and use `canTrans` such grammar to check if the currentState could trans to the target state and do the job.

If we choose this implementation, we could remove these if-else statements, and it should be more clearer to our developers.

YangKeao: `map` is a good idea 👍

controllers/twophase/state_machine_test.go
```diff
@@ -0,0 +1,219 @@
...
+			v1alpha1.ExperimentPhasePaused,
+			v1alpha1.ExperimentPhaseRunning,
+			v1alpha1.ExperimentPhaseWaiting,
+		}
+
+		It("StateMachine Target Finish", func() {
+			defer mock.With("MockApplyError", errors.New("ApplyError"))()
+			defer mock.With("MockRecoverError", errors.New("RecoverError"))()
+
+			for _, status := range status {
```
Colstuwjx, line 59: for status range status ? 
```suggestion
			for _, status := range statuses {
```

controllers/twophase/state_machine_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+	RunSpecsWithDefaultAndCustomReporters(t,
+		"Twophase StateMachine Suite",
+		[]Reporter{envtest.NewlineReporter{}})
+}
+
+var _ = Describe("TwoPhase StateMachine", func() {
+	Context("TwoPhase", func() {
+
+		status := []v1alpha1.ExperimentPhase{
```
Colstuwjx, line 47:
```suggestion
		statuses := []v1alpha1.ExperimentPhase{
```

controllers/twophase/state_machine_test.go
```diff
@@ -0,0 +1,219 @@
...
+			defer mock.With("MockApplyError", errors.New("ApplyError"))()
+			defer mock.With("MockRecoverError", errors.New("RecoverError"))()
+
+			for _, status := range status {
+				now := time.Now()
+				sm := setupStateMachineWithStatus(status)
+
+				updated, err := sm.run(context.TODO(), v1alpha1.ExperimentPhaseFinished, now)
+
+				if status == v1alpha1.ExperimentPhaseRunning {
```
Colstuwjx, line 65: If we accept that retry Recover while `ExperimentPhaseFailed ` -> `ExperimentPhaseFinished ` and `ExperimentPhaseFinished ` -> `ExperimentPhaseFinished `, we should update this test case accordingly.

Colstuwjx (changes requested):
rest LGTM.

STRRL:
Hi @YangKeao, the "state machine" could handle "What to do" when states changing👍. But the conditions for "When to change" is still out of the state machine(now it is in line 68 - 82 in controllers/twophase/update_scheduler.go). 

Should we move it into the state machine?

And how about changing the `none` to `noop`(means no-operation).

controllers/twophase/state_machine.go
```diff
@@ -0,0 +1,296 @@
...
+		v1alpha1.ExperimentPhaseFailed:        unexpected,
+		v1alpha1.ExperimentPhaseFinished:      none,
+	},
+	v1alpha1.ExperimentPhaseFailed: {
+		v1alpha1.ExperimentPhaseUninitialized: unexpected,
+		v1alpha1.ExperimentPhaseRunning:       apply,
+		v1alpha1.ExperimentPhaseWaiting:       none,
+		v1alpha1.ExperimentPhasePaused:        none,
+		v1alpha1.ExperimentPhaseFailed:        none,
+		v1alpha1.ExperimentPhaseFinished:      none,
```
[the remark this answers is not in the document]
YangKeao: Actually `ExperimentPhaseFailed` means failing to apply. Failing to recover will keep the experiment in `ExperimentPhaseRunning`.

Maybe I should modify the name. I don't know whether this modification would break the dashboard, so I will try to modify it in another PR.

controllers/twophase/state_machine_test.go
YangKeao: Fixed

controllers/twophase/state_machine.go
YangKeao: @Yiyiyimu I have implemented the logic mentioned above. Could you have a look?

controllers/twophase/state_machine.go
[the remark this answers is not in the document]
Colstuwjx: OK 😁

controllers/twophase/state_machine.go
```diff
@@ -0,0 +1,296 @@
...
+	},
+	v1alpha1.ExperimentPhaseFinished: {
+		v1alpha1.ExperimentPhaseUninitialized: unexpected,
+		v1alpha1.ExperimentPhaseRunning:       unexpected,
+		v1alpha1.ExperimentPhaseWaiting:       unexpected,
+		v1alpha1.ExperimentPhasePaused:        unexpected,
+		v1alpha1.ExperimentPhaseFailed:        unexpected,
+		v1alpha1.ExperimentPhaseFinished:      none,
```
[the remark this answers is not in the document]
YangKeao: As described in https://github.com/chaos-mesh/chaos-mesh/pull/1221#discussion_r539010952 , we shouldn't retry here. Because the experiment will keep in `ExperimentPhaseRunning` but will not step into `ExperimentPhaseFinished` if recovering failed.

cwen0:
> Hi @YangKeao, the "state machine" could handle "What to do" when states changing👍. But the conditions for "When to change" is still out of the state machine(now it is in line 68 - 82 in controllers/twophase/update_scheduler.go).
> 
> Should we move it into the state machine?
> 
> And how about changing the `none` to `noop`(means no-operation).

@STRRL From the definition of this state machine, the input and output both are the phases of the experiment. So I think `When to change` just generates the input phase and target phase, this logic should not be defined in state mechine.

Colstuwjx (dismissed):
> @STRRL From the definition of this state machine, the input and output both are the phases of the experiment. So I think When to change just generates the input phase and target phase, this logic should not be defined in the state machine.

From my view, a well-designed state machine should contains `all available states`, `events` ( which state could trans to which state by this event ) and `event callbacks` ( in our scenario, it called the side-effect ). I think our `chaosStateMachine` actually did all these things.

The difference is that, the `chaosStateMachine` accept `currentPhase` and `targetPhase`, rather than `currentPhase` and `event`( or so-called action),  we decide the targetPhase before initializing state machine, and just pass the targetPhase into the state machine, let it judge and do state transmission.

IMO, I think the `chaosStateMachine` implementation has been more cleared than the previous commit version, and it maybe hard to define the real events from our case ( [these if-else statements check](https://github.com/chaos-mesh/chaos-mesh/pull/1221/files#diff-04c792dfed57630c1de983400da2e7639c6d5c3a929c0124f7b5c87aa9a19105R68-R80) can't be replaced while decide the target phase, right? ).

Colstuwjx:
LGTM

cwen0 (dismissed):

YangKeao:
Sorry. It seems that the `PAUSE` still has some bugs :(

Colstuwjx:
> Sorry. It seems that the PAUSE still has some bugs :(

Confirmed, I have been tested the experiment on my laptop, even I changed the annotation to false:

```
kind: PodChaos
metadata:
  name: pod-kill-example
  namespace: default
  annotations:
    "experiment.chaos-mesh.org/pause": "false"
spec:
  action: pod-kill
  mode: one
  selector:
    namespaces:
      - default
    labelSelectors:
      "app": "web-show" # the label of the pod for chaos injection
  duration: "9m" # duration for the injected chaos experiment
  scheduler:
    cron: '@every 10m'
```

The controller still keep `Paused`:

```
2020-12-10T05:49:30.845Z	INFO	controllers.podchaos.default/pod-kill-example	decide target phase	{"reconciler": "podchaos", "resource name": "default/pod-kill-example", "target phase": "Paused"}
2020-12-10T05:49:30.845Z	INFO	controllers.podchaos.default/pod-kill-example	change phase	{"reconciler": "podchaos", "resource name": "default/pod-kill-example", "current phase": "Paused", "target phase": "Paused"}
2020-12-10T05:49:30.845Z	INFO	controllers.podchaos.default/pod-kill-example	requeue	{"reconciler": "podchaos", "resource name": "default/pod-kill-example", "requeue after": "3m28.154828018s"}
```

controllers/twophase/types.go
```diff
@@ -61,206 +61,73 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-		chaos.SetNextRecover(time.Time{})
-
-		status.Experiment.EndTime = &metav1.Time{
-			Time: time.Now(),
-		}
-		status.Experiment.Phase = v1alpha1.ExperimentPhaseWaiting
-		status.FailedMessage = emptyString
-	} else if (status.Experiment.Phase == v1alpha1.ExperimentPhaseFailed ||
-		status.Experiment.Phase == v1alpha1.ExperimentPhasePaused) &&
-		!chaos.GetNextRecover().IsZero() && chaos.GetNextRecover().After(now) {
```
Colstuwjx, line 138: @YangKeao We've been missing this part check, we're still in this current round!

YangKeao: Yes. It failed to move forward from a `PhasePause`... `targetPhase` was initialized as `PhasePause`, and it doesn't have enough evidence to move forward to another phase.

I have added an ugly check to decide whether it's resuming from pause.

```
if !chaos.IsPaused() && status.Experiment.Phase == v1alpha1.ExperimentPhasePaused {
	// Running and Waiting has the same logic for resuming
	targetPhase = v1alpha1.ExperimentPhaseRunning
}
```

Actually, the state machine has a lot more input than `Phase`. The `nextStart`, `nextRecover` and `lastStart` are all the inputs to resolve the pause problem... I failed to design a clean mechanism 😿 .

Colstuwjx: The ideal state-machine implementation should be an event-driven model, rather than deciding the target phase according these `nextStart`, `nextRecover`, `lastStart `...

The  root cause why we need these if-else check, I think it is that we've been **mixed** schedule and state machine input event into the same one controller, so we cared about these chaos current state, and according these state to decide the target state.

controllers/twophase/types.go
```diff
@@ -61,206 +61,80 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	if !chaos.GetNextRecover().IsZero() && chaos.GetNextRecover().Before(now) {
+		targetPhase = v1alpha1.ExperimentPhaseWaiting
 	}
 
-	scheduler := chaos.GetScheduler()
-	if scheduler == nil {
-		r.Log.Info("Scheduler should be defined currently")
-		return ctrl.Result{}, fmt.Errorf("misdefined scheduler")
+	if chaos.GetNextStart().Before(now) {
+		targetPhase = v1alpha1.ExperimentPhaseRunning
```
Yiyiyimu, line 73: The logic here seems not correct, if we use the cron that does not start right away, for example, if I want to start a chaos at 00:00 each day for 10 minutes. I still think a `Previous` function, to get `lastStart` and `lastRecover` could be useful in this case. But since it's not the purpose of this PR, we could add this in another PR

YangKeao: Yes. It's still a problem 🤦‍♂️ .

cwen0: @Yiyiyimu This is a bug here, Can you help to create an issue about this bug?

Yiyiyimu: @cwen0 Sure, created as #1270

cwen0 (dismissed):

controllers/twophase/types.go
```diff
@@ -61,206 +61,80 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-		zero := 0 * time.Second
-		duration = &zero
+	if chaos.IsPaused() {
+		targetPhase = v1alpha1.ExperimentPhasePaused
 	}
 
-	status := chaos.GetStatus()
+	// TODO: find a better way to solve the pause and resume problem.
+	// Or pause is a bad design for the scheduler :(
+	if !chaos.IsPaused() && status.Experiment.Phase == v1alpha1.ExperimentPhasePaused {
```
Colstuwjx, line 82: Little confusion here.

At previous commit, each if-else, e.g. `!chaos.GetNextRecover().IsZero() && chaos.GetNextRecover().Before(now)` and `!chaos.IsPaused() && status.Experiment.Phase == v1alpha1.ExperimentPhasePaused` wouldn't run at the same time, but now, it did.

Think about the following case:

1. a chaos created and started;
2. it has been paused before nextRecover reached;
3. nextRecover reached but nothing changed since the chaos is still been `Paused` :(
4. the chaos has been unpaused

and now, both `!chaos.GetNextRecover().IsZero() && chaos.GetNextRecover().Before(now)` and `!chaos.IsPaused() && status.Experiment.Phase == v1alpha1.ExperimentPhasePaused` would be true, the targetPhase become `ExperimentPhaseRunning`!

Although it would not make the chaos really run `apply`, because [Line 149] if check would be false, and the targetPhase would be changed to `ExperimentPhaseWaiting` after did `resume`.

Is this acceptable?

controllers/twophase/types.go
```diff
@@ -61,206 +61,80 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	if nextRecoverAfter > 0 && requeueAfter > nextRecoverAfter {
+		requeueAfter = nextRecoverAfter
 	}
+	if nextStartAfter > 0 && requeueAfter > nextStartAfter {
+		requeueAfter = nextStartAfter
+	}
+
+	var err error
+	if requeueAfter == math.MaxInt64 {
+		err = errors.Errorf("unexpected behavior, now is greater than nextRecover and nextStart")
```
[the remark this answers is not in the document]
cwen0: Good catch, I think this is a bug!

controllers/twophase/types.go
cwen0: This implement keeps unifying with the current version, you can check here https://github.com/chaos-mesh/chaos-mesh/blob/master/controllers/twophase/types.go#L138. I think this implement is acceptable at present. If we want to improve this, we can do it in another pr.

controllers/twophase/state_machine.go
```diff
@@ -0,0 +1,288 @@
...
+			m.Log.Error(err, "failed to get the next start time")
+			return nil, nil, err
+		}
+
+		counter++
+		if counter > iterMax {
+			err = errors.Errorf("the number of iterations exceeded with nextRecover(%s) nextStart(%s)", nextRecover, nextStart)
+			return nil, nil, err
+		}
+	}
```
Yiyiyimu, line 234: I have a question here. Since we are trying to find the first start after `nextRecover`, and `nextRecover` is not changed in the loop, I think we could directly do `nextStart = utils.NextTime(*scheduler, nextRecover)` without iteration

controllers/twophase/types.go
Colstuwjx: I think it's not the same case, what you mean in L138 should be `started -> paused -> now -> nextRecover -> nextStart`, and when I unpaused at **now**, it do retrying.

But what I mean is that `started -> paused -> nextRecover -> now -> nextStart`, and when I unpaused at **now**, it will enter both [Line 68](https://github.com/chaos-mesh/chaos-mesh/pull/1221/files#diff-04c792dfed57630c1de983400da2e7639c6d5c3a929c0124f7b5c87aa9a19105R68) and [Line 82](https://github.com/chaos-mesh/chaos-mesh/pull/1221/files#diff-04c792dfed57630c1de983400da2e7639c6d5c3a929c0124f7b5c87aa9a19105R72), and targetPhase become `ExperimentPhaseRunning` at this moment 😅

And then,  it enter `resume()` [Line 156](https://github.com/chaos-mesh/chaos-mesh/pull/1221/files#diff-d396848e41155ba5b591cf171a8ab550845161333a956422f14c840b611d16bcR156).

It's not a big problem since it's finally do `noop`, but it still will leave one line log say:

```
2020-12-11T10:25:24.033Z	INFO	controllers.networkchaos.default/web-show-network-delay	change phase	{"reconciler": "networkchaos", "resource name": "default/web-show-network-delay", "current phase": "Paused", "target phase": "Running"}
```

and then turn to `Waiting` soon:

```
2020-12-11T10:25:24.051Z	INFO	controllers.networkchaos.default/web-show-network-delay	change phase	{"reconciler": "networkchaos", "resource name": "default/web-show-network-delay", "current phase": "Waiting", "target phase": "Waiting"}
```

controllers/twophase/state_machine.go
YangKeao: For a `cron` scheduler, `nextStart = utils.NextTime(*scheduler, nextRecover)` is enough. For `@every` scheduler, the answer is not correct :cry: 

The two kinds of scheduler is a disaster to the implementation :sob: .

Yiyiyimu: TLDR: I think the [loop in `resume`](https://github.com/chaos-mesh/chaos-mesh/pull/1221/files#diff-d396848e41155ba5b591cf171a8ab550845161333a956422f14c840b611d16bcR148) is essential but not this one

I made some categorized discussion (Chinglish! 🙈) for `@every` scheduler, and it seems it could work in this situation. I'm not sure if there's some miscalculation in my discussion...

There are two situations we need to use `IterateNextStart`: `apply` and `resume`. And to make it clear, let's say `NEXTSTART` is the real nextStart we want, same with `NEXTRECOVER`.
1. For [`apply`](https://github.com/chaos-mesh/chaos-mesh/pull/1221/files#diff-d396848e41155ba5b591cf171a8ab550845161333a956422f14c840b611d16bcR53), 
1.1. normally `startTime = now`, `nextStart = NEXTSTART`, `nextRecover = now + duration = NEXTRECOVER`. So would not enter this loop and returns. 
2.2. only when `duration > cycle time` should we iterate `nextStart` to the next one after `nextRecover`, in this case, `nextStart = utils.NextTime(*scheduler, nextRecover)` could works well. We will ignore this case for following discussion to keep it simple.
2. For [`resume`](https://github.com/chaos-mesh/chaos-mesh/pull/1221/files#diff-d396848e41155ba5b591cf171a8ab550845161333a956422f14c840b611d16bcR117), let's say a cycle = (first start,  second start]
2.1. when `nextRecover/nextStart` is in the same cycle with the resume time: one of them is after `now` so it would not enter `IterateNextStart` and return
2.2. when `nextRecover/nextStart` is sereval loops before us, if we pause for a long time. It would enter this loop, find the first `nextStart` after `nextRecover` and returns, and keep iterating in loop in `resume`. Since when we resume, `nextRecover/nextStart` are in the same loop, only apply `nextStart, err = utils.NextTime(*scheduler, *nextStart)` one time could be enough. So also, `nextStart = utils.NextTime(*scheduler, nextRecover)` could works.
2.3.when  `nextRecover/nextStart` is in the last cycle of resume time, after we iterate several 2.2 in `resume`. Go into `IterateNextStart`, we get `nextStart = NEXTSTART`, `nextRecover = NEXTRECOVER`, go into 2.1. and finish. No need to enter this loop and return.

So there is two situations (1.2 and 2.2) we need this loop and both of them could be replaced in this case.

YangKeao: For example:

```
now := "12:10"
scheduler := "@every 5m"
duration := "6m"
```

Then we expected `nextRecover = "12:16"` and `nextStart = "12:20"`, but if we use `nextStart = utils.NextTime(*scheduler, nextRecover)`, it will be `12:21`.

controllers/twophase/types.go
YangKeao: Yes. It's the dirty part. I have commented that target `Running` and `Recovering` has the same behavior for `resume`. I haven't come up with a good idea on how to solve the problem 😭

controllers/twophase/state_machine.go
Yiyiyimu: You're right I misunderstood what `scheduler.Next()` do for `@every`. Thanks for the explanation!

controllers/twophase/types.go
STRRL: I always think this is a feature that you could see the chaos experiments being resumed IMMEDIATELY after clicking unpause.

Colstuwjx: If we did think so, how about make this targetPhase to a temporary phase, e.g. v1alpha1.ExperimentPhaseResuming   😁

YangKeao: @STRRL This phase will not be updated to `Kubernetes`. As you can see in the state machine, though we passed `targetPhase` as `Running`, it will be updated to `Waiting` (or `Failed`). It can only be seen in the log.

controllers/twophase/types.go
```diff
@@ -61,221 +61,80 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
-	r.Log.Info("Chaos action:", "chaos", chaos)
-
-	// Start to apply action
-	r.Log.Info("Performing Action")
-
-	if err := r.Apply(ctx, req, chaos); err != nil {
-		r.Log.Error(err, "failed to apply chaos action")
-
-		status.Experiment.Phase = v1alpha1.ExperimentPhaseFailed
+	if chaos.GetStatus().Experiment.Phase == v1alpha1.ExperimentPhaseFinished {
```
Colstuwjx, line 104: In the previous commit version, if it's from Paused to Paused, our controller would only run into [Line 114](https://github.com/chaos-mesh/chaos-mesh/pull/1221/files#diff-04c792dfed57630c1de983400da2e7639c6d5c3a929c0124f7b5c87aa9a19105L114), and don't re-enqueue it.

To fix the `Paused->Paused` case, how about make the changes like below:

```suggestion
	if chaos.GetStatus().Experiment.Phase == v1alpha1.ExperimentPhaseFinished || chaos.GetStatus().Experiment.Phase == v1alpha1.ExperimentPhasePaused {
```

YangKeao: Good Idea :+1:

controllers/twophase/types.go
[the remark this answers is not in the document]
Colstuwjx: I've been confirmed that [the last commit](https://github.com/chaos-mesh/chaos-mesh/pull/1221/commits/a7a8fa02a2498fc79f8573d3482f473e404267cd) did fix this.

Colstuwjx:
Merging this PR would make two benifits:

1. easier to change business codes for two-phase controller in the future;

2. fixed #1207 

And I think we could leave some existing issues to another PRs to fix.

Colstuwjx (approved):
LGTM

cwen0 (approved):

YangKeao:
The bot doesn't work 👿

PR #1242, opened 2020-12-03
chaos-daemon: support setting a specific network device

cwen0:

Support setting a specific network device. 

Add a new field in the `TcsRequest` struct. By default, we request that all TC rules in one`TcsRequest` are applied to one device.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


pkg/chaosdaemon/tc_server.go
```diff
@@ -76,7 +76,7 @@ func (s *daemonServer) SetTcs(ctx context.Context, in *pb.TcsRequest) (*empty.Em
 	}
 
 	client := buildTcClient(ctx, pid)
-	err = client.flush()
+	err = client.flush(in.Device)
```
WangXiangUSTC, line 79: how about setting the `in.Device` to `DefaultDevice` while it is empty. User  may use an older controller manager and without setting the `Device`

cwen0: Fix

WangXiangUSTC (approved):
LGTM

YangKeao (approved):
LGTM

PR #1249, opened 2020-12-04
dashboard: support set config to disable token

WangXiangUSTC:

dashboard support set config to disable token

add a config for disabling token, if it is true, will use the local client.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/helm/chaos-mesh/templates/chaos-dashboard-deployment.yaml
+++ b/helm/chaos-mesh/templates/chaos-dashboard-deployment.yaml
@@ -52,6 +52,8 @@ spec:
               value: {{ .Values.controllerManager.targetNamespace }}
             - name: CLUSTER_SCOPED
               value: "{{ .Values.clusterScoped }}"
+            - name: DISABLE_TOKEN
+              value: "{{ .Values.disableToken }}"
             {{- if .Values.controllerManager.allowedNamespaces }}
             - name: ALLOWED_NAMESPACES
               value: {{ .Values.controllerManager.allowedNamespaces }}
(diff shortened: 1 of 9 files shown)


pkg/clientpool/client.go
```diff
@@ -28,33 +28,64 @@ import (
...
+}
+
+func NewLocalClient(client pkgclient.Client) Clients {
+	return &LocalClient{
+		client: client,
+	}
+}
+
+// Client returns the local k8s client
+func (c *LocalClient) Client(token string) (pkgclient.Client, error) {
```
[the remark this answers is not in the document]
WangXiangUSTC: just used to implement the `Clients` interface

g1eny0ung (approved):
LGTM

helm/chaos-mesh/templates/chaos-dashboard-deployment.yaml
```diff
@@ -52,6 +52,8 @@ spec:
               value: {{ .Values.controllerManager.targetNamespace }}
             - name: CLUSTER_SCOPED
               value: "{{ .Values.clusterScoped }}"
+            - name: UNSAFE_MODE
```
zhouqiang-cl, line 55: How about use security_mode

zhouqiang-cl:
And I think we should also need to add an option in `install.sh` and in documents

fewdan (dismissed):
LGTM

helm/chaos-mesh/templates/chaos-dashboard-deployment.yaml
WangXiangUSTC: done

WangXiangUSTC:
> And I think we should also need to add an option in `install.sh` and in documents

I update securityMode=false in `install.sh`, PTAL again @zhouqiang-cl

zhouqiang-cl (approved):
LGTM

g1eny0ung (approved):

PR #1252, opened 2020-12-05
chore: remove duplicate 'GenerateNPods'

Yiyiyimu:

There are two `GenerateNPods` exists with little difference, so merge them together
https://github.com/chaos-mesh/chaos-mesh/blob/26fe5523a8ff9fdc3d934c43e5a0cd8ec50aec48/pkg/utils/selector_test.go#L592
https://github.com/chaos-mesh/chaos-mesh/blob/6afb841b02df35a03bbca84419e70cb95eb8c8e1/controllers/test/types.go#L128

Add a struct so don't need to type in the arguments not used in certain test.

### Checklist

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/pkg/utils/generate.go
+++ b/pkg/utils/generate.go
@@ -1,4 +1,4 @@
-// Copyright 2019 Chaos Mesh Authors.
+// Copyright 2020 Chaos Mesh Authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/pkg/utils/generate.go
+++ b/pkg/utils/generate.go
@@ -25,8 +25,8 @@ import (
 // For the others, the default values are empty.
 type PodArg struct {
 	Name            string
-	Status          v1.PodPhase `default:"corev1.PodRunning"`
-	Namespace       string      `default:"metav1.NamespaceDefault"`
+	Status          v1.PodPhase
+	Namespace       string
 	Ans             map[string]string
 	Labels          map[string]string
 	ContainerStatus v1.ContainerStatus
--- a/pkg/utils/generate.go
+++ b/pkg/utils/generate.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+// PodArg by DEFAULT use `Status=corev1.PodRunning` and `Namespace=metav1.NamespaceDefault`, when they are not specified
 type PodArg struct {
 	Name            string
 	Status          v1.PodPhase
--- a/pkg/utils/generate.go
+++ b/pkg/utils/generate.go
@@ -24,8 +24,8 @@ import (
 // PodArg by DEFAULT use `Status=corev1.PodRunning` and `Namespace=metav1.NamespaceDefault`, when they are not specified
 type PodArg struct {
 	Name            string
-	Status          v1.PodPhase
-	Namespace       string
+	Status          v1.PodPhase `default: corev1.PodRunning`
+	Namespace       string      `default: metav1.NamespaceDefault`
 	Ans             map[string]string
 	Labels          map[string]string
 	ContainerStatus v1.ContainerStatus
--- a/pkg/utils/generate.go
+++ b/pkg/utils/generate.go
@@ -25,8 +25,8 @@ import (
 // For the others, the default values are empty.
 type PodArg struct {
 	Name            string
-	Status          v1.PodPhase `default: corev1.PodRunning`
-	Namespace       string      `default: metav1.NamespaceDefault`
+	Status          v1.PodPhase `default:"corev1.PodRunning"`
+	Namespace       string      `default:"metav1.NamespaceDefault"`
 	Ans             map[string]string
 	Labels          map[string]string
 	ContainerStatus v1.ContainerStatus
--- a/pkg/utils/generate.go
+++ b/pkg/utils/generate.go
@@ -1,0 +1,109 @@
+// Copyright 2019 Chaos Mesh Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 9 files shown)


STRRL (changes requested):
rest LGTM.

Since we are build test cases, explicitly environment is very important, so using default values is not so good for the situation.

And suggestions for default value: do not apply default value as `if value == "" { value = defaultValue }`; using a constructor to setup default value;

pkg/utils/generate.go
```diff
@@ -0,0 +1,109 @@
...
+	Nodename        string
+}
+
+func newPod(p PodArg) v1.Pod {
+	if p.Status == "" {
+		p.Status = v1.PodRunning
+	}
+	if p.Namespace == "" {
+		p.Namespace = metav1.NamespaceDefault
+	}
```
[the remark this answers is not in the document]
Yiyiyimu: I think we could use default value for values that we don't care in certain test. 
When we need them, for example
https://github.com/chaos-mesh/chaos-mesh/blob/e63c877a5503b990a48b2c239847047677cba94e/pkg/utils/selector_test.go#L333-L336
When we explicitly compare pod status, we could still write them out to clarify what we test.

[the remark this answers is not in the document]
Yiyiyimu: One thing I'm not so sure about the constructor part. Do we need 
```go
func newPodArg() *PodArg {
    p := new(PodArg)
    p.Status = v1.PodRunning
    return p
}
```
or
```go
func newPodArg(status v1.PodPhase) *PodArg {
    p := new(PodArg)
    if status == nil {
        p.Status = v1.PodRunning
    }
    return p
}
```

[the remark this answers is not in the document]
STRRL: Just creating each constructors that ALL you need. 😆
As go do not support method overloading like java, so you should use different method names for the different constructors.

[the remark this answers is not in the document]
Yiyiyimu: Wait do you mean I should create `newPodArg`, `newPodArgForRunning`, `newPodArgForDefaultNs` and `newPodArgForRunningAndDefaultNs`? Yes it could explicitly declare the pod status, but it's ... just a bit weird to me... I'd rather still expose them as arguments in this case 🤪

But I still believe it is acceptable not explicitly declaring arguments that not matter. What do you think about it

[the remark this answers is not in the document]
STRRL: TLDR; You are right, it is acceptable. 🙈

It's also indeed weird to me. And I could not provide a perfect solution. 🙈 

I could share my thought about this situation:

- it's test cases, so I would like to claim all states explicitly as possible and make comments on which state will affect the logic.
- then I consider the fluent codes, the "builder pattern"`NewDefaultBuilder().SetPhase().SetNs().Build()` is the best choice for one-line construction, but it will take a lot of codes to create a builder. So I prefer just write a plain `Struct`  with all fields to here.

When another developer wants to understand these codes, he/she might feel confusing about the empty field of the struct, until he/she find the default value logic in `newPods()`.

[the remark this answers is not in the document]
Yiyiyimu: Yeah I can see that, cause I'm also a bit unsure about not explicitly declare everything, but the attraction of making code clean is just so much... I tried to add tags and comments to `PodArg` to declare default values are set. It might be of help to other developers 🙈

pkg/utils/generate.go
```diff
@@ -0,0 +1,111 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// PodArg by default use `Status=corev1.PodRunning` and `Namespace=metav1.NamespaceDefault`.
+// For the others, the default values are empty.
+type PodArg struct {
+	Name            string
+	Status          v1.PodPhase `default:"corev1.PodRunning"`
+	Namespace       string      `default:"metav1.NamespaceDefault"`
```
STRRL, line 29: "struct tag" always be used for describing specific logic on a field, like `json`, `protobuf`

 It seems we do not import any library for that, so I think we should delete `default` struct tag.

Yiyiyimu: Thanks! I didn't know that before.

Actually I got the idea from [go-defaults](https://github.com/mcuadros/go-defaults). Then I guess I would either use that library or explicitly use `reflect` to get the tag value. Which one do you think is better

STRRL: It seems you need to call `defaults.SetDefaults()` for setting default values, we often do it in the constructor. So you might need a constructor. 🙈

So how about deleting the "default" struct tag?

Yiyiyimu: As the [caveats of repo](https://github.com/mcuadros/go-defaults#caveats) shows
> At the moment, the way the default filler checks whether it should fill a struct field or not is by comparing the current field value with the corresponding zero value of that type.

So we could still put `defaults.SetDefaults()` into `newPod`, replacing the manual default values set.

I think the tags could somewhat highlight the default values, but I'm still ok to delete them.

STRRL: > So we could still put `defaults.SetDefaults()` into `newPod`, replacing the manual default values set.

Please don't, it makes `newPod` more and more complicated, just remove it. 🙈

Yiyiyimu: Got it. Deleted

pkg/utils/generate.go
```diff
@@ -0,0 +1,111 @@
+// Copyright 2019 Chaos Mesh Authors.
```
fewdan, line 1: s/2019/2020/

fewdan (changes requested):

pkg/utils/generate.go
Yiyiyimu: thx! fixed

STRRL (approved):
LGTM

WangXiangUSTC (approved):
LGTM

PR #1256, opened 2020-12-07
split pkg/utils into multiple different pkg or file

### LINKED ISSUE #1254: Reorganize `utils`
We have a pkg `pkg/utils`. However, it's a bad idea to collect these unrelated functions into one pkg, because it's a hidden danger for cycle importing.

We should investigate these functions carefully and split them into multiple pkg.

YangKeao:

Fix #1254. Collecting multiple unrelated functions into one pkg is a bad idea, because the more it imports, the less it can be imported in other pkgs.

During the splitting, I followed these rules:

1. If the function is only used by one pkg, then it will be moved into the `utils.go` in that pkg.

2. If the function is used by multiple pkgs, then it will be moved into a standalone pkg to provide the utils.

In #1206, I have faced a lot of cycle importing problems caused by other files in utils 😢 .


WangXiangUSTC:
```log
ERROR: image: "localhost:5000/pingcap/chaos-fs:98b0bf5027a1226b43ba099eb5902b748c16ff62" not present locally
```

e2e test failed, we should remove chaos-fs from e2e-test

WangXiangUSTC (approved):
LGTM

cmd/controller-manager/utils.go
```diff
@@ -11,16 +11,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
```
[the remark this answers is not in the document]
YangKeao: I wonder why we don't use `controller / reconciler` pattern in `watchConfig`. Inside the `controller`, it's a `RateLimitQueue` :(

I will find the problem or replace it into a controller.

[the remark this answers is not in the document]
YangKeao: > why we don't use `controller / reconciler`.

Because https://github.com/kubernetes-sigs/controller-runtime/issues/244 hasn't landed, we cannot listen to `ConfigMap` with a specified label. Listening to all `ConfigMap` on the cluster is a disaster for cache and network.

I can try to use `RateLimitQueue` to replace `Coalescer` here.

STRRL:
Hi @YangKeao , is this PR ready to review?

YangKeao:
@STRRL  Yes! I have merged with master.

cmd/controller-manager/main.go
```diff
@@ -189,62 +190,74 @@ func main() {
...
-		sigChan := make(chan interface{}, 10)
-		//debouncedChan := make(chan interface{}, 10)
+func setupWatchQueue(stopCh <-chan struct{}, configWatcher *watcher.K8sConfigMapWatcher) workqueue.Interface {
+	// watch for reconciliation signals, and grab configmaps, then update the running configuration
+	// for the server
+	sigChan := make(chan interface{}, 10)
 
-		// debounce events from sigChan, so we dont hammer apiserver on reconciliation
-		eventsCh := utils.Coalescer(EventCoalesceWindow, sigChan, stopCh)
+	queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
```
STRRL, line 200: I think it might be too frequent than the previous one.

How about reconsidering customizing the `RateLimiter`?

And I think we don't need `EventCoalesceWindow`  anymore, please delete it.

STRRL (changes requested):
rest LGTM, PTAL @YangKeao  😁

WangXiangUSTC (approved):
LGTM

cmd/controller-manager/main.go
YangKeao: OK! I set qps to 0.5

cmd/controller-manager/main.go
```diff
@@ -189,62 +189,75 @@ func main() {
...
+
+	return queue
+}
+
+func watchConfig(configWatcher *watcher.K8sConfigMapWatcher, cfg *config.Config, stopCh <-chan struct{}) {
+	queue := setupWatchQueue(stopCh, configWatcher)
+
+	for {
+		item, shutdown := queue.Get()
+		defer queue.Done(item)
```
STRRL, line 246: I think we should not use `defer` in a for "loop", it will add the defer logic for outer function `watchConfig`, not for each loop.

We could use a closure `func() {}()` for using defer in the for loop.

```go
for {
  if item, shutdown := queue.Get(); shutdown {
    break
  }
  func (){
    defer queue.Done(item)
    // other codes here
  }()
}
```

pkg/chaosctl/cmd/logs.go
```diff
@@ -26,7 +26,7 @@ import (
 	"github.com/chaos-mesh/chaos-mesh/api/v1alpha1"
 	"github.com/chaos-mesh/chaos-mesh/controllers/config"
 	cm "github.com/chaos-mesh/chaos-mesh/pkg/chaosctl/common"
-	"github.com/chaos-mesh/chaos-mesh/pkg/utils"
+	sele "github.com/chaos-mesh/chaos-mesh/pkg/selector"
```
STRRL, line 29: This abbr sounds a little weird. 🙈 How about using the full "selector".

pkg/chaosctl/common/common.go
```diff
@@ -34,9 +34,10 @@ import (
 
 	"github.com/chaos-mesh/chaos-mesh/api/v1alpha1"
 	ctrlconfig "github.com/chaos-mesh/chaos-mesh/controllers/config"
+	daemonClient "github.com/chaos-mesh/chaos-mesh/pkg/chaosdaemon/client"
 	"github.com/chaos-mesh/chaos-mesh/pkg/chaosdaemon/pb"
 	"github.com/chaos-mesh/chaos-mesh/pkg/portforward"
-	"github.com/chaos-mesh/chaos-mesh/pkg/utils"
+	sele "github.com/chaos-mesh/chaos-mesh/pkg/selector"
```
STRRL, line 40: Same request for using `selector` instead of `sele`

STRRL (changes requested):
rest LGTM

cmd/controller-manager/main.go
YangKeao: Fixed

pkg/chaosctl/cmd/logs.go
YangKeao: Fixed

pkg/chaosctl/common/common.go
YangKeao: Fixed

STRRL (approved):
LGTM

WangXiangUSTC (approved):
LGTM

PR #1257, opened 2020-12-08
change the default hostNetwork value of the Chaos daemon to true

devincd:
Change the default hostNetwork value of the Chaos daemon to true for NetworkChaos/TimeChaos experiment

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/helm/chaos-mesh/values.yaml
+++ b/helm/chaos-mesh/values.yaml
@@ -74,7 +74,7 @@ chaosDaemon:
   grpcPort: 31767
   httpPort: 31766
   env: {}
-  hostNetwork: false
+  hostNetwork: true
 
   podAnnotations: {}
 


YangKeao:
`hostNetwork` is only needed when your cluster doesn't support `hostPort`.

 And in the latest version (in master), `chaos-daemon` doesn't need `hostNetwork` and `hostPort` anymore.

devincd:
> `hostNetwork` is only needed when your cluster doesn't support `hostPort`.
> 
> And in the latest version (in master, as #1206 is merged), `chaos-daemon` doesn't need `hostNetwork` and `hostPort` anymore.

Ok, This is a good news! Connect to the daemon through the endpoint, if the daemon's IP changed, did it handler the reconnection? Or is it a new connection every time? Would it be inappropriate to create a connection every time?

STRRL:
Hi @devincd, it will establish new connections every time.

I don't think it will be the performance bottleneck in the next several versions. So we do not need to "improve" this right now.

PR #1260, opened 2020-12-08
Upgrade toda and use nsexec

YangKeao:

1. Removed modified glibc
2. `toda` should run in the target `pid` and `mnt` namespace
3. `nsexec` is a new tool to replace `nsenter`. The significant advantages of `nsexec` is that:
    1. `nsexec` will passthrough `SIGTERM` and `SIGINT` to the child process, but `nsenter` will not.
    2. `nsexec` can launch the program at local mnt namespace even if it doesn't exist in the target namespace. This should also be valuable for other cases (like debug tools)

This will solve the problem of child processes leak. And it may also fix the problem of running `toda` on CentOS or other distributions directly :+1:

After passing the e2e test, I will move `nsexec` into the `chaos-mesh` organization.


pkg/bpm/build_linux.go
```diff
@@ -48,9 +48,13 @@ func (b *ProcessBuilder) Build() *ManagedProcess {
 	if len(b.nsOptions) > 0 {
 		args = append([]string{"--", cmd}, args...)
 		for _, option := range b.nsOptions {
-			args = append([]string{"-" + nsArgMap[option.Typ] + option.Path}, args...)
+			args = append([]string{"-" + nsArgMap[option.Typ], option.Path}, args...)
```
cwen0, line 51: The commends of this function should be updated

pkg/bpm/bpm.go
```diff
@@ -32,26 +32,31 @@ type NsType string
...
+	IpcNS: "i",
+	NetNS: "n",
+	PidNS: "p",
+	// user namespace is not supported by nsexec yet
+	// UserNS:  "U",
 }
 
 const (
 	pausePath   = "/usr/local/bin/pause"
 	suicidePath = "/usr/local/bin/suicide"
```
cwen0, line 57: Can we remove `suicide`?

pkg/bpm/build_linux.go
YangKeao: ?

cwen0: [L27 ~ L38](https://github.com/chaos-mesh/chaos-mesh/pull/1260/files/3e228a2db9f7032501587f6a81cb0e7cfe7d6830#diff-5f25abd3df2ef9744e8fdf10bd4d1f99bcb23bd450fb3b886e63074ab0e3b56fL27-L39)

pkg/bpm/bpm.go
YangKeao: Nice catch. Yes, we can.

pkg/bpm/build_linux.go
YangKeao: Do you mean `comments`? I have deleted these comments :beers: , as it's easier to understand and there's nothing special now.

cwen0 (approved):
LGTM

pkg/bpm/build_linux.go
cwen0: Yes, my fault.

WangXiangUSTC (approved):
LGTM

STRRL:
Hi @YangKeao , may I ask why the "uts" nad "user" namespace is not supported yet? 

Its implements need more extra operations? Or meet some limitations? Or just not write such codes yet? Or just haven't write code yet?

PR #1261, opened 2020-12-09
dashboard: add authority for archive experiments and events

WangXiangUSTC:

add authority for archive experiments and events

when querying the archive experiments and events, will use `SelfSubjectAccessReviews` to judge whether the token can list the information.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


WangXiangUSTC:
@g1eny0ung I update the HTTP request on front-end at 408ef7d, PTAL

pkg/apiserver/event/event.go
```diff
@@ -80,12 +80,19 @@ func (s *Service) listEvents(c *gin.Context) {
...
 		FinishTimeStr:       c.Query("finishTime"),
 		ExperimentName:      c.Query("experimentName"),
-		ExperimentNamespace: c.Query("experimentNamespace"),
+		ExperimentNamespace: c.Query("namespace"),
 		UID:                 c.Query("uid"),
 		Kind:                c.Query("kind"),
 		LimitStr:            c.Query("limit"),
 	}
 
+	fmt.Println("listEvents", filter)
```
g1eny0ung, line 89: Omit debug print.

ui/src/api/events.ts
```diff
@@ -2,13 +2,17 @@ import { Event, EventsParams } from './events.type'
...
   http.get<Event[]>('/events', {
-    params,
+    params: {
+      namespace,
+    },
   })
 
 // Without pods field
-export const dryEvents = (params?: EventsParams) =>
+export const dryEvents = (namespace = '') =>
```
g1eny0ung, line 13: `namespace = ''` to `namespace = null` will be better.

g1eny0ung: And `EventsParams` can be removed. It will be useless.

pkg/apiserver/event/event.go
WangXiangUSTC: done

ui/src/api/events.ts
WangXiangUSTC: done

g1eny0ung (approved):
LGTM

WangXiangUSTC:
@STRRL PTAL

STRRL (commented):
How about printing some log when `utils.CanListChaos` return `false`.

The current implementation is NOT so good for presenting authorization failed with HTTP API: 
- It seems just to return empty results(not empty array`[]`) with HTTP status 200, and the frontend needs more logic to resolve it.
- We should return `403` with authorization failed.

If this implementation is a temporary solution, just appending some logs is OK to me.

WangXiangUSTC:
> How about printing some log when `utils.CanListChaos` return `false`.
> 
> The current implementation is NOT so good for presenting authorization failed with HTTP API:
> 
> * It seems just to return empty results(not empty array`[]`) with HTTP status 200, and the frontend needs more logic to resolve it.
> * We should return `403` with authorization failed.
> 
> If this implementation is a temporary solution, just appending some logs is OK to me.

I take the `gin.Context` as a parameter of `CanListChaos`, and will take the error info to `gin.Context`, if `CanListChaos` return false, will print the error log in dashboard @STRRL

STRRL:
Oh, I missed it.🙈

But I think making errors in `CanListChaos` is not so good. 😰 Any operations about HTTP should be restricted in Controllers.

cwen0:
Why not implementing a specific middleware to define this authority?

pkg/apiserver/archive/archive.go
```diff
@@ -93,6 +93,11 @@ func (s *Service) list(c *gin.Context) {
 	name := c.Query("name")
 	ns := c.Query("namespace")
 
+	canList := utils.CanListChaos(c, ns)
+	if !canList {
+		return
```
fewdan, line 98: Is it more reasonable to return some errors here, such as returning error codes (403 and so on?) ?

WangXiangUSTC: yes, you are right, I will refine this logic, maybe use the middleware @cwen0 mentioned

WangXiangUSTC:
> Why not implementing a specific middleware to define this authority?

refine this logic by using the middleware in https://github.com/chaos-mesh/chaos-mesh/pull/1261/commits/584aa1875f6c3afdacb804eec8a2b037551577c8, PTAL again @STRRL @cwen0 @g1eny0ung @fewdan

fewdan (approved):
LGTM

g1eny0ung (approved):
LGTM

PR #1286, opened 2020-12-15
Bugfix: Use `Last` to get current phase at start time of standard cron

### LINKED ISSUE #1270: Bug: chaos would directly start when it got applied, even if the cron doesn't want it be so
## Bug Report

**What did you do?**
If we use the cron that does not start right away, for example, if I want to start a chaos at 00:00 each day for 10 minutes, we normally should go into Waiting phase. However, in [the current implementation](https://github.com/chaos-mesh/chaos-mesh/blob/cf9e1ca11670df27b2b8995b0f4942cf96142330/controllers/twophase/types.go#L152) or [the new implementation](https://github.com/chaos-mesh/chaos-mesh/pull/1221/files/3542444758bc14fbca49fbec2b6515555dc4779f#diff-04c792dfed57630c1de983400da2e7639c6d5c3a929c0124f7b5c87aa9a19105)  of two-phase controllers, the chaos would directly go into Running state.

To deal with it, if we use standard cron, it would be a good idea to figure out whether right now should be in waiting or running phase, to go into the correct one. Thus we need to get `lastStart` and `lastRecover`. Although someone creates [a PR about this feature](https://github.com/robfig/cron/pull/216) in cron repo, the maintainer thinks it's outside the scope of that repo so we need to implement this ourselves. Plus, this feature would also be of help to implement pausing for duration(#1016)

Related discussion: [1](https://github.com/chaos-mesh/chaos-mesh/pull/1221#discussion_r535293482), [2](https://github.com/chaos-mesh/chaos-mesh/pull/1221#discussion_r539943354)

chaos go from uninitialized to waiting, since it's not the time to start

**What did you see instead?**
chaos go from uninitialized to running

Yiyiyimu:
fix #1270

Add function `Last` to get the last start time of standard cron, so we could know the current phase that chaos should go into when start.

### Checklist

Tests

- [x] Unit test
- [x] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


WangXiangUSTC:
get a new error
```log
• Failure [19.740 seconds]
[2020-12-17T03:06:08.400Z] [6] [Basic]
[2020-12-17T03:06:08.400Z] [6] /home/jenkins/agent/workspace/chaos-mesh-e2e/go/src/github.com/chaos-mesh/chaos-mesh/test/e2e/chaos/basic.go:50
[2020-12-17T03:06:08.400Z] [6]   [TimeChaos]
[2020-12-17T03:06:08.400Z] [6]   /home/jenkins/agent/workspace/chaos-mesh-e2e/go/src/github.com/chaos-mesh/chaos-mesh/test/e2e/chaos/basic.go:114
[2020-12-17T03:06:08.400Z] [6]     [TimeSkew]
[2020-12-17T03:06:08.400Z] [6]     /home/jenkins/agent/workspace/chaos-mesh-e2e/go/src/github.com/chaos-mesh/chaos-mesh/test/e2e/chaos/basic.go:140
[2020-12-17T03:06:08.400Z] [6]       [StartAtWaiting] [It]
[2020-12-17T03:06:08.400Z] [6]       /home/jenkins/agent/workspace/chaos-mesh-e2e/go/src/github.com/chaos-mesh/chaos-mesh/test/e2e/chaos/basic.go:150
[2020-12-17T03:06:08.400Z] [6] 
[2020-12-17T03:06:08.400Z] [6]       Dec 17 03:06:07.966: create time chaos error
[2020-12-17T03:06:08.400Z] [6]       Unexpected error:
[2020-12-17T03:06:08.400Z] [6]           <*errors.StatusError | 0xc001156280>: {
[2020-12-17T03:06:08.400Z] [6]               ErrStatus: {
[2020-12-17T03:06:08.400Z] [6]                   TypeMeta: {Kind: "", APIVersion: ""},
[2020-12-17T03:06:08.400Z] [6]                   ListMeta: {
[2020-12-17T03:06:08.400Z] [6]                       SelfLink: "",
[2020-12-17T03:06:08.400Z] [6]                       ResourceVersion: "",
[2020-12-17T03:06:08.400Z] [6]                       Continue: "",
[2020-12-17T03:06:08.400Z] [6]                       RemainingItemCount: nil,
[2020-12-17T03:06:08.400Z] [6]                   },
[2020-12-17T03:06:08.400Z] [6]                   Status: "Failure",
[2020-12-17T03:06:08.400Z] [6]                   Message: "admission webhook \"vtimechaos.kb.io\" denied the request: spec.scheduler.cron: Invalid value: \"9 * * * *\": the scheduling interval:\"9 * * * *\" must be greater than the duration:9m0s",
[2020-12-17T03:06:08.400Z] [6]                   Reason: "spec.scheduler.cron: Invalid value: \"9 * * * *\": the scheduling interval:\"9 * * * *\" must be greater than the duration:9m0s",
[2020-12-17T03:06:08.400Z] [6]                   Details: nil,
[2020-12-17T03:06:08.400Z] [6]                   Code: 403,
[2020-12-17T03:06:08.400Z] [6]               },
[2020-12-17T03:06:08.400Z] [6]           }
[2020-12-17T03:06:08.400Z] [6]           admission webhook "vtimechaos.kb.io" denied the request: spec.scheduler.cron: Invalid value: "9 * * * *": the scheduling interval:"9 * * * *" must be greater than the duration:9m0s
[2020-12-17T03:06:08.400Z] [6]       occurred
[2020-12-17T03:06:08.400Z] [6] 
[2020-12-17T03:06:08.400Z] [6]       /home/jenkins/agent/workspace/chaos-mesh-e2e/go/src/github.com/chaos-mesh/chaos-mesh/test/e2e/chaos/timechaos/time_skew.go:262
```

Yiyiyimu:
@WangXiangUSTC I think that is a bug of this part.
https://github.com/chaos-mesh/chaos-mesh/blob/9fb861813f53eb0c2203802ec32ff15ee9ab4dcd/api/v1alpha1/common_validation.go#L73-L81
The interval should be `scheduler.Next(tmpTime).Sub(nextTime)`. The original one only considers `@every` situation but not `* * * * *`

Yiyiyimu:
Hi @WangXiangUSTC I just updated to the new version.

WangXiangUSTC:
> @WangXiangUSTC I think that is a bug of this part.
> https://github.com/chaos-mesh/chaos-mesh/blob/9fb861813f53eb0c2203802ec32ff15ee9ab4dcd/api/v1alpha1/common_validation.go#L73-L81
> 
> 
> The interval should be `scheduler.Next(tmpTime).Sub(nextTime)`. The original one only considers `@every` situation but not standard cron `* * * * *`.
> Besides, I think we could remove this part, since interval of standard cron is not always the same, for example: `*/29 * * * *`. And right now we have [`iterateNextTime`](https://github.com/chaos-mesh/chaos-mesh/blob/b0e257c9859c8601e8fe636da91c6acf99292acd/controllers/twophase/state_machine.go#L205) so if interval is less than duration, the next start would be directly skipped.

agree with you

Yiyiyimu:
Hi @WangXiangUSTC it seems the same problem happens again 😢

WangXiangUSTC:
> Hi @WangXiangUSTC it seems the same problem happens again 😢

Can these tests run successfully in your local environment?

And we are refining the e2e test to make it faster and stable recently 😸

Yiyiyimu:
Ah... it works this time... 🙉 

Yeah running single case (the failed ones) would succeed in my local environment. But since the e2e test in CI creates 6 routines but in local it is just 1, and the problem seems coming from routines, I could not really reproduce this on my local computer 😢

WangXiangUSTC:
> Ah... it works this time... 🙉
> 
> Yeah running single case (the failed ones) would succeed in my local environment. But since the e2e test in CI creates 6 routines but in local it is just 1, and the problem seems coming from routines, I could not really reproduce this on my local computer 😢
> 
> I noticed the discussion in #1279 and it relieves me a bit 🤣 But success in 1 out of 4 still unacceptable for CI working 🙈

Yes, the success rate is too low now...

Yiyiyimu:
Hi @STRRL @Colstuwjx it seems I create a super-unstable e2e test (success rate is 2 out of 6 for now), do you mind take a look?

The thing is, the new test itself would always succeed, but one test running at the same time with the new test in another routine would failed somehow. Running the failed test itself would succeed in my local environment, and the failed test is changing all the time, so I believe it is caused by the new test. But I could not figure out any possible reasons.

cwen0:
@Yiyiyimu PR #1355 has been merged. Can you update your branch and run e2e tests.

YangKeao:
After this PR gets merged, I would like to add a `pause` controller to split the logic in twophase controller.

What it needs to do is really simple: 1. If the `pause` annotation is on, the `router` will pass it to the `pause` controller. 2. The `pause` controller will recover the chaos (if needed, which can be checked by status), and remove all time related message in status (like `nextStart` and `nextRecover`...). 3. When the `pause` annotation is off, it will be passed to corresponding controller (`twophase` or `common`) just like a newly created chaos (which doesn't have `nextStart` and `nextRecover`).

And then, the `twophase` controller could be much simpler :beers: !

It's only a rough imagination right now, and I don't know whether it will actually work. Do you have any idea about it?

Yiyiyimu:
Sorry for the late reply. Will fix it this weekend

api/v1alpha1/common_validation.go
```diff
@@ -63,22 +63,11 @@ func ValidateScheduler(schedulerObject InnerSchedulerObject, spec *field.Path) f
...
-
-		if scheduler != nil {
-			tmpTime := time.Time{}
-			nextTime := scheduler.Next(tmpTime)
-			interval := nextTime.Sub(tmpTime)
-			if *duration >= interval {
-				allErrs = append(allErrs, field.Invalid(cronField, spec.Cron,
-					fmt.Sprintf("the scheduling interval:\"%s\" must be greater than the duration:%s", spec.Cron, *duration)))
-			}
-		}
```
WangXiangUSTC, line 81: Why remove the check, seems that we can get the `interval` in another way(get the next time t1, and the next time t2 of t1, interval = t2 - t1)?

test/e2e/chaos/timechaos/time_skew.go
```diff
@@ -215,3 +216,100 @@ func TestcaseTimeSkewPauseThenUnpause(
...
+		err = cli.Get(ctx, chaosKey, chaos)
+		framework.ExpectNoError(err, "get time chaos error")
+		if chaos.Status.Experiment.Phase == v1alpha1.ExperimentPhaseWaiting {
+			return true, nil
+		}
+		return false, err
+	})
+	framework.ExpectNoError(err, "check chaos in waiting failed")
+
+	framework.ExpectNoError(err, "get timer pod error")
```
WangXiangUSTC, line 279: this line can be deleted

test/e2e/chaos/timechaos/time_skew.go
```diff
@@ -215,3 +216,100 @@ func TestcaseTimeSkewPauseThenUnpause(
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	By("wait e2e helper ready")
+	err := util.WaitE2EHelperReady(c, port)
+	framework.ExpectNoError(err, "wait e2e helper ready error")
+
+	By("create cron")
+	minuteNow := time.Now().Minute()
+	crontab := fmt.Sprintf("%d * * * *", (minuteNow+3)%60)
```
WangXiangUSTC, line 235: seems that this test will run at least 3 minutes, can we decrease the time cost?

pkg/scheduler/scheduler.go
```diff
@@ -0,0 +1,149 @@
...
+
+	"github.com/chaos-mesh/chaos-mesh/api/v1alpha1"
+)
+
+const (
+	// Set the top bit if a star was included in the expression.
+	starBit = 1 << 63
+)
+
+// LastTime returns the next time this schedule activated, less than or equal with the given time.
```
WangXiangUSTC, line 30:
```suggestion
// LastTime returns the last time this schedule activated, less than or equal with the given time.
```

pkg/scheduler/scheduler.go
```diff
@@ -0,0 +1,149 @@
...
+func LastTime(spec v1alpha1.SchedulerSpec, now time.Time) (*time.Time, error) {
+	scheduler, err := cron.ParseStandard(spec.Cron)
+	if err != nil {
+		return nil, fmt.Errorf("fail to parse runner rule %s, %v", spec.Cron, err)
+	}
+	var next time.Time
```
WangXiangUSTC, line 36:
```suggestion
	var last time.Time
```

api/v1alpha1/common_validation.go
Yiyiyimu: https://github.com/chaos-mesh/chaos-mesh/pull/1286#issuecomment-747232617
We talked about it before 🤣

WangXiangUSTC: 🥵

pkg/scheduler/scheduler.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// Last returns the last time this schedule activated, less than or equal with the given time.
+// If no time can be found to satisfy the schedule, return the zero time.
+// Modified from the original `Next` function in robfig/cron at Dec 15, 2020
+func (s *cusSchedule) Last(t time.Time) time.Time {
+	// General approach:
+	// For Month, Day, Hour, Minute, Second:
+	// Check if the time value matches.  If yes, continue to the next field.
+	// If the field doesn't match the schedule, then increment the field until it matches.
```
WangXiangUSTC, line 70: Should `increment` change to `decrease`?

WangXiangUSTC: this function decrease 1 second every time, may loop too much times, is there other better ways to get the last time?

test/e2e/chaos/timechaos/time_skew.go
Yiyiyimu: Actually `+3` means it would take 2-3 mins (say we do the plus at 59 seconds and it would only take 2 mins to start)
And since we need to wait at most 1 minutes to make sure chaos start and into wait phase, and another minute to make sure chaos not take effect, I think `+3` is the minimum time we need to wait. Or we need to cut out those "1 minutes" to lower duration say 30s, maybe that would import some unstability

pkg/scheduler/scheduler.go
Yiyiyimu: > Should increment change to decrease?
Thanks! Fixed.

> this function decrease 1 second every time, may loop too much times, is there other better ways to get the last time?
Implementing `-1s`(😎) is to decrease to the last block of time, say last year. For example, when we are at the first second of 2021, -1s could take us to 2020 but not to miss any seconds in 2020.
Since it's used to decrease block of time, the efficiency is still the same with [function Next in cron pkg](https://github.com/robfig/cron/blob/master/spec.go).

test/e2e/chaos/timechaos/time_skew.go
Yiyiyimu: Thanks! Fixed

pkg/scheduler/scheduler.go
Yiyiyimu: Thanks! Fixed

pkg/scheduler/scheduler.go
Yiyiyimu: Thanks! Fixed

Yiyiyimu:
@WangXiangUSTC PTAL

WangXiangUSTC:
@YangKeao PTAL

WangXiangUSTC (approved):
LGTM

Yiyiyimu:
> After this PR gets merged, I would like to add a `pause` controller to split the logic in twophase controller.
> 
> What it needs to do is really simple: 1. If the `pause` annotation is on, the `router` will pass it to the `pause` controller. 2. The `pause` controller will recover the chaos (if needed, which can be checked by status), and remove all time related message in status (like `nextStart` and `nextRecover`...). 3. When the `pause` annotation is off, it will be passed to corresponding controller (`twophase` or `common`) just like a newly created chaos (which doesn't have `nextStart` and `nextRecover`).
> 
> And then, the `twophase` controller could be much simpler 🍻 !
> 
> It's only a rough imagination right now, and I don't know whether it will actually work. Do you have any idea about it?

Sorry I just realized I missed this message 😬 

Split pause from twophase will of course make the case much easier~~
Just one question, which I think we've discussed before, is that how should we define or what do people need of "pause", like when pause is off, should we obey the original cron cycle (like what we do for now), or just start the chaos like a new one.

YangKeao (approved):
LGTM

cwen0:
@Yiyiyimu Can you cherry-pick this pr to release-1.1 branch manualy?

Yiyiyimu:
> @Yiyiyimu Can you cherry-pick this pr to release-1.1 branch manualy?

Got it

PR #1300, opened 2020-12-18
update e2e base image

YangKeao:

Contain a lot of go build-cache, download cache, and binaries in the base e2e image.


Makefile
```diff
@@ -175,70 +167,123 @@ tidy: clean
...
+CLEAN_TARGETS += $(2)
+ifeq ($(CHAOS_MESH_BUILD_IN_DOCKER),1)
+$(2): image-build-env go_build_cache_directory
+	docker run --rm --workdir /mnt/ --volume $(shell pwd):/mnt \
+		$(BUILD_INDOCKER_ARG) --env UI=${UI} --env SWAGGER=${SWAGGER} \
+		--user $(shell id -u):$(shell id -g) ${DOCKER_REGISTRY_PREFIX}pingcap/build-env \
+		/usr/bin/make $(2)
+endif
+image-$(1)-dependencies := $(image-$(1)-dependencies) $(2)
+endef
```
WangXiangUSTC, line 215: do we need to add some indent in this block?

Makefile
```diff
@@ -60,11 +66,16 @@ ifeq ($(UI),1)
 	BUILD_TAGS += ui_server
 endif
 
-all: yaml image
+CLEAN_TARGETS :=
+
+CHAOS_MESH_BUILD_IN_DOCKER ?= 1
+ifeq (${IN_DOCKER},1)
+	CHAOS_MESH_BUILD_IN_DOCKER = 0
+endif
```
WangXiangUSTC, line 74: why defining a new variable, can we use `IN_DOCKER` directly?

Makefile
```diff
@@ -175,70 +167,123 @@ tidy: clean
...
-taily-build:
-	if [ "$(shell docker ps --filter=name=$@ -q)" = "" ]; then \
-		docker build -t pingcap/chaos-binary ${DOCKER_BUILD_ARGS} .; \
-		docker run --rm --mount type=bind,source=$(shell pwd),target=/src \
-			--name $@ -d pingcap/chaos-binary tail -f /dev/null; \
-	fi;
-
-taily-build-clean:
-	docker kill taily-build && docker rm taily-build || exit 0
+image: image-chaos-daemon image-chaos-mesh image-chaos-dashboard
```
WangXiangUSTC, line 182: I run `make image` on my local computer, but get the error below:
```log
make: *** No rule to make target `image-chaos-daemon', needed by `image'.  Stop.
```

Makefile
YangKeao: Makefile is indent-sensitive. I don't know whether I can add indents without breaking it.

Makefile
YangKeao: `CHAOS_MESH_BUILD_IN_DOCKER ` means whether the user wants to build chaos-mesh binary in a docker container, which may fail for docker in VM because of the volume mount (like docker in Mac OS and windows). However, even with `CHAOS_MESH_BUILD_IN_DOCKER `, it may fail to build on Mac OS and windows (as the cgo is used).

But as the `--mount=cache` cannot be exported, and build an image depending on another image is not well supported by `buildx` (I cannot load a local image into the buildkit), I cannot find a way to keep the go build-cache while supporting building on mac os.

Makefile
YangKeao: Remove `=` will solve this problem. Fixed

Makefile
YangKeao: Maybe I could fall back to the traditional way for Mac OS users 😋

Makefile
```diff
@@ -91,45 +102,26 @@ else
...
-
-chaos-dashboard:
-ifeq ($(SWAGGER),1)
-	make swagger_spec
-endif
-ifeq ($(UI),1)
-	make ui
-	hack/embed_ui_assets.sh
-endif
-	$(CGO) build -ldflags "$(LDFLAGS)" -tags "${BUILD_TAGS}" -o bin/chaos-dashboard cmd/chaos-dashboard/*.go
```
WangXiangUSTC, line 116: seems that we can not build binary later

YangKeao: `make binary` has been added :smile_cat:

Makefile
YangKeao: It has been removed.

WangXiangUSTC (approved):
LGTM

fewdan (approved):
LGTM

PR #1302, opened 2020-12-18
feat: chaosctl for print debug info (#1074)

### LINKED ISSUE #694: Debug information collector for chaos-mesh

Yiyiyimu:
cherry-pick #1074 to release-1.0

---

### What's this PR did
fix #694 

### Checklist
**SUPPORT CHAOS**
- [x] NetworkChaos
- [x] StressChaos
- [x] IoChaos

**DEBUG**
- [x] compare with expected (a demo for networkChaos)
- [x] move to ctrl.client from shell
- [x] chaos applied on multiple pods (larger radius): use utils.selectPods
- [x] replace extractFromJson with type assert (as how controllers do)
- [x] create daemon client locally to get pid more robustly
- [x] add test case
- [x] add backup for pod that not has `ls`, `ps` and `cat`
- [x] add feature to print log from all chaos-mesh components as `chaosctl logs`

**CTL**
- [x] add color to output title
- [x] use format "chaosctl debug xxxchaos"
- [x] support chaos in certain namespace
- [x] use kubectl compatible auto-completion as `chaosctl debug xxxchaos CHAOSNAME -n NAMESPACE`


Yiyiyimu:
Since #1074 use the result of #1252, without cherry-pick the later one I directly move the [test functions it needs](https://github.com/chaos-mesh/chaos-mesh/pull/1302/files#diff-3d439b4a7761fc348bbb91cb83b009d3e90f12b3ef3c7e10af777e2da429ede4) to `pkg/utils/generate.go`, to make the test work

STRRL (approved):
LGTM

PR #1309, opened 2020-12-19
chaosdaemon: make more TC filters work

cwen0:

Support more TC filters. 

If the TC request doesn't include `ipset` field, the other filters won't work. Because we use `ipset` to distinguish if the tc rules need filters.  

https://github.com/chaos-mesh/chaos-mesh/blob/master/pkg/chaosdaemon/tc_server.go#L122-L127 

```
	for _, tc := range in.Tcs {
		if tc.Ipset == "" {
			globalTc = append(globalTc, tc)
		} else {
			// TODO: support multiple tc with one ipset
			filterTc[tc.Ipset] = append(filterTc[tc.Ipset], tc)
		}
	}
```  

Check all TC filters(ipset, source port, egress port, Protocol) to distinguish if we need to set TC rules with filters. 

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


STRRL (approved):
LGTM

pkg/chaosdaemon/tc_server.go
```diff
@@ -405,3 +431,21 @@ func convertTbfToArgs(tbf *pb.Tbf) string {
 
 	return args
 }
+
+func abstractTcFilter(tc *pb.Tc) string {
```
YangKeao, line 435: Use `filterExist(tc *pb.Tc) bool`, as the `string` is meaningless.

YangKeao (changes requested):

pkg/chaosdaemon/tc_server.go
cwen0: This filter key is used to distinguish TC rules with the same filter. If we don't consider this, `filterExist(tc *pb.Tc) bool` is great!

YangKeao: OK. I understand.

YangKeao (approved):
LGTM

PR #1316, opened 2020-12-22
add EnterNS field for grpc request

WangXiangUSTC:

add `withoutNS` field for grpc request, if `withoutNS` is true, ignore `SetNS` for command

used for chaosd, so don't need to deploy `enterns` or `nsexec`

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


pkg/chaosdaemon/dns_server.go
```diff
@@ -45,10 +45,12 @@ func (s *DaemonServer) SetDNSServer(ctx context.Context,
 		}
 
 		// backup the /etc/resolv.conf
-		cmd := bpm.DefaultProcessBuilder("sh", "-c", fmt.Sprintf("ls %s.chaos.bak || cp %s %s.chaos.bak", DNSServerConfFile, DNSServerConfFile, DNSServerConfFile)).
-			SetNS(pid, bpm.MountNS).
-			SetContext(ctx).
-			Build()
+		processBuilder := bpm.DefaultProcessBuilder("sh", "-c", fmt.Sprintf("ls %s.chaos.bak || cp %s %s.chaos.bak", DNSServerConfFile, DNSServerConfFile, DNSServerConfFile)).SetContext(ctx)
+		if !req.WithoutNS {
```
cwen0, line 49: Why not add a `EnterNS` filed instead of `WithoutNS`?

WangXiangUSTC: the default value is false. if use EnterNS, I must modify every request in chaos mesh

STRRL: I prefer `EnterNS`(or further `EnterNSForContainer`), the default value is evil. 
I think it's necessary to declare explicitly that we need to enter the target namespaces of a specific container.

WangXiangUSTC: OK, I update `WithoutNS` to `EnterNS`

WangXiangUSTC:
@STRRL @cwen0 PTAL

pkg/chaosdaemon/ipset_server.go
```diff
@@ -49,41 +49,42 @@ func (s *DaemonServer) FlushIPSets(ctx context.Context, req *pb.IPSetsRequest) (
 	return &empty.Empty{}, nil
 }
 
-func flushIPSet(ctx context.Context, pid uint32, set *pb.IPSet) error {
+func flushIPSet(ctx context.Context, pid uint32, set *pb.IPSet, enterNS bool) error {
```
STRRL, line 52: I think the order of the parameters of all methods in this file should be reconstructed. Only when `enterNS` is true, `pid` is useful. So we'd better put `pid` and `enterNS` together. Furthermore, `enterNS` is required of `pid`, so `enterNS` should be ahead of `pid`. For example:

```golang
func flushIPSet(ctx context.Context, enterNS bool, pid uint32, set *pb.IPSet) error 
```

pkg/chaosdaemon/iptables_server.go
```diff
@@ -40,13 +40,13 @@ func (s *DaemonServer) SetIptablesChains(ctx context.Context, req *pb.IptablesCh
...
 	iptables := buildIptablesClient(ctx, pid)
-	err = iptables.initializeEnv()
+	err = iptables.initializeEnv(req.EnterNS)
 	if err != nil {
 		log.Error(err, "error while initializing iptables")
 		return nil, err
 	}
 
-	err = iptables.setIptablesChains(req.Chains)
+	err = iptables.setIptablesChains(req.Chains, req.EnterNS)
```
STRRL, line 49: We saw that struct `iptablesClient` already has a state called `pid`, so we'd better also put `enterNS` into `iptablesClient`.

pkg/chaosdaemon/tc_server.go
```diff
@@ -85,7 +85,7 @@ func (s *DaemonServer) SetTcs(ctx context.Context, in *pb.TcsRequest) (*empty.Em
 	setDefaultTcsRequest(in)
 
 	tcCli := buildTcClient(ctx, pid)
-	err = tcCli.flush(in.Device)
+	err = tcCli.flush(in.Device, in.EnterNS)
```
STRRL, line 88: Same issue with `iptablesClient`

pkg/chaosdaemon/tc_linux.go
```diff
@@ -20,7 +20,7 @@ import (
 	"github.com/chaos-mesh/chaos-mesh/pkg/mock"
 )
 
-func applyTc(ctx context.Context, pid uint32, args ...string) error {
+func applyTc(ctx context.Context, pid uint32, enterNS bool, args ...string) error {
```
STRRL, line 23: Same issues with parameters order in `ipset_server.go`

STRRL (changes requested):
we can make the function signatures more clear.

PTAL @WangXiangUSTC 😁

WangXiangUSTC:
all comments addressed, PTAL again @STRRL

STRRL (approved):
LGTM

cwen0 (approved):
LGTM

PR #1381, opened 2021-01-11
chaos-daemon: support setting node selector

cwen0:
 

Need to set nodeSelector to define which nodes can install chaos-daemon

Update helm charts 

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

image

set nodeSelector and helm upgrade: 

image

### DIFF
--- a/helm/chaos-mesh/templates/chaos-daemon-daemonset.yaml
+++ b/helm/chaos-mesh/templates/chaos-daemon-daemonset.yaml
@@ -89,10 +89,6 @@ spec:
               hostPort: {{ .Values.chaosDaemon.grpcPort }}
             - name: http
               containerPort: {{ .Values.chaosDaemon.httpPort }}
-      {{- with .Values.chaosDaemon.tolerations }}
-      tolerations:
-{{ toYaml . | indent 8 }}
-      {{- end }}
 {{- if .Values.bpfki.create }}
         - name: bpfki
           image: {{ .Values.bpfki.image }}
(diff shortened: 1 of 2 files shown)


helm/chaos-mesh/values.yaml
```diff
@@ -111,8 +111,12 @@ chaosDaemon:
     #   cpu: 250m
     #   memory: 512Mi
 
+  nodeSelector: {}
```
YangKeao, line 114: What is the behavior of empty `nodeSelector`?

WangXiangUSTC (approved):
LGTM

helm/chaos-mesh/templates/chaos-daemon-daemonset.yaml
```diff
@@ -140,3 +136,15 @@ spec:
           hostPath:
             path: /usr/src
 {{- end }}
+      {{- with .Values.chaosDaemon.nodeSelector }}
+      nodeSelector:
+{{ toYaml . | indent 8 }}
+      {{- end }}
+      {{- with .Values.chaosDaemon.affinity }}
+      affinity:
```
STRRL, line 144: I think daemonset chaos-daemon do not need affinity/anti-affinity. It will only runs one instance on one node.

STRRL (commented):
I think we should resolve this situation:  chaos might want to be injected into the pod which belongs to a node with no "chaos-daemon"

fewdan (approved):
LGTM

YangKeao (approved):
LGTM

PR #1392, opened 2021-01-13
pack chaosctl in controller-manager image

YangKeao:

Pack chaosctl in controller-manager's image, so that the users can use them without downloading extra binary.

And we have met a situation where the users failed to run `chaosctl` on his machine, but we haven't found the problem :( . Preparing a stable environment to run our software is always better than running it in a unpredictable and unreproducible environment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -117,10 +117,6 @@ endif
 watchmaker:
 	$(CGO) build -ldflags '$(LDFLAGS)' -o bin/watchmaker ./cmd/watchmaker/...
 
-# Build chaosctl
-chaosctl:
-	$(GO) build -ldflags '$(LDFLAGS)' -o bin/chaosctl ./cmd/chaosctl/*.go
-
 # Run against the configured Kubernetes cluster in ~/.kube/config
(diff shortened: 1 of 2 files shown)


YangKeao:
@STRRL The problem is that it failed to get the correct kubernetes api server host inside the container.

It tried to connect `localhost:8080` and failed...

STRRL:
> @STRRL The problem is that it failed to get the correct kubernetes api server host inside the container.
> 
> It tried to connect `localhost:8080` and failed...

Oh, let me take a look!

STRRL:
It's a bug, I will fix it ASAP.

YangKeao:
I close this PR, because running `chaosctl` in the controller-manager requires a lot more privilege :cry:  .

STRRL:
For more details:  we need kubernetes permission with "create" on "pods/portforward".

If one day we implement any other way to communicate with chaos-daemon(for example: by service), we could reopen this PR again.

Delivery `chaosctl` with our image is very convenient.

PR #1418, opened 2021-01-18
Workflow: docs, interfaces and serial scheduler

STRRL:

First PR for Workflow. 🎉

Rendered markdown of document: [link](https://github.com/STRRL/chaos-mesh/blob/workflow-step-by-step/workflow/docs/design.md).

Do not scare about the line of codes. I think it's not a big PR, most of the addition is error handling and documents.

Create new concepts for workflow:

- models: node, template, workflow.
- schedulers: serial_scheduler.
- other codes about error handling.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

I will append mock codes and unit tests in the next PR.

Side effects

Related changes

```release-note
NONE
```


pkg/workflow/engine/model/workflow/workflow.go
```diff
@@ -0,0 +1,53 @@
...
+// limitations under the License.
+
+package workflow
+
+import (
+	"github.com/chaos-mesh/chaos-mesh/pkg/workflow/engine/model/node"
+	"github.com/chaos-mesh/chaos-mesh/pkg/workflow/engine/model/template"
+)
+
+type WorkflowSpec interface {
```
Colstuwjx, line 21: How we validate the workflow spec?

pkg/workflow/engine/model/node/node.go
```diff
@@ -0,0 +1,50 @@
...
+
+type Node interface {
+	GetName() string
+	GetNodePhase() NodePhase
+	GetParentNodeName() string
+	GetTemplateName() string
+}
+
+// FIXME: remove this interface, it's should belongs to node status
+type NodeTreeNode interface {
```
Colstuwjx, line 39: How we know the workflow instance real-world execution order? By tree node naming?

Colstuwjx (commented):
Great Job! I asked a few questions about the implement 😁

pkg/workflow/engine/model/template/network_chaos.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"time"
+
+	chaosmeshv1alph1 "github.com/chaos-mesh/chaos-mesh/api/v1alpha1"
+	"github.com/chaos-mesh/chaos-mesh/pkg/workflow/engine/errors"
+)
+
+type NetworkChaosTemplate interface {
+	Template
+	FetchChaosNamePrefix() string
```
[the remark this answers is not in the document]
STRRL: It's my personal habit. 🙈 
I usually use `Get` prefix for returning one field(with the same name) directly, and use `Fetch` for returning some value that needs a little logic to modify/calculate.

I will remove `Get` or `Fetch` prefix.

pkg/workflow/engine/model/template/pod_chaos.go
```diff
@@ -0,0 +1,36 @@
...
+package template
+
+import (
+	"time"
+
+	chaosmeshv1alph1 "github.com/chaos-mesh/chaos-mesh/api/v1alpha1"
+	"github.com/chaos-mesh/chaos-mesh/pkg/workflow/engine/errors"
+)
+
+type PodChaosTemplate interface {
```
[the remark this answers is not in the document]
STRRL: I indeed want to create `XXXChaosTempalteSpec`, like `PodSpec` and `PodTemplateSpec`, and generate different static codes for each type of Chaos.

Thanks for your suggestion! It seems another way for "schemaless". I will take a look!

pkg/workflow/engine/model/workflow/workflow.go
STRRL: Any implementation of `WorkflowSpec` should guarantee correctness by itself. 

As an implementation with Kubernetes, it should be resolved by webhook. (WIP)

pkg/workflow/engine/model/node/node.go
STRRL: I didn't think about this question before. There should be a field called `create time` in the `Node`.

Only `Serial` guarantees the execution order, it represents as nodes order in the parent's `children` field.

STRRL:
The current design with state-machine is against https://github.com/kubernetes/community/blob/master/contributors/design-proposals/architecture/principles.md. 😰

We should rethink using condition instead of phase & event.

Colstuwjx:
> We should rethink using condition instead of phase & event.

Could you share more details about this refactoring? As far as I know, argo also has maintained `Phase` in its `Status` for controlling node transition:

image

Does it mean that argo's design is also an anti-pattern according to the k8s community doc?

STRRL:
> Does it mean that argo's design is also an anti-pattern according to the k8s community doc?

Yes.

But we are still thinking about whether this pattern is suitable for our application(workflow) or not. 🤔

TLDR; 
- I think we do NOT need to follow this pattern strictly. 
- The state-machine style implementation will exist at least for a few "y"(in x.y.z version style) release versions of chaos-mesh. 
- These patterns are very helpful for build a High Available application. 😉

OK, here we go.

Why Kubernetes design principle asks for:
- "level-based"/"level-driven", not phase and event
- Object status must be 100% reconstructable by observation
- Cluster-wide invariants are difficult to enforce correctly (it's so important)

To figure out this question, let's take a look at what Kubernetes want to do. Kubernetes could orchestrate containers for the various workloads, Network, Storage, Application's Configuration also could be orchestrated by kubernetes. So it hides so much logic about "how to configure these things correctly", "how to fallback/recover things from unexpected situations".

System operations on computing devices are not easy like a database transaction: it goes to stable after "commit".  Operations might fail for so many weird reasons. So Kubernetes assumes that "Cluster-wide invariants are difficult to enforce correctly", which means controllers should try again and again to fix the current state to the desired state.

So I think these patterns are very useful for modifications on infrastructure.

But for workflow, actually, we use Kubernetes just like a KV storage which supports trigger-on-changes. Our CRD does not mirror/represent any physical entity in the world. Its status will not change unless our controller changes it.

workflow/docs/design.md
```diff
@@ -0,0 +1,406 @@
...
+
+### Others
+
+- Persistent Repo or just "repo", fetches and updates status for Workflow and Node.
+
+Detail about instating template:
+
+After calling `Scheduler.ScheduleNext()`, it will return
+
+- which template should be instated
```
WangXiangUSTC, line 311: what is the meaning of `instated`, is it a typo?

workflow/docs/design.md
```diff
@@ -0,0 +1,406 @@
...
+
+> Do we really need a tree? How about a pipeline?
+
+## Core Concepts
+
+**Workflow**: a resource that defines the orchestration of chaos experiments.
+
+**Template**: the node in **Workflow** "tree", represents for operation. There are various operations: like "create chaos experiments", "wait for 5 minutes", "create a serial job with children templates", etc.
+
+**Node**: a running/completed step within a workflow, it represents the status; We could think of it as an instance rendered from a **Template**.
```
WangXiangUSTC, line 77: It is said that `Template` is the node in **Workflow** "tree", so what is the relationship between `Template` and `Node`?

workflow/docs/design.md
STRRL: Yes, it's a typo. I will fix it right now.

workflow/docs/design.md
STRRL: The relationship between `Template` and `Node` just like `Class` and `Object` in OOP.

pkg/workflow/engine/errors/jsonify_error.go
```diff
@@ -0,0 +1,43 @@
...
+		if wrapper, ok := err.(xerrors.Wrapper); ok {
+			return fmt.Sprintf(
+				"failed to jsonify error on type %s, json error: %s; origin error message: %s",
+				reflect.TypeOf(origin).Name(),
+				err,
+				wrapper.Unwrap().Error(),
+			)
+		}
+		return fmt.Sprintf(
+			"failed to jsonify erro on type %s, json error, %s; also failed to Unwrap() on it.",
```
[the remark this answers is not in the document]
STRRL: Fixed! PTAL

workflow/docs/design.md
STRRL: Fixed! PTAL

WangXiangUSTC (approved):
LGTM

workflow/docs/design.md
```diff
@@ -0,0 +1,406 @@
...
+  - This Node did not make "Effect" yet.
+- For Serial:
+  - This Node did not create child node yet;
+  - Or previous child node just succeed;
+
+**WaitingForChild** is only available on Serial, Parallel and Task, it means at least 1 child node is in **Running** state.
+
+**Running** is available on type Chaos, Suspend, Task; It means an **Actor** is doing dirty work for this node. For Chaos, both of "create Chaos CRD resource" and "delete Chaos CRD resource" are presented as **Running**.
+
+**Evaluating** is only available on Task, it means Task is collecting result of user's pod, then picks templates to execute.
```
cwen0, line 218: I think we can delete this phase. For Task, evaluation maybe is a step of the task, and I think we also can use `Running` as the phase of the node when collecting the result of user's pod.

workflow/docs/design.md
```diff
@@ -0,0 +1,406 @@
...
+            // TODO: manually ask for next trigger
+          }
+        }()
+      }
+    }
+  }
+}
+```
+
+### States for Node
```
cwen0, line 191: This design combines logic tasks, chaos tasks, and etc into one state machine, which makes the state machine more complicated. How about splitting them into two or more state machines?  Such as logic node machine and execute node machine.

workflow/docs/design.md
```diff
@@ -0,0 +1,406 @@
...
+
+```go
+type OperableTrigger interface {
+	Trigger
+	Notify(event Event) error
+	NotifyDelay(event Event, delay time.Duration) error
+}
+```
+
+- `Notify` as `Enqueue`
```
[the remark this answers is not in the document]
STRRL: Actually what we need is only `Trigger`, and `Trigger` has only one method `Acquire()`.

`OperableTrigger` is just like implementation for common purpose `Trigger`, it works like a pipe/channel. 

The reason why `OperableTrigger` is an interface, not a struct is I didn't write the codes for this implementation during designing yet.

workflow/docs/design.md
```diff
@@ -0,0 +1,406 @@
+# Internal Design of Chaos Mesh Workflow
```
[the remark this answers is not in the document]
STRRL: I also think this file should be moved to another directory. 🙈

But I'm not sure about `cmd/workflow-engine`. Since the last talk with @YangKeao, we prefer workflow should be a part of "controller-manager", so we will not create another binary. How do you think about it?

And I have no idea where to place this file.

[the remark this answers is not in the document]
Colstuwjx: I don't think so. Design doc lived in this dir should be fine, once it gets ready to be public, it would be better to move into [website repo](https://github.com/chaos-mesh/website/tree/master/docs), and each feature owner should take ownership, see an [example](https://github.com/kubernetes/website/pull/16018) of EndpointSlice feature from k8s website.

workflow/docs/design.md
STRRL: Yes, it is just like a mess. We should reconsider how to recognize codes about different types of state-machine.

I will append this part in the next PRs.

workflow/docs/design.md
STRRL: I prefer to keep this phase. But it indeed brings some confusion with `WaitingForSchedule`. I realize that we might need evaluating for every type of node which needs schedule sub-nodes, and a node in `Evaluating` means "the scheduler is running".
Currently, the life cycle of a `Task` likes this:

`Init` -> `Running` -> `WaitingForChild`(The User's Pod) -> `Evaluating` -> `WaitingForSchedule` -> `WaitingFotChild`(Sub Nodes) -> `Succeed`

When a `Task` node is in phase `Evaluating`, it means the user's job has exited with success, and we are making a decision about which templates should be executed, then write these templates' name into the storage. It aims to make `WaitingForChild` before `Evaluating` and `WaitingForSchedule` after `Evaluating` be simple and observable.

pkg/workflow/engine/model/template/pod_chaos.go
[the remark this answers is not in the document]
YangKeao: Code generation or schemeless is both OK for me. Actually, I prefer code generation (as it can provide richer type information)

pkg/workflow/engine/model/template/template.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+type Template interface {
+	Name() string
+	TemplateType() TemplateType
+}
+
+// func IsCompositeType. CompositeType means this Template could have children Templates.
+func (it TemplateType) IsCompositeType() bool {
+	return it == Serial || it == Parallel || it == Task
```
YangKeao, line 41: Why do we need this function?

workflow/docs/design.md
```diff
@@ -0,0 +1,406 @@
...
+
+**Holding** is available on Chaos, Suspend; It means current node is waiting for next action. For example: a Chaos node which in **Holding** is waiting for the end of this ChaosExperiments, then delete it.
+
+**Succeed** means this node is completed.
+
+**Failed** means this node failed. One **Failed** node will cause the whole Workflow fall into **Failed**.
+
+Examples:
+
+A **NetworkChaos** Node: **Init** -> **Running** -> **Holding** -> **Running** -> **Succeed**
```
YangKeao, line 230: What will happen if the user delete the created `NetworkChaos` manually? Will it break the status machine?

workflow/docs/design.md
```diff
@@ -0,0 +1,406 @@
...
+
+**WaitingForSchedule** is only available on Serial, Parallel and Task, it means this Node is idle and safe for next scheduling; It is presents:
+
+- For Chaos, Suspend, Task, Parallel:
+  - This Node did not make "Effect" yet.
+- For Serial:
+  - This Node did not create child node yet;
+  - Or previous child node just succeed;
+
+**WaitingForChild** is only available on Serial, Parallel and Task, it means at least 1 child node is in **Running** state.
```
YangKeao, line 214: What if the child node is `Holding`?

STRRL: `WaitingForChild` also works for children nodes `Holding`. I will update the doc.

workflow/docs/design.md
STRRL: I did not think of that before. 🤔 

Because of the pattern of kubernetes controller, I think workflow SHOULD re-create a new `NetworkChaos` CRD.

I will append for this in future PRs.

workflow/docs/design.md
STRRL: fixed, PTAL

YangKeao (approved):
LGTM

WangXiangUSTC (approved):
LGTM

PR #1441, opened 2021-01-21
[update] update container_kill chaos status

fewdan:
If at least one pod does not have a specified container in the target pods, the state of chaos becomes failed.

If at least one pod does not have a specified container in the target pods, return the error info.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


controllers/podchaos/containerkill/types.go
```diff
@@ -96,6 +98,10 @@ func (r *endpoint) Apply(ctx context.Context, req ctrl.Request, obj v1alpha1.Inn
 		return err
 	}
 
+	if allHaveContainer == false {
+		return fmt.Errorf("at least one pod contains no designated container")
+	}
+
 	podchaos.Status.Experiment.PodRecords = make([]v1alpha1.PodStatus, 0, len(pods))
 	for _, pod := range pods {
```
fewdan, line 106: I have a doubt: if some pods have a container and some pods do not have a container in the selected pods, at this time, if the above returns directly, the corresponding pod_records will not be recorded. Is this reasonable? Or should you record pod_records and then "return fmt.Errorf("at least one pod contains no designated container")"?

controllers/podchaos/containerkill/types.go
```diff
@@ -88,6 +89,7 @@ func (r *endpoint) Apply(ctx context.Context, req ctrl.Request, obj v1alpha1.Inn
 		}
 
 		if haveContainer == false {
+			allHaveContainer = false
 			r.Log.Error(nil, fmt.Sprintf("the pod %s doesn't have container %s", pod.Name, podchaos.Spec.ContainerName))
```
cwen0, line 93: We should return this error here

controllers/podchaos/containerkill/types.go
cwen0: Yes, I think we should record this pod_records.

controllers/podchaos/containerkill/types.go
fewdan: @cwen0
Then there will be a situation that is, if we have pod1, pod2, and there is a specified container in pod2, if we traverse pod1 first, and then return an error directly here, it is possible that pod2 is not processed.

STRRL:
Hi @fewdan, I prefer to complete the checking of `allHaveContainers` **before killing some containers**.

Current implementation will bring a mess, for example:

I have 2 pods that have two containers so-called "web-app" and "monitoring",  and we have another 3 pods that have two containers so-called "another-web-app" and "monitoring". All 5 pods have one label "app: web".

Then I publish a container kill chaos, which wants to kill the "web-app" container, with labelSelector "app: web".

Obviously, the last 3 pods do not contain the container called "web-app", so this chaos experiment must be failed.

How do we resolve the first 2 pods which indeed contain the "web-app" container? The current implementation will kill these containers, and the Chaos Experiment falls to failure, emphasize again, the chaos failed, but something bad (the containers have been killed) already happened.

I prefer to make a check at first, if any pods mismatch with the required container, the chaos experiment goes failed, and no containers will be killed

cwen0:
> Hi @fewdan, I prefer to complete the checking of `allHaveContainers` **before killing some containers**.
> 
> Current implementation will bring a mess, for example:
> 
> I have 2 pods that have two containers so-called "web-app" and "monitoring", and we have another 3 pods that have two containers so-called "another-web-app" and "monitoring". All 5 pods have one label "app: web".
> 
> Then I publish a container kill chaos, which wants to kill the "web-app" container, with labelSelector "app: web".
> 
> Obviously, the last 3 pods do not contain the container called "web-app", so this chaos experiment must be failed.
> 
> How do we resolve the first 2 pods which indeed contain the "web-app" container? The current implementation will kill these containers, and the Chaos Experiment falls to failure, emphasize again, the chaos failed, but something bad (the containers have been killed) already happened.
> 
> I prefer to make a check at first, if any pods mismatch with the required container, the chaos experiment goes failed, and no containers will be killed

I agree with this idea.

fewdan:
> Hi @fewdan, I prefer to complete the checking of `allHaveContainers` **before killing some containers**.
> 
> Current implementation will bring a mess, for example:
> 
> I have 2 pods that have two containers so-called "web-app" and "monitoring", and we have another 3 pods that have two containers so-called "another-web-app" and "monitoring". All 5 pods have one label "app: web".
> 
> Then I publish a container kill chaos, which wants to kill the "web-app" container, with labelSelector "app: web".
> 
> Obviously, the last 3 pods do not contain the container called "web-app", so this chaos experiment must be failed.
> 
> How do we resolve the first 2 pods which indeed contain the "web-app" container? The current implementation will kill these containers, and the Chaos Experiment falls to failure, emphasize again, the chaos failed, but something bad (the containers have been killed) already happened.
> 
> I prefer to make a check at first, if any pods mismatch with the required container, the chaos experiment goes failed, and no containers will be killed

ok, I think it's a good idea, i'll modify it.

cwen0 (approved):
LGTM

controllers/podchaos/containerkill/types.go
```diff
@@ -64,17 +64,38 @@ func (r *endpoint) Apply(ctx context.Context, req ctrl.Request, obj v1alpha1.Inn
...
 
+		for containerIndex := range pod.Status.ContainerStatuses {
+			containerName := pod.Status.ContainerStatuses[containerIndex].Name
+			if containerName == podchaos.Spec.ContainerName {
+				haveContainer = true
+				break
+			}
+		}
+
+		if haveContainer == false {
```
STRRL, line 80:
```suggestion
		if !haveContainer {
```

controllers/podchaos/containerkill/types.go
```diff
@@ -64,17 +64,38 @@ func (r *endpoint) Apply(ctx context.Context, req ctrl.Request, obj v1alpha1.Inn
...
 		pod := &pods[podIndex]
 		haveContainer := false
 
+		for containerIndex := range pod.Status.ContainerStatuses {
+			containerName := pod.Status.ContainerStatuses[containerIndex].Name
+			if containerName == podchaos.Spec.ContainerName {
+				haveContainer = true
+				break
+			}
+		}
```
STRRL, line 78: I prefer that:
```suggestion
		for _, item := range pod.Status.ContainerStatuses {
			if item.Name == podchaos.Spec.ContainerName {
				haveContainer = true
				break
			}
		}
```

controllers/podchaos/containerkill/types.go
```diff
@@ -64,17 +64,38 @@ func (r *endpoint) Apply(ctx context.Context, req ctrl.Request, obj v1alpha1.Inn
 		return err
 	}
 
-	g := errgroup.Group{}
+	podsNotHaveContainer := []string{}
```
STRRL, line 67:
```suggestion
	var podsNotHaveContainer []string
```

STRRL (changes requested):
@fewdan PTAL

controllers/podchaos/containerkill/types.go
fewdan: Okay, let me revise the above question.

STRRL (approved):
LGTM

cwen0 (approved):

PR #1444, opened 2021-01-22
docker client return error if the pid is zero

YangKeao:

(I have forgotten to fix this bug for a long time :crying_cat_face:, sorry )

If the pid of a docker container is 0 (when the container is not running), an error will be returned.


WangXiangUSTC (approved):
LGTM

STRRL (approved):
LGTM

PR #1451, opened 2021-01-26
fix yaml generate

YangKeao:

As the `api/v1alpha1` is a standalone module now, it won't be detected and codes will not be generated until we `cd` into it :angry: 

Now `make yaml` and `make generate` would work well.


WangXiangUSTC (approved):
LGTM

cwen0 (approved):
LGTM

Issue #1467, opened 2021-01-29
Get an empty value when LabelField loses focus


shaowenchen:
## Bug Report

![](https://raw.githubusercontent.com/shaowenchen/testing/main/img/1.gif)

**What version of Kubernetes are you using?**

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.8", GitCommit:"9f2892aab98fe339f3bd70e3c470144299398ace", GitTreeState:"clean", BuildDate:"2020-08-13T16:12:48Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.9", GitCommit:"4fb7ed12476d57b8437ada90b4f93b17ffaeed99", GitTreeState:"clean", BuildDate:"2020-07-15T16:10:45Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

**What version of Chaos Mesh are you using?**

master

PR #1468, opened 2021-01-29
fix: unhandled empty value in LabelField

shaowenchen:
Signed-off-by: shaowenchen <mail@chenshaowen.com>

Resolve https://github.com/chaos-mesh/chaos-mesh/issues/1467.


g1eny0ung:
Good catch! But I think it's better to check first, like:

```js
const t = text.trim()

// Here
if (t === '') {
  return
}

if (isKV && !/^.+:.+$/.test(t)) {
  setError('Invalid key:value format')

  return
}
```

shaowenchen:
> Good catch! But I think it's better to check first, like:
> 
> ```js
> const t = text.trim()
> 
> // Here
> if (t === '') {
>   return
> }
> 
> if (isKV && !/^.+:.+$/.test(t)) {
>   setError('Invalid key:value format')
> 
>   return
> }
> ```

Adopted

g1eny0ung (approved):
LGTM

fewdan (approved):
LGTM

Issue #1472, opened 2021-01-30
Enriching AWS chaos


STRRL:
Here is an overview of the plan for supporting more chaos about AWS.

We have already try to implement AWS Chaos like "stop/restart" EC2, but it still lacks e2e tests. Thanks to localstack, we could emulation the AWS API at the local machine. So we could write some test cases based on localstack.

Except for e2e tests, I think we could implement more chaos about AWS based services, such as S3 or DynamoDB.

fewdan:
I think this is a very good idea!

someshkoli:
Hey @STRRL, I'd like to work on this issue


> Here is an overview of the plan for supporting more chaos about AWS.
> 
> Except for e2e tests, I think we could implement more chaos about AWS based services, such as S3 or DynamoDB.

From what I understand, We need to use the api that awscli uses to create more chaos on aws services such just like we do on EC2 ? 

> 
> We have already try to implement AWS Chaos like "stop/restart" EC2, but it still lacks e2e tests. Thanks to localstack, we could emulation the AWS API at the local machine. So we could write some test cases based on localstack.

Not yet clear about this, could you please elaborate more ?

STRRL:
> We need to use the api that awscli uses to create more chaos on aws services such just like we do on EC2?

It's the way that we used to implement AWS Chaos for now. We invoke the official AWS API for making some chaos. It indeed makes some breaks on the real machines.

If we could, we want to try another way to make more chaos: use localstack instead of official AWS API. You might already notice that `awscli` could connect to another self-hosted localhost by providing `--endpoint`. 

We might run a **modified** localstasck that could provide some internal API for simulation chaos. But it's just an imagination for injecting chaos into "self-hosted cloud infrastructure". 

Honstly, if we modify the localstack, these chaoses might not work on AWS, it might be useless. We will figure out the feasibility and usability in several days.

STRRL:
And this issue has been taken as an lfx mentorship project, you could look up [here](https://mentorship.lfx.linuxfoundation.org/project/1d49ebad-29fa-430c-8492-44f29ee61cc1). 😜

someshkoli:
> > We need to use the api that awscli uses to create more chaos on aws services such just like we do on EC2?
> 
> It's the way that we used to implement AWS Chaos for now. We invoke the official AWS API for making some chaos. It indeed makes some breaks on the real machines.
> 
[...]

PR #1477, opened 2021-02-01
awschaos: add integration test with localstack

WangXiangUSTC:

part of issue https://github.com/chaos-mesh/chaos-mesh/issues/1472

add an integration for awschaos to stop ec2.

### Checklist

Tests

- [x] Integration test
- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


fewdan (approved):
LGTM

test/integration_test/aws/run.sh
```diff
@@ -0,0 +1,73 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eu
+
+cur=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+cd $cur
+
+# wait localstash pod status to running
```
cwen0, line 21: We don't need to generate a temp file `pods.status`. The following command can be used to count the number of pods that are not running.  

```
kubectl get pods -n "${namespace}" ${labels} --no-headers | grep -v Running)
```

test/integration_test/aws/run.sh
```diff
@@ -0,0 +1,73 @@
...
+LOCALSTACK_SERVER="http:\/\/$NODE_IP\:$NODE_PORT"
+
+aws configure set aws_access_key_id test
+aws configure set aws_secret_access_key test
+aws configure set default.region us-east-1
+aws configure set default.output_format text
+
+echo "run ec2 instance, and the state is pending, will switch to running later"
+aws --endpoint-url=http://127.0.0.1:4566 ec2 run-instances --image-id ami --count 1 --instance-type t2.micro --key-name test > run_instance.log
+check_contains "pending" run_instance.log
```
cwen0, line 48: I didn't found the definition of `check_contains`.  Where is it?

test/integration_test/aws/run.sh
```diff
@@ -0,0 +1,73 @@
...
+INSTANCE_ID=`cat run_instance.log | grep "InstanceId" | sed 's/.*\"InstanceId\": \"\([0-9,a-z,-]*\)\",/\1/g' | xargs`
+
+sleep 2
```
cwen0, line 51: Why does this script need to exec `sleep 2` here?  Can we use another way instead of `sleep 2`?  `sleep` action is an indeterminate behavior, and we cannot guarantee that 2 seconds will be sufficient.

cwen0: If there is no other good way to instead it, sleep operation is also acceptable for me.

test/integration_test/aws/run.sh
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/env bash
+
+# Copyright 2020 Chaos Mesh Authors.
```
cwen0, line 3: s/2020/2021

test/integration_test/aws/run.sh
WangXiangUSTC: done in https://github.com/chaos-mesh/chaos-mesh/pull/1477/commits/e39ff58e5a3f906734dd712c2e21ec87bcbc83e2

test/integration_test/aws/run.sh
WangXiangUSTC: there is a script file https://github.com/chaos-mesh/chaos-mesh/blob/master/test/integration_test/utils/check_contains

test/integration_test/aws/run.sh
WangXiangUSTC: In most case 2 seconds is enough, and need to do some retry if remove the `sleep`. We can keep it now, and maybe modify it later if necessary.

test/integration_test/aws/run.sh
```diff
@@ -0,0 +1,70 @@
...
+LOCALSTACK_SERVER="http:\/\/$NODE_IP\:$NODE_PORT"
+
+aws configure set aws_access_key_id test
+aws configure set aws_secret_access_key test
+aws configure set default.region us-east-1
+aws configure set default.output_format text
+
+echo "run ec2 instance, and the state is pending, will switch to running later"
+aws --endpoint-url=http://127.0.0.1:4566 ec2 run-instances --image-id ami --count 1 --instance-type t2.micro --key-name test > run_instance.log
+check_contains "pending" run_instance.log
```
STRRL, line 45: I prefer to use `jq` select the specific field from the output json, then make an assertion with `test`, for example:

```bash
test $(aws --endpoint-url=http://127.0.0.1:4566 ec2 run-instances --image-id ami --count 1 --instance-type t2.micro --key-name test | jq -rM '.Instances[0].State.Name') = "pending"
```

test/integration_test/aws/run.sh
```diff
@@ -0,0 +1,70 @@
...
+    not_run_num=`kubectl get pods -l app.kubernetes.io/name=localstack --no-headers | grep -v Running | wc -l`
+
+    if [ $not_run_num == 0 ]; then
+        break
+    fi
+
+    sleep 3
+done
+
+kubectl port-forward svc/localstack 4566:4566 &
```
STRRL, line 32: Do we need to cleanup this port-forwarding process?

STRRL (commented):
rest LGTM, PTAL @WangXiangUSTC

test/integration_test/aws/run.sh
WangXiangUSTC: done in 69dcf65

test/integration_test/aws/run.sh
WangXiangUSTC: done in 69dcf65

STRRL (approved):
LGTM

fewdan (approved):

PR #1481, opened 2021-02-02
Cronjob (twophase scheduler) through standalone resource

YangKeao:

@STRRL 

While developing #1464 I have suffered a lot on struggling with scheduler and selector —— I have to manage the status to decide whether this time I need to reselect or rerun the scheduler, which is really complicated.

In this PR, I will try to develop a cronjob engine, which is expected to work for both `*Chaos` (with scheduler and duration), and the `CronWorkflow` in the future. In the newly created `cronjob` controller, a new common chaos is created every time it arrives the `nextStart`, and will be deleted when it arrives `nextRecover`. There are a lot of benefits:

1. Make it easier to develop dynamic configuration and selector related features in the future ( I have thought a lot about these features. I believe before I move on them, this PR should be merged first )
2. All `Apply` and `Recover` function calls are from `common` controller now, so that it shouldn't worry about the scheduler status. Especially when it's needed to enter the controller several times (e.g. in #1464, it has to enter the controller several times to check the status of podnetworkchaos)
3. It can share the same implementation with `CronWorkflow`, and could be a good solution before we remove `scheduler` and `duration` fields
4. The new controller is much easier to understand! It takes only 100-200 lines, while the old one takes 300+ lines.

Some fields of the parent will synchronous with the created one (like `Spec` and `Annotation` and `Label`), while some fields will synchronous back (like `.Status.Experiment`)

There are still several problems which I haven't solved:

1. Dynamically modify `scheduler` field. After reviewing the code in `twophase` controller, I thought it's hard to define a easy-to-understand behavior when modifying the `scheduler` field. Maybe we should use webhook to avoid modifying `scheduler` :anger: 
2. `PodKill` is still very special. It doesn't represent a status (e.g. the connection is blocked or the file system is unavailable). Creating a `PodKill` common chaos is really awkward. 

For these two problems, do you have any ideas?


STRRL:
> Some fields of the parent will synchronous with the created one (like Spec and Annotation and Label), while some fields will synchronous back (like .Status.Experiment)

I do not think sync things is a good idea, it makes `CronJob` and actual `Chaos` depends on each other. I prefer to use a `TemplateSpec` just like `CronJob` does to `Job` and `Deployment` does to `Pod`.

And I saw that the `CronJob` also extends the `StatefulObject`, which confuses me, I prefer the `CronJob` is not a chaos, it controls the chaos.

> * Dynamically modify `scheduler` field. After reviewing the code in `twophase` controller, I thought it's hard to define a easy-to-understand behavior when modifying the `scheduler` field. Maybe we should use webhook to avoid modifying `scheduler` 💢

I could not find out what's the complex thing for modifying `schduler` dynamically after we make up `CronJob`. Could you tell me more detail? 🥺

> * `PodKill` is still very special. It doesn't represent a status (e.g. the connection is blocked or the file system is unavailable). Creating a `PodKill` common chaos is really awkward.

I do not have any good idea about that now. Not only `PodKill`, but AWS Chaos also meet the same situation. There are lots of chaos do not use `duration`, and could not `Recover`.

For `PodKill`, I think both "killing once" and "killing pods continuously" should be supported, the common chaos `PodKill` makes sense. After we do not use `duration` in `*ChaosSpec` anymore, I think one object of `PodKill` CRD means one operation that killing the pod is acceptable.  😑

I think we should make more detail and expressive for describing the behavior of these chaoses, it means the API will change, and I'm not so sure about doing this.

YangKeao:
@STRRL 

> I prefer the CronJob is not a chaos
> I do not think sync things is a good idea, it makes CronJob and actual Chaos depends on each other.

Cannot agree more :beers: ! This bad design is for the compatibility :crying_cat_face:  (with existing `scheduler` and `duration` fields). In this PR, I won't break any existing chaos definition files. Without considering the compatibility, `TemplateSpec` should be the best way and I think your `CronWorkflow` should follow the best design :smile: (as it doesn't have any compatibility concern).

STRRL:
> In the newly created cronjob controller, a new common chaos is created every time it arrives the nextStart, and will be deleted when it arrives nextRecover.

If we do not append the new `CronJob` CRD, it brings me more confusing, how do we implement creating and deleting common chaos?

PR #1483, opened 2021-02-03
dynamically reload config from configMap

YangKeao:

However, this PR nearly doesn't make any progress :cry: .

1. I don't suggest users to modify the ConfigMap, as it will be overided by `helm upgrade`
2. Running `helm upgrade --set xxx=xxx` will cause a restart.

It doesn't make any difference. Without merging this PR, we can `helm upgrade --set xxx=xxx` to set a environment variable with restarting.

To achieve the goal to dynamically reload config, we should also dynamically reload certs. However, I think restarting to reload the config is good enough for us :crying_cat_face:. So whether this PR is needed, is still a thing to discuss.

====================

`viper` is full of bugs and misleading usages :angry:.


YangKeao:
This feature has been discussed and shouldn't be added to Chaos Mesh.

Issue #1486, opened 2021-02-03
NetworkChaos direction:Both has no affect 


Robbie558:
## Bug Report

**What version of Kubernetes are you using?**
```
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.1", GitCommit:"206bcadf021e76c27513500ca24182692aabd17e", GitTreeState:"clean", BuildDate:"2020-09-09T11:18:22Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
```

**What version of Chaos Mesh are you using?**
```
Controller manager Version: version.Info{GitVersion:"v1.1.1", GitCommit:"88e265c1ad2bcc3931645371a71683dd52b576fb", BuildDate:"2021-01-21T11:24:39Z", GoVersion:"go1.14.6", Compiler:"gc", Platform:"linux/amd64"}
```

**What did you do?**
We have the following `NetworkChaos` experiment spec and two pods of the target application running, but when running there appears no network loss at all:

```
spec:
  action: loss
  mode: all
  value: ''
  selector:
    namespaces:
      - <app_namespace>
    labelSelectors:
      app: <app_label_name>
  loss:
    loss: '100'
    correlation: '100'
  direction: both
```

Testing using separate `to` and `from` NetworkChaos experiment specs does appear to work for us.  The configuration used for these specs is as follows:

**From:**
```
spec:
  action: loss
  direction: from
  loss:
    correlation: "100"
    loss: "100"
  mode: all 
  selector:
    namespaces:
    - <app_namespace>
  target:
    mode: all 
    selector:
      labelSelectors:
        app: <app_label_name>
      namespaces:
      - <app_namespace>
    value: ""
  value: ""
```

In this we are limiting (non-target section `selector` option) on namespace as we wish to affect traffic from all pods in that given namespace going to our target pod.

**To:**
```
spec:
  action: loss
  direction: to
  loss:
    correlation: "100"
    loss: "100"
  mode: all 
  selector:
    labelSelectors:
      app: <app_label_name>
    namespaces:
    - <app_namespace>
  target:
    mode: all 
    selector:
      namespaces:
      - <app_namespace>
    value: ""
  value: ""

```

**What did you expect to see?**
Bi-directional network loss for pods 

**What did you see instead?**
No network loss in either direction for the target pods 

**Output of chaosctl**

Not captured

-----
[...]

PR #1489, opened 2021-02-04
add e2e test for bidirection netem

YangKeao:
An e2e test for #1486


test/e2e/chaos/networkchaos/network_delay.go
```diff
@@ -257,7 +261,45 @@ func TestcaseNetworkDelay(
...
+	)
+	By("Injecting both direction chaos for 0")
+	err = cli.Create(ctx, bothDirectionNetem.DeepCopy())
+	framework.ExpectNoError(err, "create network chaos error")
+	wait.Poll(time.Second, 15*time.Second, func() (done bool, err error) {
+		result = probeNetworkCondition(c, networkPeers, ports, true)
+		if len(result[networkConditionBlocked]) != 0 || len(result[networkConditionSlow]) != 2 {
+			return false, nil
+		}
+		return true, nil
```
STRRL, line 292: How about inverting this `if` statement.

```go
		if len(result[networkConditionBlocked]) == 0 && len(result[networkConditionSlow]) == 2 {
			return true, nil
		}
		return false, nil
```

STRRL (commented):
rest LGTM

WangXiangUSTC:
Is this pr should be merged after https://github.com/chaos-mesh/chaos-mesh/pull/1490?

STRRL:
> Is this pr should be merged after #1490?

I think it does not need after #1490.  It seems that this test case does not relate to #1490.

test/e2e/chaos/networkchaos/network_delay.go
YangKeao: I prefer to keep the same pattern with other test cases (as it's hard to notice the difference for the next time modifying the test case)

fewdan (approved):
LGTM

test/e2e/chaos/networkchaos/network_delay.go
STRRL: Oh, I just realize it. 👍

STRRL (approved):
LGTM

Issue #1450, opened 2021-01-26
unexpected behavior of NetworkChaos Netem


YangKeao:
```go
	case v1alpha1.Both:
		err = r.applyTc(ctx, pods, pods, externalCidrs, m, networkchaos)
		if err != nil {
			r.Log.Error(err, "failed to apply traffic control", "sources", pods, "targets", pods)
			return err
		}
```

When the direction is both, all connection between all pods will be applied with `netem`, which is not expected.

For example, if the source is A, B, and the target is C, D, now the connection between A and B will also be affected by this chaos, but what we actually need is to only apply chaos between AC, AD, BC, BD.

PR #1490, opened 2021-02-04
NetworkChaos: fix `both` direction has no effect

### LINKED ISSUE #1486: NetworkChaos direction:Both has no affect

cwen0:

Fix #1486 #1450 

If the target and external targets are both empty, we don't need to set `dstIpset`. 

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

image


YangKeao:
Setting direction to `To` (as the default value) if the target is empty seems to be a better idea. (because `from` will also cause a misleading behavior).

cwen0:
> Setting direction to `To` (as the default value) if the target is empty seems to be a better idea. (because `from` will also cause a misleading behavior).

If users set the direction to `both`, this action will change the definition of chaos experiment. I don't think this is not good behavior. About `from` will also cause a misleading behavior`, we should verify the chaos experiment when to apply it.

cwen0:
> > Setting direction to `To` (as the default value) if the target is empty seems to be a better idea. (because `from` will also cause a misleading behavior).
> 
> If users set the direction to `both`, this action will change the definition of chaos experiment. This behavior is not friendly for users. About `from will also cause a misleading behavior`, we should verify the chaos experiment when to apply it.

I will update the validatewebhook  to check it.

controllers/networkchaos/trafficcontrol/types.go
```diff
@@ -128,7 +128,10 @@ func (r *endpoint) Apply(ctx context.Context, req ctrl.Request, chaos v1alpha1.I
 			return err
 		}
 	case v1alpha1.Both:
-		err = r.applyTc(ctx, pods, pods, externalCidrs, m, networkchaos)
+		if len(targets) > 0 {
+			targets = pods
+		}
+		err = r.applyTc(ctx, pods, targets, externalCidrs, m, networkchaos)
```
STRRL, line 134: I have a question, I see that we resolve the "direction: both" with certain codes, it also bring another issue: https://github.com/chaos-mesh/chaos-mesh/issues/1450

Why not this:

```go
case v1alpha1.Both:
    err = r.applyTc(ctx, sources, targets, externalCidrs, m, networkchaos)
    // resolve error
    err = r.applyTc(ctx, targets, sources, []string{}, m, networkchaos)
    // resolve error
```

It means "direction: both" is the composition of both "direction: from" and "direction: to".

STRRL (commented):
PTAL @cwen0

controllers/networkchaos/trafficcontrol/types.go
cwen0: Great! I will fix it.

api/v1alpha1/networkchaos_webhook.go
```diff
@@ -138,16 +138,22 @@ func (in *NetworkChaos) ValidatePodMode(spec *field.Path) field.ErrorList {
...
 			field.Invalid(target.Child("direction"), in.Spec.Direction,
 				fmt.Sprintf("external targets cannot be used with `from` direction in netem action yet")))
 	}
 
+	if in.Spec.Direction == From && in.Spec.ExternalTargets == nil && in.Spec.Target == nil {
+		allErrs = append(allErrs,
+			field.Invalid(target.Child("direction"), in.Spec.Direction,
+				fmt.Sprintf("`from` direction cannot be used when targets and external targets are both empty")))
+	}
+
```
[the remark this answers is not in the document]
cwen0: This check is ok! I will update it.

STRRL (dismissed):
LGTM, and this PR could fix #1450. I will append e2e test cases soon.

controllers/networkchaos/trafficcontrol/types.go
YangKeao: This has another problem :crying_cat_face: . #1448 assumes that the ipset with the same name should have the same content. However, if you `applyTc` twice, they will have the same `ipset` name with different content.

```
dstIpset := ipset.BuildIPSet(targets, externalTargets, networkchaos, string(tcType)[0:5], m.Source)
```

The `postFix` should be modified from `string(tcType)[0:5]` to the `tgt`, `src` like the partition chaos

controllers/networkchaos/trafficcontrol/types.go
```diff
@@ -128,9 +128,15 @@ func (r *endpoint) Apply(ctx context.Context, req ctrl.Request, chaos v1alpha1.I
 			return err
 		}
 	case v1alpha1.Both:
-		err = r.applyTc(ctx, pods, pods, externalCidrs, m, networkchaos)
+		err = r.applyTc(ctx, sources, targets, externalCidrs, m, networkchaos)
```
YangKeao, line 131: There is another problem :crying_cat_face: . #1448 assumes that the ipset with the same name should have the same content. However, if you `applyTc` twice, they will have the same `ipset` name with different content.

```
dstIpset := ipset.BuildIPSet(targets, externalTargets, networkchaos, string(tcType)[0:5], m.Source)
```

The `postFix` should be modified from `string(tcType)[0:5]` to the `tgt`, `src` like the partition chaos

YangKeao (changes requested):

controllers/networkchaos/trafficcontrol/types.go
cwen0: Fixed

YangKeao: Why `toTargetIPSetPostFix` and `bothTargetIPSetPostFix` are different? Should we use the same pattern for these names?

controllers/networkchaos/trafficcontrol/types.go
```diff
@@ -283,8 +301,12 @@ func (r *endpoint) applyTc(ctx context.Context, sources, targets []v1.Pod, exter
 		return nil
 	}
 
+	if len(ipSetPostFix) == 0 {
+		ipSetPostFix = string(tcType[0:5])
+	}
+
 	// create ipset contains all target ips
-	dstIpset := ipset.BuildIPSet(targets, externalTargets, networkchaos, string(tcType)[0:5], m.Source)
+	dstIpset := ipset.BuildIPSet(targets, externalTargets, networkchaos, ipSetPostFix, m.Source)
```
STRRL, line 309: How about that:

```go
dstIpset := ipset.BuildIPSet(targets, externalTargets, networkchaos, string(tcType)[0:5] + ipSetPostFix, m.Source)
```

It avoids the modification for `ipSetPostFix`, and still shows the `tcType` in the ipset name whether `ipSetPostFix` is empty or not.

YangKeao: Maybe we could also shrink the length of `string(tcType)[0:5]`, as I know the only two possible `TcType` is `netem` and  `bandwidth`, and they are easy to distinguish

cwen0: `string(tcType)[0:2]` ?

STRRL (approved):
LGTM

YangKeao (approved):
LGTM

PR #1504, opened 2021-02-18
use tls mutual authentication for chaos-daemon

### LINKED ISSUE #1431: Security issue about the `chaos-daemon`
`chaos-daemon` doesn't have any authentication mechanism, which means any client in the cluster could modify the iptables and ipset config / cause stress / inject IO latency for nearly all the pods.

gRPC with server side SSL/TLS authentication is enough. However, it may break the `chaosctl` (which is definitely an evidence for the risk), or we need to distribute the certificate to it (or it can pull the certificate from `controller-manager` automatically).

YangKeao:

Fix #1431

Add TLS authentication for grpc connection to `chaos-daemon`.


pkg/chaosdaemon/server.go
```diff
@@ -44,6 +49,15 @@ type Config struct {
 	Host      string
 	Runtime   string
 	Profiling bool
+
+	TLSConfig
+}
+
+// TLSConfig contains the config of TLS Server
+type TLSConfig struct {
```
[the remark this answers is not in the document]
YangKeao: This `TLSConfig` is only used for simplifying the parameters (shrink three parameters into one). And I don't think depending on the controller config package is a good idea for `chaos-daemon`.

Maybe I should unexport this `TLSConfig`.

pkg/chaosdaemon/client/utils.go
```diff
@@ -16,6 +16,12 @@ package client
 import (
 	"context"
 
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/chaos-mesh/chaos-mesh/controllers/config"
```
WangXiangUSTC, line 23: how about formating this import?

pkg/chaosdaemon/server.go
```diff
@@ -15,9 +15,14 @@ package chaosdaemon
 
 import (
 	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
+	"io/ioutil"
 	"net"
 
+	"google.golang.org/grpc/credentials"
```
WangXiangUSTC, line 24: ditto

WangXiangUSTC (commented):
rest LGTM

WangXiangUSTC (approved):
LGTM

pkg/grpc/utils.go
```diff
@@ -37,33 +36,35 @@ var RPCTimeout = DefaultRPCTimeout
...
-	var endpoints v1.Endpoints
-	err := c.Get(ctx, types.NamespacedName{
-		Namespace: ns,
-		Name:      "chaos-daemon",
-	}, &endpoints)
-	if err != nil {
-		return nil, err
-	}
+// CreateGrpcConnectionWithAddress create a grpc connection with given port and address
+func CreateGrpcConnectionWithAddress(address string, port int, caCert string, cert string, key string) (*grpc.ClientConn, error) {
```
STRRL, line 40: `caCert`, `cert`, and `key` are paths of these files, please present this meaning in argument name, like `caCertFile`.

pkg/grpc/utils.go
```diff
@@ -37,33 +36,35 @@ var RPCTimeout = DefaultRPCTimeout
...
+	options := []grpc.DialOption{grpc.WithUnaryInterceptor(TimeoutClientInterceptor)}
 
-	daemonIP := findIPOnEndpoints(&endpoints, nodeName)
```
STRRL, line 55: We could remove method `findIPOnEndpoints()`.

pkg/chaosdaemon/client/utils.go
YangKeao: Fxied

pkg/chaosdaemon/server.go
YangKeao: Fixed

pkg/chaosdaemon/client/utils.go
```diff
@@ -41,16 +47,55 @@ func (c *GrpcChaosDaemonClient) Close() error {
...
+			}
+		}
+	}
+
+	return ""
+}
+
 // NewChaosDaemonClient would create ChaosDaemonClient
-func NewChaosDaemonClient(ctx context.Context, c client.Client, pod *v1.Pod, port int) (ChaosDaemonClientInterface, error) {
+func NewChaosDaemonClient(ctx context.Context, c client.Client, pod *v1.Pod) (ChaosDaemonClientInterface, error) {
```
STRRL, line 85: We need a little refactoring with `NewChaosDaemonClient` and `NewChaosDaemonClientLocally`:

This file is under package `pkg/chaosdaemon/client`, it should be the minimal client constructor of the chaos-daemon, so I do not think kuernetes and chaos-controller-manager things should be here.

I think we could:

- remove these two methods
- create a constructor of `GrpcChaosDaemonClient`
- If we need the origin `NewChaosDaemonClient`, we could write a similar one at `pkg/controllers`

helm/chaos-mesh/templates/chaos-daemon-daemonset.yaml
```diff
@@ -21,8 +21,9 @@ spec:
         app.kubernetes.io/name: {{ template "chaos-mesh.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
         app.kubernetes.io/component: chaos-daemon
-    {{- with .Values.chaosDaemon.podAnnotations }}
       annotations:
+        rollme: {{ randAlphaNum 5 | quote }}
```
STRRL, line 25: What's this annotation for? I do not see any reference in codes. 🙈

STRRL (changes requested):
rest LGTM 😁

helm/chaos-mesh/templates/chaos-daemon-daemonset.yaml
YangKeao: It's used to restart the `chaos-daemon` after upgrade/reinstall helm chart, so that it can reload the newly generated certificate

pkg/grpc/utils.go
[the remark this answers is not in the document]
YangKeao: Fixed

YangKeao: Fixed

pkg/grpc/utils.go
YangKeao: Fixed

pkg/chaosctl/common/common.go
```diff
@@ -309,7 +312,8 @@ func GetPidFromPod(ctx context.Context, pod v1.Pod, daemon v1.Pod) (uint32, erro
 		pfCancel()
 	}()
 
-	daemonClient, err := daemonClient.NewChaosDaemonClientLocally(int(localPort))
+	// TODO: support specify the cert file or get cert file automatically
+	daemonClient, err := NewChaosDaemonClientLocally(int(localPort), "", "", "")
```
YangKeao, line 316: @Yiyiyimu `chaosctl` will fail to connect to the chaos-daemon after this PR :crying_cat_face: 

Do you have any suggestions on how to specify the `ca`, `cert` and `key`? Should we add command line argument or use environment variable or get them directly from secrets?

Yiyiyimu: I'm not so sure that I couldn't say if anyone is better than the other two. 
It seems a bit redundant to use extra three arguments each time we use `chaosctl`. But if we do the settings once with arguments say `chaosctl config --ca=xx`, and then it would be the same with environment variables. Besides, I 'm not sure if directly get secrets from `controller-manager` would break the security.

Have you ever used any other command line tools which support tls? I guess maybe some references could be of help.

pkg/controllerutils/chaosdaemon.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/chaos-mesh/chaos-mesh/controllers/config"
```
STRRL, line 25: Good! The dependency is much cleaner now!

But I still prefer to move this file out of `/pkg`, place it at `/controller/something` 🙈

controllers/dnschaos/types.go
```diff
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/chaos-mesh/chaos-mesh/pkg/controllerutils"
+
```
STRRL, line 23: Please re-format/groupimport this line.

You might need a little manual work.

controllers/kernelchaos/types.go
```diff
@@ -18,6 +18,8 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/chaos-mesh/chaos-mesh/pkg/controllerutils"
+
```
STRRL, line 22: ditto, groupimport issue.

controllers/podchaos/containerkill/types.go
```diff
@@ -18,14 +18,15 @@ import (
 	"errors"
 	"fmt"
 
+	"github.com/chaos-mesh/chaos-mesh/pkg/controllerutils"
```
STRRL, line 21: ditto, groupimport issue.

controllers/podiochaos/types.go
```diff
@@ -19,15 +19,15 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/chaos-mesh/chaos-mesh/pkg/controllerutils"
```
STRRL, line 22: ditto, groupimport issue.

controllers/podnetworkchaos/ipset/ipset.go
```diff
@@ -17,15 +17,14 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/chaos-mesh/chaos-mesh/pkg/controllerutils"
```
STRRL, line 20: ditto, groupimport issue.

controllers/podnetworkchaos/iptable/iptable.go
```diff
@@ -17,20 +17,19 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/chaos-mesh/chaos-mesh/pkg/controllerutils"
```
STRRL, line 20: ditto, groupimport issue.

controllers/podnetworkchaos/tc/tc.go
```diff
@@ -17,18 +17,17 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/chaos-mesh/chaos-mesh/pkg/controllerutils"
```
STRRL, line 20: ditto, groupimport issue.

controllers/stresschaos/types.go
```diff
@@ -21,6 +21,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/chaos-mesh/chaos-mesh/pkg/controllerutils"
```
STRRL, line 24: ditto, groupimport issue.

controllers/timechaos/types.go
```diff
@@ -19,6 +19,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/chaos-mesh/chaos-mesh/pkg/controllerutils"
```
STRRL, line 22: ditto, groupimport issue.

STRRL (changes requested):
PTAL @YangKeao

pkg/chaosctl/common/common.go
Yiyiyimu: `viper` seems cool! I'll take a look at it~

STRRL: I suggest that we could implement this feature first. Then migration flag/configuration resolving to `viper`.

pkg/controllerutils/chaosdaemon.go
YangKeao: Nop. `/controller` contains only a lot of controllers (or "sub"controllers)

controllers/timechaos/types.go
YangKeao: Fixed

pkg/controllerutils/chaosdaemon.go
YangKeao: Oops. As it already has some other things, I will move this pkg to `/controller/`

YangKeao: Moved

STRRL (dismissed):
LGTM

YangKeao:
Wait! Let me test this PR on `cert-manager`

YangKeao:
> Wait! Let me test this PR on `cert-manager`

I have tested on `cert-manager` and done some modifications! Now `cert-manager` will self-sign a CA first and then sign other certificates through this CA :+1: 

Prepared for review :smile_cat:

STRRL (dismissed):
LGTM

pkg/config/controller.go
```diff
@@ -21,10 +21,23 @@ import (
...
+	// ChaosMeshCACert is the path of chaos mesh ca cert
+	ChaosMeshCACert string `envconfig:"CHAOS_MESH_CA_CERT" default:""`
+}
+
 // ChaosControllerConfig defines the configuration for Chaos Controller
 type ChaosControllerConfig struct {
 	// ChaosDaemonPort is the port which grpc server listens on
 	ChaosDaemonPort int `envconfig:"CHAOS_DAEMON_PORT" default:"31767"`
+
+	TLSConfig
```
cwen0, line 39: Can we use a pointer here?

cwen0 (commented):
Rest LGTM

pkg/config/controller.go
YangKeao: pointer is not good here. Because `pkg/grpc/utils` use `caCertPath`, `certPath` and `keyPath` three arguments to pass these fields, so that we have to do nil check outside the function call, which is a redundant behavior (than pass an empty value "")

But we cannot use `*TLSConfig` in `pkg/grpc/utils` directly, because depending on the `pkg/config` will trigger the `init` function and try to parse the config from env, which is not expected as a generally used package (rather than only used by controller).

YangKeao: Putting the logic of "getting config from env" and the config structure leads to this problem. I don't know what's the best practice in go to parse the config . Maybe it should be split in another pkg, which obviously should be done in another PR.

pkg/chaosdaemon/server_test.go
```diff
@@ -37,15 +37,15 @@ var _ = Describe("netem server", func() {
...
+			_, err := newGRPCServer(containerRuntimeContainerd, &MockRegisterer{}, tlsConfig{})
 			Expect(err).To(BeNil())
 		})
 
 		It("should panic", func() {
 			Ω(func() {
 				defer mock.With("MockContainerdClient", &MockClient{})()
 				defer mock.With("PanicOnMustRegister", "mock panic")()
-				newGRPCServer(containerRuntimeContainerd, &MockRegisterer{})
+				newGRPCServer(containerRuntimeContainerd, &MockRegisterer{}, tlsConfig{})
```
AsterNighT, line 48: Are there any tests on the new chaos-daemon grpc, with tls enabled? Or is it possible to introduce such tests?

YangKeao: There isn't any test even without TLS :cry:. I will file an issue to add test for chaos-daemon server and client (with/without tls).

YangKeao: #1536

AsterNighT:
Rest LGTM

AsterNighT (dismissed):

AsterNighT (dismissed):
LGTM

AsterNighT (approved):

pkg/config/controller.go
cwen0: Got it

cwen0 (approved):
LGTM

Issue #286, opened 2020-02-26
Write document/example about validation hook


YangKeao:
Add validation related part in [`dev_hello_world.md`](https://github.com/pingcap/chaos-mesh/blob/master/doc/dev_hello_world.md) document

PR #1518, opened 2021-02-24
Bugfix: Use `Last` to get current phase at start time of standard cron (#1286)

### LINKED ISSUE #1270: Bug: chaos would directly start when it got applied, even if the cron doesn't want it be so

Yiyiyimu:
cherry-pick #286 to release-1.1

---

fix #1270

Add function `Last` to get the last start time of standard cron, so we could know the current phase that chaos should go into when start.

### Checklist

Tests

- [x] Unit test
- [x] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/api/v1alpha1/common_validation.go
+++ b/api/v1alpha1/common_validation.go
@@ -63,21 +63,10 @@ func ValidateScheduler(schedulerObject InnerSchedulerObject, spec *field.Path) f
 func validateSchedulerParams(duration *time.Duration, durationField *field.Path, spec *SchedulerSpec, schedulerField *field.Path) field.ErrorList {
 	allErrs := field.ErrorList{}
 	if duration != nil && spec != nil {
-
 		cronField := schedulerField.Child("cron")
-		scheduler, err := ParseCron(spec.Cron, cronField)
+		_, err := ParseCron(spec.Cron, cronField)
 		if len(err) != 0 {
 			allErrs = append(allErrs, err...)
-		}
-
-		if scheduler != nil {
-			tmpTime := time.Time{}
-			nextTime := scheduler.Next(tmpTime)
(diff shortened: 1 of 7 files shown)


YangKeao (approved):
LGTM

fewdan (approved):
LGTM

PR #1521, opened 2021-02-24
chore: init workflow api types

g1eny0ung:

This is a draft PR that describes the type definitions of workflow API base on [the workflow design](https://github.com/chaos-mesh/chaos-mesh/blob/master/workflow/docs/design.md).

> This PR may not necessarily be merged.

The whole package starts from `Workflow`. Suppose we take restful API as the standard, this type usually used in the below routes:

- `GET` `/workflows`
- `GET` `/workflows/{id}`
- `PUT` `/workflows/{id}`
- `DELETE` `/workflows/{id}`

For creating a workflow, we should use the `Node` type. Each workflow can be described as a `[]Node`. So we can also define the route as:

- `POST` `/workflows`

If we want to get the detail of a workflow, we should use the `Detail` type. The route definition:

- `GET` `/workflows/{id}/detail`

---

And others that need to note are:

- All `Status` and `State` types maybe duplicate in the code of the workflow engine. These definitions are just placeholders.
- `Topology` represents the topology structure of a workflow. It will be used to display the entire workflow process on the front-end.
- `Node` has specialized fields like `Serial`. This represents this `Node` has some extra definitions.
- The `spec` field under a `Template` is just a YAML definition. So I use `interface{}` to define it.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [x] No code

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/pkg/core/workflow/types.go
+++ b/pkg/core/workflow/types.go
@@ -58,7 +58,7 @@ type Topology struct {
 // TopologyNode defines the basic structure of a node.
 type TopologyNode struct {
 	Type  nodeType  `json:"type"`
-	State nodeState `json:"state"`
+	State nodeState `json:"state,omitempty"`
 }
 
 // Node defines the single step of a workflow.
@@ -151,8 +151,8 @@ const (
 // Detail defines the detail of a workflow.
 type Detail struct {
 	gorm.Model
-	WorkflowID uint       `json:"workflow_id"`
-	Templates  []Template `json:"templates"`
+	WorkflowUID string     `json:"workflow_uid"`
+	Templates   []Template `json:"templates"`
 }
 
 // Template defines a complete structure of a template.
--- a/pkg/core/workflow/types.go
+++ b/pkg/core/workflow/types.go
@@ -1,0 +1,164 @@
+// Copyright 2021 Chaos Mesh Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package workflow includes all type definitaions about Workflow.
+package workflow
+
+import (
+	"github.com/jinzhu/gorm"
+)
+
+// Workflow defines the root structure of a workflow.
+type Workflow struct {
+	gorm.Model
+	UID         string         `gorm:"index:uid" json:"uid"`
+	Name        string         `json:"name"`
+	Entry       string         `json:"entry"` // the entry node name
+	Status      workflowStatus `json:"status"`
+	CurrentNode TopologyNode   `json:"current_node"`
+	Topology    Topology       `json:"topology"`
+}
+
+// workflowState defines the current state of a workflow.
+//
+// Includes: Initializing, Running, Errored, Finished.
+//
+// Const definitions can be found below this type.
+type workflowStatus string
+
+const (
+	// WorkflowInitializing represents a workflow is being initialized.
+	WorkflowInitializing workflowStatus = "Initializing"
+
+	// WorkflowRunning represents that a workflow is running.
+	WorkflowRunning workflowStatus = "Running"
+
+	// WorkflowErrored represents an error in a workflow.
+	WorkflowErrored workflowStatus = "Errored"
+
+	// WorkflowFinished represents that a workflow has ended.
+	WorkflowFinished workflowStatus = "Finished"
+)
+
+// Topology describes the process of a workflow.
+type Topology struct {
+	Nodes []TopologyNode `json:"nodes"`
+}
+
+// TopologyNode defines the basic structure of a node.
+type TopologyNode struct {
+	Type  nodeType  `json:"type"`
+	State nodeState `json:"state"`
+}
+
+// Node defines the single step of a workflow.
+type Node struct {
+	TopologyNode
+	Serial   NodeSerial   `json:"serial,omitempty"`
+	Parallel NodeParallel `json:"parallel,omitempty"`
+	Template Template     `json:"template"`
+}
+
+// NodeSerial defines SerialNode's specific fields.
+type NodeSerial struct {
+	Tasks    []Node `json:"tasks"`
+	Circular bool   `json:"circular"`
+}
+
+// NodeParallel defines ParallelNode's specific fields.
+type NodeParallel struct {
+	Tasks []Node `json:"tasks"`
+}
+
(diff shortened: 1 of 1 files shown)


pkg/core/workflow/types.go
```diff
@@ -0,0 +1,164 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package workflow includes all type definitaions about Workflow.
+package workflow
+
+import (
+	"github.com/jinzhu/gorm"
```
STRRL, line 18: I think we do not need `gorm` because these entities are not stored at DB, but stored at kubernetes API.

pkg/core/workflow/types.go
```diff
@@ -0,0 +1,164 @@
...
+)
+
+// Workflow defines the root structure of a workflow.
+type Workflow struct {
+	gorm.Model
+	UID         string         `gorm:"index:uid" json:"uid"`
+	Name        string         `json:"name"`
```
STRRL, line 25: We might not use `namespace + name` instead of UID as a unique identity, just mentioning, it's not so important. 😁

pkg/core/workflow/types.go
```diff
@@ -0,0 +1,164 @@
...
+	TopologyNode
+	Serial   NodeSerial   `json:"serial,omitempty"`
+	Parallel NodeParallel `json:"parallel,omitempty"`
+	Template Template     `json:"template"`
+}
+
+// NodeSerial defines SerialNode's specific fields.
+type NodeSerial struct {
+	Tasks    []Node `json:"tasks"`
+	Circular bool   `json:"circular"`
```
STRRL, line 75: What does the `Circular` field mean?

 Is this field affect the drawing of workflow?

pkg/core/workflow/types.go
```diff
@@ -0,0 +1,164 @@
...
+type Node struct {
+	TopologyNode
+	Serial   NodeSerial   `json:"serial,omitempty"`
+	Parallel NodeParallel `json:"parallel,omitempty"`
+	Template Template     `json:"template"`
+}
+
+// NodeSerial defines SerialNode's specific fields.
+type NodeSerial struct {
+	Tasks    []Node `json:"tasks"`
```
STRRL, line 74: I prefer `Tasks []string`.

Because we already put all the Node in the field `Workflow.Topology` as an array. If we using a struct here, there might a little redundancy in our JSON.

STRRL (commented):
Basically LGTM! 😁

We might change some names of fields for keeping consistent with the backend later. But before that, I still have some questions about definations.

PTAL @g1eny0ung

pkg/core/workflow/types.go
g1eny0ung: Got it!

pkg/core/workflow/types.go
g1eny0ung: Know about it. This conversation can be saved here to remind others who view this PR in the future.

g1eny0ung: Know about it 👌🏽.

pkg/core/workflow/types.go
g1eny0ung: > Is this field affect the drawing of workflow?

Not affect this.

I remembered that we are talking about the serial node that can be re-executed when it's ended. So I add `Circular ` to this struct.

But I now realize this kind of loop execution can be controlled by the task node. I don't know if I understand it right. If it is, this field is no needed.

pkg/core/workflow/types.go
g1eny0ung: The `Workflow.Topology` is just a skeleton of a Node. It only includes `Type` and `State` (can be `nil` when creating a workflow). I think I forgot to define an API about getting the complete nodes of a workflow:

- `GET` `/workflows/{id}/nodes`

If we defined it as `[]string`, we can only get the `task` detail from the template (by its name string). According to my current understanding, a `Node` is an instance of a `Template`. One design for runtime, the other is a static template.

pkg/core/workflow/types.go
STRRL: OK. 🙈 

As you say, it's a little complex to determine that field is true or false.

I think it's safe to delete it.

g1eny0ung: 👌🏽

pkg/core/workflow/types.go
STRRL: `Node` also has a unique name. There might some `Node`s that refs to one same `Template`, we use node name to distinguish.

Does the sorting of items in `Workflow.Topology` is required? If YES, what's the expected order?

g1eny0ung: > There might some Nodes that refs to one same Template, we use node name to distinguish.

🙋🏽‍♂️ I think I understand. I will change `[]Node` to `[]string` later.

> Does the sorting of items in Workflow.Topology is required? If YES, what's the expected order?

Yes. Should be sorted from entry to the end.

g1eny0ung: Done.

STRRL: > Yes. Should be sorted from entry to the end.

But the topology is a tree, we could not describe it only by this array. 🤔

STRRL: > 🙋🏽‍♂️ I think I understand. I will change []Node to []string later.

I think we should a new field `name` in `TopologyNode`. After that, both `[]string` and `[]Node` are OK to me.

g1eny0ung: > I think we should a new field name in TopologyNode.

Already added in 7e1f08a.

pkg/core/workflow/types.go
```diff
@@ -0,0 +1,153 @@
...
+package workflow
+
+// Workflow defines the root structure of a workflow.
+type Workflow struct {
+	UID         string         `json:"uid"`
+	Name        string         `json:"name"`
+	Entry       string         `json:"entry"` // the entry node name
+	Status      workflowStatus `json:"status"`
+	Topology    Topology       `json:"topology"`
+	CurrentNode Node           `json:"current_node"`
```
STRRL, line 24: Due to the `parallel node`, this field should be an array.

g1eny0ung: Done.~

pkg/core/workflow/types.go
g1eny0ung: After discussing with @STRRL. This conversation's content is outdated.

We removed the `TopologyNode` and put its fields into `Node`.

Please refer to the latest code. I will mark this conversation as resolved.

g1eny0ung:
**Some updates:**

- `CurrentNodes` (Previous `CurrentNode`) is an array because of a parallel node execs multiple nodes at the same time.
- The `Template` name in the `Node` define as `string`. It's needed to get a complete template with an API.

g1eny0ung:
Close due to #1527.

PR #1529, opened 2021-03-02
Adding parameterized timeoutSeconds for webhooks

anuragpaliwal80:
Allowing setting `timeoutSeconds` for webhooks. Right now we are using the default timeout which is quite high.

Default timeout for a webhook call is 30 seconds for webhooks created using admissionregistration.k8s.io/v1beta1. It is recommended that admission webhooks should evaluate as quickly as possible (typically in milliseconds), since they add to API request latency. It is encouraged to use a small timeout for webhooks.
If the webhook call times out, the request is handled according to the webhook's failure policy.

Refer: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts for more details.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -14,6 +14,7 @@ webhooks:
       path: /inject-v1-pod
   failurePolicy: Fail
   name: vpod.kb.io
+  timeoutSeconds: 5
   rules:
   - apiGroups:
     - ""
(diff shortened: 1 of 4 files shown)


YangKeao:
Please run `./hack/update_install_script.sh` to update the `install.sh` automatically. It seems that there are a little difference between the automatically generated `install.sh` and the one in this PR :smile_cat:

YangKeao (approved):
LGTM

fewdan (approved):
LGTM

PR #1549, opened 2021-03-10
Fix meaningless retry operations caused by the updates of status field 

cwen0:

cherry-pick #1533 

Use `WithEventFilter` to filter update events and ignore updates to CRD object status.

To compare the spec and meta fields of the CRD object, we add a `GetSpecAndMetaString` function for each chaos CRD. If we update the CRD version to v1, we can delete `GetSpecAndMetaString` function, because we can use `unstructured.Unstructured` struct to parse CRD object directly.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```

### DIFF
--- a/api/v1alpha1/common_types.go
+++ b/api/v1alpha1/common_types.go
@@ -228,6 +228,7 @@ type InnerObject interface {
 	IsDeleted() bool
 	IsPaused() bool
 	GetChaos() *ChaosInstance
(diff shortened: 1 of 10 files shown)


AsterNighT (approved):
LGTM

PR #1551, opened 2021-03-10
workflow basic CRD and reconcilers

STRRL:

That looks like a huge PR, but most of the codes still generated. :)

> Almost 12422 lines of codes are generated yaml and zz_generated codes.

Implement Chaos Mesh Workflow by kubernetes operator. 

These codes are not related to the design and codes in the previous PR. I am trying to implement the workflow with kubernetes pattern, I think I have found the way.

There are not annoying whatever-driven state-machine anymore, only desire and current state. Instead of state machines, I use various reconcilers to describe the behavior of workflow, for example:
- serial node reconciler: create new nodes one-by-one
- chaos node reconciler: create and maintenance Chaos CR 
- deadline reconciler: update the state represented by `deadline` and `duration`
- ...

and there are some conditions(could generate by observation) for scheduling and chaos injection:

- Accomplished: means node finished actively
- DeadlineExceed: means node finished with deadline exceed
- ChaosInjected: means Chaos CR is created and synced.
- ...

Here is an example at `/examples/workflow/workflow.yaml`.

These features are now supported:
- suspend node
- serial node
- networkchaos
- podchaos

(WIP)

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


STRRL:
I think this PR is ready to review, PTAL @chaos-mesh/committers

api/v1alpha1/workflow_types.go
```diff
@@ -0,0 +1,102 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:shortName=wf
+type Workflow struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec defines the behavior of a pod chaos experiment
```
AsterNighT, line 26: comments need fixing

api/v1alpha1/workflow_types.go
```diff
@@ -0,0 +1,102 @@
...
+	Spec WorkflowSpec `json:"spec"`
+
+	// +optional
+	// Most recently observed status of the chaos experiment about pods
```
AsterNighT, line 30: ditto

api/v1alpha1/workflow_types.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+type WorkflowStatus struct {
+	// +optional
+	EntryNode *string `json:"entry_node,omitempty"`
+}
+
+type TemplateType string
+
+const (
```
AsterNighT, line 68: Should this be generated rather than hand-written later?

api/v1alpha1/workflownode_types.go
```diff
@@ -0,0 +1,108 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:shortName=wfn
+type WorkflowNode struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	// Spec defines the behavior of a pod chaos experiment
```
AsterNighT, line 27: comments

api/v1alpha1/workflownode_types.go
```diff
@@ -0,0 +1,108 @@
...
+	Spec WorkflowNodeSpec `json:"spec"`
+
+	// +optional
+	// Most recently observed status of the chaos experiment about pods
```
AsterNighT, line 31: ditto

api/v1alpha1/workflownode_types.go
```diff
@@ -0,0 +1,108 @@
...
+	Status WorkflowNodeStatus `json:"status"`
+}
+
+type WorkflowNodeSpec struct {
+	TemplateName string       `json:"template_name"`
+	WorkflowName string       `json:"workflow_name"`
```
AsterNighT, line 37: What does `WorkflowName` mean here? Is it the workflow the node belongs to?

api/v1alpha1/workflownode_types.go
```diff
@@ -0,0 +1,108 @@
...
+	// +optional
+	ChaosResource *corev1.TypedLocalObjectReference `json:"chaos_resource,omitempty"`
+
+	// ActiveChildren means the created children node
+	// +optional
+	ActiveChildren []corev1.LocalObjectReference `json:"active_children,omitempty"`
+
+	// Children is necessary for representing the order when replicated child template references by parent template.
+	// +optional
+	FinishedChildren []corev1.LocalObjectReference `json:"finished_children,omitempty"`
```
AsterNighT, line 64: Could you explain the function of `FinishedChildren` a bit more in detail pls? I don't quite get it.

pkg/workflow/controllers/accomplish_watcher.go
```diff
@@ -0,0 +1,147 @@
...
+		}, &parentNode)
+
+		if err != nil {
+			return client.IgnoreNotFound(err)
+		}
+
+		// filter out accomplished node
+		var newActiveChildren []corev1.LocalObjectReference
+
+		for _, item := range parentNode.Status.ActiveChildren {
```
AsterNighT, line 112: uhh this is not too good. What about an abstract `filter`

pkg/workflow/controllers/chaos_node_reconciler.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (it *ChaosNodeReconciler) injectChaos(ctx context.Context, node v1alpha1.WorkflowNode) error {
+	var chaosObject runtime.Object
+	var meta metav1.Object
+
+	if node.Spec.Type == v1alpha1.TypeNetworkChaos {
```
AsterNighT, line 71: rly? one if per chaos type?

pkg/workflow/controllers/chaos_node_reconciler.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (it *ChaosNodeReconciler) recoverChaos(ctx context.Context, node v1alpha1.WorkflowNode) error {
+	if node.Status.ChaosResource == nil {
+		return nil
+	}
+
+	var chaosObject runtime.Object
+
+	var err error
+	if node.Spec.Type == v1alpha1.TypeNetworkChaos {
```
AsterNighT, line 153: Seems there're tons of code to be generated. But I fear that code generating would not be the best practice to do such things.

pkg/workflow/controllers/workflow_entry_reconciler.go
```diff
@@ -0,0 +1,94 @@
...
+		entryNode := nodes[0]
+		err = it.kubeClient.Create(ctx, entryNode)
+		if err != nil {
+			it.logger.V(1).Info("failed to create workflow nodes")
+			return reconcile.Result{}, err
+		}
+
+		it.eventRecorder.Event(&workflow, corev1.EventTypeNormal, v1alpha1.EntryCreated, "Entry node created")
+
+		updateError := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
AsterNighT, line 71: There're lots of `RetryOnConflict` in the code, some of them seems unnecessary to me ( like this one ). Are we supporting multiple controller-managers?

pkg/workflow/controllers/utils.go
```diff
@@ -0,0 +1,57 @@
...
+func SetCondition(status *v1alpha1.WorkflowNodeStatus, condition v1alpha1.WorkflowNodeCondition) {
+	currentCond := GetCondition(*status, condition.Type)
+	if currentCond != nil && currentCond.Status == condition.Status && currentCond.Reason == condition.Reason {
+		return
+	}
+	newConditions := filterOutCondition(status.Conditions, condition.Type)
+	status.Conditions = append(newConditions, condition)
+}
+
+func GetCondition(status v1alpha1.WorkflowNodeStatus, conditionType v1alpha1.WorkflowNodeConditionType) *v1alpha1.WorkflowNodeCondition {
```
AsterNighT, line 31: Could really use some abstract `filter`, `find` sort of things.

AsterNighT: Apply to the rest of this file.

AsterNighT (changes requested):

api/v1alpha1/workflownode_types.go
STRRL: > Is it the workflow the node belongs to?
Yes.

api/v1alpha1/workflownode_types.go
STRRL: This field refs to the children nodes which accomplished. For example, One serial node called A has three children nodes: B, C, D; when B and C finished, D is running, the status of A just like:

```jsonc
{
"expected_children": 3,
"active_children": ["D"],
"finished_children": ["B", "C"],
}
```

pkg/workflow/controllers/accomplish_watcher.go
STRRL: Ok, I will extract these lines of codes into a new method, like:
```go
var newActiveChildren []corev1.LocalObjectReference

parentNode.Status.ActiveChildren = filterChildrens(parentNode.Status.ActiveChildren, func(item corev1.LocalObjectReference) bool {
return item.Name != childNode.Name
})
```

pkg/workflow/controllers/chaos_node_reconciler.go
STRRL: > Seems there're tons of code to be generated.

Yeah, I suppose that we will use code generation in the future. I will finish that in the next PRs.

>But I fear that code generating would not be the best practice to do such things.

Do you have any suggestions?

pkg/workflow/controllers/accomplish_watcher.go
AsterNighT: I thought that there will be something like a haskell `filter` in go but to my surprise there's no such thing. Then it comes to me that there's no generic type in go 🤦 . In fact what I prefer is a
```go
func filter<T>(arr: []T, f: func(T) bool) []T
```
Since there's no such <T> thing, there may have to be some type assert with struct{}.

pkg/workflow/controllers/workflow_entry_reconciler.go
STRRL: > Are we supporting multiple controller-managers?

Different reconciles are working concurrently within one controller-manager process, and users could also modify the objects. And I think we should respect the optimistic lock of kubernetes API, and resolve it explicitly.

And one more thing, multiple controller-managers will election one leader, and only the leader will execute reconcilers.

pkg/workflow/controllers/chaos_node_reconciler.go
AsterNighT: I'm not sure if we can actually do this in go, so  it's just a thought.
Is there any chance we can do something like
```go
type = some_match_function(node.Spec.Type); // yields some reflect.Type
chaos = reflect.New(type); // and fill in the specs
```
there will still be some code generation to do in `some_match_function` but that will be less duplication.

pkg/workflow/controllers/accomplish_watcher.go
STRRL: I do not think we will use `struct{}` or `interface{}` if you mean.

> Actually, these codes are copied from kubernetes replicaset controller utils, it also struggling with no generic type.

I think It's ok for now.🙈

pkg/workflow/controllers/chaos_node_reconciler.go
STRRL: Yeah, we could do that, but it still needs code generation in `some_match_function()`. I will compare these different codes and choose the sipmler one.

api/v1alpha1/workflownode_types.go
```diff
@@ -0,0 +1,108 @@
...
+	// +optional
+	// +patchMergeKey=type
+	// +patchStrategy=merge
+	Conditions []WorkflowNodeCondition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+}
+
+type WorkflowNodeConditionType string
+
+const (
+	CondiitonAccomplished   WorkflowNodeConditionType = "Accomplished"
```
WangXiangUSTC, line 76:
```suggestion
	ConditionAccomplished   WorkflowNodeConditionType = "Accomplished"
```

pkg/workflow/controllers/chaos_node_reconciler.go
AsterNighT: In fact I was wondering if there should be one and only one way to apply transition between a `string` and some corresponding `reflect.Type`. In fact there're already different cases in the code now. If I haven't got it wrong, one is here 
https://github.com/chaos-mesh/chaos-mesh/blob/c9d9f0706ae1c1c1ee0cc06d3df6b5370ff5505b/pkg/apiserver/experiment/experiment.go#L147-L155
and 
https://github.com/chaos-mesh/chaos-mesh/blob/c9d9f0706ae1c1c1ee0cc06d3df6b5370ff5505b/pkg/apiserver/experiment/experiment.go#L812-L827
also
https://github.com/chaos-mesh/chaos-mesh/blob/c9d9f0706ae1c1c1ee0cc06d3df6b5370ff5505b/cmd/chaos-builder/validator.go#L21-L32

api/v1alpha1/workflow_types.go
STRRL: fixed in ade7869

api/v1alpha1/workflow_types.go
STRRL: fixed in ade7869

api/v1alpha1/workflownode_types.go
STRRL: fixed in ade7869

api/v1alpha1/workflownode_types.go
STRRL: fixed in ade7869

api/v1alpha1/workflownode_types.go
STRRL: fixed in 09cde3d

pkg/workflow/controllers/chaos_node_reconciler.go
STRRL: Yeah, I think we could reuse the `Kind<type>Chaos` and `chaosKindMap`. 

But I think I will write another (duplicated) code generation for the type transition things, and make refactor to "one and only one way for `string`/`reflect.Type` transition" after that.

STRRL: FYI @WangXiangUSTC

api/v1alpha1/workflow_types.go
STRRL: Yes, it should be generated. I will finish the code generation in the next PRs.

pkg/workflow/controllers/chaos_node_reconciler.go
STRRL: There're only several limited chaos types for the demo. 😰

It should be one method like `fillChaosSpec()` in the future.

pkg/workflow/controllers/new_node.go
```diff
@@ -0,0 +1,116 @@
...
+			now := metav1.NewTime(time.Now())
+			var deadline *metav1.Time = nil
+
+			if template.Duration != nil {
+				duration, err := time.ParseDuration(*template.Duration)
+				if err != nil {
+					// TODO: logger
+					return nil, err
+				}
+				result := metav1.NewTime(now.DeepCopy().Add(duration))
```
WangXiangUSTC, line 55: how about renaming the `result`, it is duplicate with L42

pkg/workflow/controllers/serial_node_reconciler.go
```diff
@@ -0,0 +1,144 @@
...
+		for _, item := range childrenNodes {
+			childrenNames = append(childrenNames, item.Name)
+		}
+		it.logger.Info("serial node's child created", "key", request.NamespacedName, "children", childrenNames)
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func NewSerialNodeReconciler(kubeClient client.Client, eventRecorder record.EventRecorder, logger logr.Logger) *SerialNodeReconciler {
```
WangXiangUSTC, line 142: how about moving this function before `Reconcile` function, just like other file

pkg/workflow/controllers/workflow_entry_reconciler.go
```diff
@@ -0,0 +1,94 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/chaos-mesh/chaos-mesh/api/v1alpha1"
+)
+
+// WorkflowEntryReconciler watches on Workflow, creates ew Entry Node for created Workflow.
```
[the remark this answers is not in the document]
STRRL: I want to say, `new`, that's a typo. 🙈

pkg/workflow/controllers/chaos_node_reconciler.go
```diff
@@ -0,0 +1,205 @@
...
+
+	if availableChaos(string(node.Spec.Type)) {
+		err = it.injectChaos(ctx, node)
+		return reconcile.Result{}, err
+	}
+
+	return reconcile.Result{}, nil
+}
+
+func (it *ChaosNodeReconciler) injectChaos(ctx context.Context, node v1alpha1.WorkflowNode) error {
```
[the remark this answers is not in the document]
STRRL: It has no other meaning, like `self`. Golang does not restrict the name of the method receiver, I like using `it`, just like a kotlin closure. 🙈

pkg/workflow/controllers/accomplish_watcher.go
```diff
@@ -0,0 +1,147 @@
...
+			return reconcile.Result{}, nil
+		}
+
+		// unexpected situation
+		if len(owners) > 1 {
+			it.logger.V(1).Info("node has more than one owner, it will not take any operates", "node", request.NamespacedName, "owners", owners)
+			return reconcile.Result{}, nil
+		}
+
+		if len(owners) == 1 {
```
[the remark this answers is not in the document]
STRRL: I am afraid we can't. The `owners` could be modified by users or other components(for no reason), and we could not suppose that it has only one element. We should resolve that situation and give some info or logs about it.

pkg/workflow/controllers/serial_node_reconciler.go
```diff
@@ -0,0 +1,144 @@
...
+		updateError := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			node := v1alpha1.WorkflowNode{}
+			err := it.kubeClient.Get(ctx, request.NamespacedName, &node)
+			if err != nil {
+				return err
+			}
+			for _, item := range childrenNodes {
+				node.Status.ActiveChildren = append(node.Status.ActiveChildren, corev1.LocalObjectReference{Name: item.Name})
+			}
+			return it.kubeClient.Update(ctx, &node)
```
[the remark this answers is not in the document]
STRRL: Yes, it might make a mess with the high-presssure situation. 

The reason is now I do not make "create node" and "update state" as an atomic operation. I think I could fix that by "labeling when creating" and use "list with label selector" to resolve that.

I think I should resolve it in another PR.🤔 Currently, it could work for most situations.

api/v1alpha1/workflownode_types.go
```diff
@@ -0,0 +1,108 @@
...
+	Tasks []string `json:"tasks,omitempty"`
+	// +optional
+	*EmbedChaos `json:",inline,omitempty"`
+}
+
+type WorkflowNodeStatus struct {
+
+	// ExpectedChildren means the expected children to execute
+	// +optional
+	ExpectedChildren *int `json:"expected_children,omitempty"`
```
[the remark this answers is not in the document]
STRRL: fixed in 187a713

pkg/workflow/controllers/chaos_node_reconciler.go
```diff
@@ -0,0 +1,205 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: fmt.Sprintf("%s-", node.Name),
+				Namespace:    node.Namespace,
+			},
+			Spec: *node.Spec.PodChaos,
+		}
+		meta = podChaos.GetObjectMeta()
+		chaosObject = &podChaos
+	} else {
+		it.logger.Info("unsupported chaos nodes", "key", fmt.Sprintf("%s/%s", node.Namespace, node.Name), "type", node.Spec.Type)
```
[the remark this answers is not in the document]
STRRL: fixed in 187a713

pkg/workflow/controllers/new_node.go
```diff
@@ -0,0 +1,116 @@
...
+				renderedNode.Labels[LabelControlledBy] = workflow.Name
+			}
+
+			renderedNode.Finalizers = append(renderedNode.Finalizers, metav1.FinalizerDeleteDependents)
+
+			result = append(result, &renderedNode)
+			continue
+		}
+		return nil, errors.New(
+			fmt.Sprintf("workflow %s do not contains template claled %s",
```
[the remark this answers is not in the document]
STRRL: fixed in 187a713

pkg/workflow/controllers/workflow_entry_reconciler.go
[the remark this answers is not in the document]
STRRL: fixed in 187a713

pkg/workflow/controllers/new_node.go
STRRL: fixed in 187a713

pkg/workflow/controllers/serial_node_reconciler.go
STRRL: fixed in 8d3d31f

STRRL:
Hi @WangXiangUSTC, comments addressed, PTAL!

pkg/workflow/controllers/accomplish_watcher.go
```diff
@@ -0,0 +1,147 @@
...
+				if err != nil {
+					return reconcile.Result{}, err
+				}
+				if parentNode.Spec.Type == v1alpha1.TypeSerial {
+					err = it.updateParentSerialNode(ctx, node, parentNode)
+					return reconcile.Result{}, err
+				}
+				it.logger.Info("unsupported owner node type", "node type", parentNode.Spec.Type)
+			} else if owner.Kind == GetKindOf(&v1alpha1.Workflow{}) {
+				it.logger.Info("unfinished update for workflow", "kind", owner.Kind)
```
WangXiangUSTC, line 85: why it is unfinished?

pkg/workflow/controllers/chaos_node_reconciler.go
```diff
@@ -0,0 +1,206 @@
...
+		err := it.kubeClient.Get(ctx, types.NamespacedName{
+			Namespace: node.Namespace,
+			Name:      node.Name,
+		}, &node)
+		if err != nil {
+			return client.IgnoreNotFound(err)
+		}
+		node.Status.ChaosResource = &chaosRef
+
+		// TODO: this condition should be set by observation
```
WangXiangUSTC, line 133: what is the `observation` means?

pkg/workflow/controllers/accomplish_watcher.go
STRRL: It's still working in progress, the status of Workflow is still unfinished. 🙈 

I think I might add a comment with TODO.

pkg/workflow/controllers/chaos_node_reconciler.go
STRRL: It's related to the above comments about labeling resources when creating.

Currently, we set that `ConditionChaosInjected` to `true`, because we have created a Chaos Object. It will cause the similar issue with https://github.com/chaos-mesh/chaos-mesh/pull/1551#discussion_r597441206. 

The best practice is pretty same with the above comment: labeling when creating resource, and if we could list out some resource with that label, change that condition to true.

pkg/workflow/controllers/chaos_node_reconciler.go
```diff
@@ -0,0 +1,205 @@
...
+			types.NamespacedName{Namespace: node.Namespace, Name: node.Status.ChaosResource.Name},
+			&target)
+		chaosObject = &target
+	} else {
+		it.logger.Info("unsupported chaos nodes", "key", fmt.Sprintf("%s/%s", node.Namespace, node.Name), "type", node.Spec.Type)
+		return nil
+	}
+
+	if apierrors.IsNotFound(err) {
+		// TODO: debug log
```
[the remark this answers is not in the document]
STRRL: fixed in 3eb11e6

pkg/workflow/controllers/accomplish_watcher.go
STRRL: fixed in 3eb11e6

pkg/workflow/controllers/chaos_node_reconciler.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	err = it.kubeClient.Delete(ctx, chaosObject)
+
+	if client.IgnoreNotFound(err) != nil {
+		return err
+	}
+
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		node := v1alpha1.WorkflowNode{}
```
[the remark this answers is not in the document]
STRRL: I think we do not need that. Whatever the type is, we will just clear the `node.Status.ChaosResource` to `nil`.

pkg/workflow/controllers/accomplish_watcher.go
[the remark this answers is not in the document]
STRRL: Ohh I watched on the incorrect line. 

Fixed in a430a5a

WangXiangUSTC (approved):
LGTM

AsterNighT (dismissed):
LGTM

STRRL:
waiting for #1588

STRRL:
I just update the code that fixes the injection logic. 😰

PTAL again. @WangXiangUSTC  @AsterNighT .

pkg/workflow/controllers/chaos_node_reconciler.go
```diff
@@ -120,10 +122,12 @@ func (it *ChaosNodeReconciler) injectChaos(ctx context.Context, node v1alpha1.Wo
 	}
 
 	updateError := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		namespace := node.Namespace
+		name := node.Name
 		node := v1alpha1.WorkflowNode{}
```
WangXiangUSTC, line 127: how about renaming the `node`?  It's easy to confuse it with the `node` out of this function 😢

STRRL: sure!

STRRL: fixed in b80e9d2

WangXiangUSTC (approved):
LGTM

pkg/workflow/controllers/chaos_node_reconciler.go
YangKeao: In fact, I always prefer code generation than `reflect` (if code generation is possible). But code generator is really hard to write. I wonder whether there is something like [`quote!`](https://github.com/dtolnay/quote) in the go world? (I think go ecosystem depends heavily on the code generation)

YangKeao (approved):
LGTM

STRRL:
Hi @YangKeao @WangXiangUSTC, as https://github.com/chaos-mesh/chaos-mesh/pull/1581 merged, the definition of the workflow needs updating, PTAL.

WangXiangUSTC (approved):
LGTM

YangKeao (approved):
LGTM

PR #1578, opened 2021-03-18
Unified selector and new controller design

YangKeao:
Please notice: this PR doesn't merge into `master`, but a `nirvana` branch to land the https://github.com/chaos-mesh/chaos-mesh/discussions/1520 in 2.0 .

This is the first step of https://github.com/chaos-mesh/chaos-mesh/discussions/1520 . And the corresponding RFC is at https://github.com/chaos-mesh/rfcs/pull/15 

You can feel the fascination of current implementation in the implementation of [podkill](https://github.com/chaos-mesh/chaos-mesh/blob/a55dc76a64d6a46817af45223d24883e99939fcb/controllers/chaosimpl/podchaos/podkill/impl.go). It doesn't need to select and the logic is really simple now.

### TODO

- [ ] Fix dashboard event (as the `StartTime` and `EndTime` have been removed, the events in `chaos-dashboard` should be modified)
- [ ] Migrate from `containerName` to `containerNames`, especially for `chaos-dashboard`
- [ ] Send Kubernetes events to store the error message
- [ ] Use `uber-go/fx` to do dependency injection
- [ ] Add `GCPChaos`, `HTTPChaos`, `KernelChaos`, `JVMChaos`.
- [ ] `Condition` based status display.


YangKeao:
This PR would merge to `nirvana` branch first, so that other developers could develop new features on this patch.

PR #1583, opened 2021-03-22
chaos-dashboard web api for workflow

STRRL:

Implementation of chaos-dashboard web API, should not be merged before #1551.

Implementation of chaos-dashboard web API, build a repo with kube client.
And it updates the API about the parameter about list, get, and the entity of Workflow and WorkflowDetail

### Checklist

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


STRRL:
PTAL 
/cc @g1eny0ung

pkg/core/workflow.go
```diff
@@ -97,49 +103,12 @@ const (
 	TaskNode NodeType = "TaskNode"
 )
 
-// NodeState represents a node in different stage.
```
AsterNighT, line 100: Umm why are these statuses deleted?

AsterNighT (approved):

pkg/apiserver/workflow/workflow.go
```diff
@@ -29,46 +32,76 @@ func Register(r *gin.RouterGroup, s *Service) {
 	endpoint := r.Group("/workflows")
 	endpoint.GET("", s.listWorkflows)
 	endpoint.POST("/new", s.createWorkflow)
-	endpoint.GET("/detail/:uid", s.getWorkflowDetail)
-	endpoint.DELETE("/:uid", s.deleteWorkflow)
-	endpoint.PUT("/:uid", s.updateWorkflow)
+	endpoint.GET("/detail/:namespace/:name", s.getWorkflowDetail)
```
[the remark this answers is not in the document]
g1eny0ung: More simpler path, `GET /:namespace/:name`?

pkg/core/workflow.go
```diff
@@ -154,3 +123,162 @@ type Template struct {
...
+		Name:      name,
+	}, &kubeWorkflow)
+	if err != nil {
+		return err
+	}
+	return it.kubeclient.Delete(ctx, &kubeWorkflow)
+}
+
+// func MutateWithKubeClient could spawn a new repo with the new kubeclient with another auth session.
+func (it *KubeWorkflowRepository) MutateWithKubeClient(anotherKubeclient client.Client) *KubeWorkflowRepository {
```
g1eny0ung, line 203: It seems that this method is no longer used.

pkg/apiserver/server.go
```diff
@@ -37,6 +38,7 @@ var (
 		fx.Provide(
 			newEngine,
 			newAPIRouter,
+			core.NewKubeWorkflowRepository,
```
g1eny0ung, line 41: We haven't provided the `client.Client` instance in `fx.provide`. I think this registration will fail.

g1eny0ung: BTW, all APIs use `github.com/chaos-mesh/chaos-mesh/pkg/clientpool` as the kube client. Should workflow API also use it?

pkg/core/workflow.go
STRRL: We have use Conditions to describe that in kubernetes API, I think we have no necessary to translate it to these NodeState(s).

pkg/apiserver/server.go
STRRL: yeah, we should use the `clientpool`, I will fix it ASAP.😅

STRRL:
fixed @g1eny0ung , PTAL

/lgtm cancel

pkg/apiserver/server.go
g1eny0ung: Seems we don't need to provide this currently.

pkg/core/common.go
```diff
@@ -0,0 +1,33 @@
...
+
+// KubeObjectYAMLDescription defines the YAML structure of an object stored in kubernetes API.
+type KubeObjectYAMLDescription struct {
+	APIVersion string                 `json:"apiVersion"`
+	Kind       string                 `json:"kind"`
+	Metadata   KubeObjectYAMLMetadata `json:"metadata"`
+	Spec       interface{}            `json:"spec"`
+}
+
+// KubeObjectYAMLMetadata defines the metadata of YAMLDescription.
```
g1eny0ung, line 27: `YAMLDescription` => `KubeObjectYAMLDescription`

STRRL: Fixed in f9459d6, PTAL

g1eny0ung (approved):
LGTM

PR #1584, opened 2021-03-22
fix-typo (aws-finalizer)

fewdan:

fix-typo


WangXiangUSTC:
What will happen if we don't fix it？Or it is only a typo?

fewdan:
> What will happen if we don't fix it？Or it is only a typo?

@WangXiangUSTC It is just a typo, and there will be no consequences if it is not modified. Modify it to be consistent with the case of other places.

WangXiangUSTC (approved):
LGTM

cwen0 (approved):
LGTM

PR #1599, opened 2021-03-26
Support chaosctl tls connection

### LINKED ISSUE #1544: Add TLS config for `chaosctl`
## Bug Report

After adding `tls` configuration for the connection between `chaos-controller-manager` and `chaos-daemon`, `chaosctl` cannot work anymore. Please add some configurations (through environment variable or through parameters) to `chaosctl` to specify the cert to connect to the `chaos-daemon`.

### LINKED ISSUE #1618: Chaosctl fail to create grpc connection
## Bug Report

```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.4", GitCommit:"e87da0bd6e03ec3fea7933c4b5263d151aafd07c", GitTreeState:"clean", BuildDate:"2021-02-18T16:12:00Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}

Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.6", GitCommit:"7015f71e75f670eb9e7ebd4b5749639d42e20079", GitTreeState:"clean", BuildDate:"2019-12-04T07:33:18Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```
Release 1.2, built with go 1.16
**What did you do?**
run `chaosctl debug networkchaos xxx`

Good things.
**What did you see instead?**
`failed get pid from pod default/web-show-6d9d9c6579-89gj9: rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: x509: certificate relies on legacy Common Name field, use SANs or temporarily enable Common Name matching with GODEBUG=x509ignoreCN=0"`

Chaosctl should surpress all errors previously. I'm appalled and sorry for my carelessness.

It seems that go dropped legacy Common Name field support in 1.15 (https://github.com/golang/go/issues/39568#issuecomment-671424481). I see we're using go 1.14.4 but this would cause trouble upon upgrading.

AsterNighT:
Fix #1544 
Fix #1618 

Chaosctl now fetch tls configs from k8s secrets, and use it in grpc connection to chaos daemon
Try to fix the issue that cn verification is not support by go115 or higher. 
I don't know a lot about certs. Help is really appreciated on this.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

1. Start a `networkchaos`
2. run `chaosctl debug networkchaos <chaos-name>`
3. it works well

And try #1620.

```release-note
NONE
```


fewdan (approved):
LGTM

pkg/chaosctl/common/common.go
```diff
@@ -408,17 +408,46 @@ Please check network policy / firewall, or see FAQ on website`, daemon.Name, dae
...
+	var cfg rawTLSConfig
+	restconfig, err := config.GetConfig()
+	if err != nil {
+		return nil, err
+	}
+	kubeClient, err := kubernetes.NewForConfig(restconfig)
+	if err != nil {
+		return nil, err
+	}
+	secret, err := kubeClient.CoreV1().Secrets("chaos-testing").Get("chaos-mesh-daemon-client-certs", metav1.GetOptions{})
```
Yiyiyimu, line 439: It seems this place hardcoded the namespace 😕 Also maybe we could avoid magic words

Yiyiyimu (commented):
LGTM and that's so cool to achieve it this simple!

pkg/chaosctl/common/common.go
AsterNighT: Absolutely! I was wondering if there's a way to automatically get chaosmesh namespace rather than setting it to a fixed value. Have got any ideas on how to do that?

Yiyiyimu: Maybe you could try to get the pod `chaos-daemon`, then to get its namespace. A reference could be: https://github.com/chaos-mesh/chaos-mesh/blob/master/pkg/chaosctl/cmd/logs.go#L90. Hope that be of help 😀

AsterNighT: https://github.com/chaos-mesh/chaos-mesh/blob/6ee0cb9f67c671eefb9110a99e021a260e2f5ac7/pkg/chaosctl/cmd/logs.go#L99

If I haven't got that wrong, it seems we still need to specify ns. I can see that code generation will definitely be a viable method but I'm afraid that'll be an overkill.

Yiyiyimu: I'm not so sure about this, but it seems normally when `clusterscoped` is set to `true`, arguments about namespace would be ignored. Although it seems only STRRL is using `chaosctl logs`, since he hasn't changed this part, maybe we could get chaos components as we need this way.
https://github.com/chaos-mesh/chaos-mesh/blob/f8c3fccd8e7317667b813973bda27c4748a0caa7/pkg/config/controller.go#L61-L63

Besides, I found this parameter (but I haven't tested it yet 😿 ) which seems is what we need.
https://github.com/chaos-mesh/chaos-mesh/blob/f8c3fccd8e7317667b813973bda27c4748a0caa7/pkg/config/controller.go#L72-L73

STRRL (commented):
Please appending comments for `NewChaosDaemonClientLocally`, or change the name.

Rest LGTM

AsterNighT:
> * Please appending comments for `NewChaosDaemonClientLocally`, or change the name.
> * It seems we could not load cert from files anymore, that's bad. I think loading certs from files is the basic operation that we should support. :(
> 
> Rest LGTM

On a discussion with @YangKeao we agree that it will be much better to use ca cert to sign a brand new cert and use it in `chaosctl`. That may be met by some shell script or so. But however is it done, specifying a file to load for `chaosctl` will require extra configuration, make it not that "out of the box". I say we may make it a second choice, specified in some env variable. What do you think of that?

STRRL:
> But however is it done, specifying a file to load for chaosctl will require extra configuration, make it not that "out of the box". I say we may make it a second choice, specified in some env variable. What do you think of that?

Yeah, I like "out of the box", but I still want I could load it from files. I think we should keep the basic functionality first, then improve the user experience.

Maybe appending a few flags could resolve that, I like flags more than ENVs. 😜

pkg/chaosctl/common/common.go
AsterNighT: That does not seem to work since they are both empty string 🤦

AsterNighT:
@STRRL I add three flags to it. The implementation is ugly, very ugly. Do you have any ideas on that?

pkg/chaosctl/common/common.go
```diff
@@ -407,18 +414,62 @@ Please check network policy / firewall, or see FAQ on website`, daemon.Name, dae
...
+		PrettyPrint("TLS Files are not complete, fall back to use secrets.", 0, Green)
+		config, err := getTLSConfigFromSecrets()
+		if err != nil {
+			return nil, err
+		}
+		return grpcUtils.CreateGrpcConnectionFromRaw("localhost", port, config.caCert, config.cert, config.key)
+	}
+	PrettyPrint("Using TLS Files.", 0, Green)
+	return grpcUtils.CreateGrpcConnection("localhost", port, TLSFiles.CaCert, TLSFiles.Cert, TLSFiles.Key)
+}
```
STRRL, line 444: I think we only need `grpcUtils.CreateGrpcConnectionFromRaw`, you could make a method called `getTLSConfigFromFiles()`;

STRRL:
>The implementation is ugly, very ugly. Do you have any ideas on that?

Actually, the codes are OK for me. 🙈 And I have no idea before making a refactor on`chaosctl`. About the refactor:

- create struct for each operation with the minimum, necessary states
- let `DebugOption` only stands for configuration, and remove the logic from it
- something like `chaos daemon client pool` or `chaos daemon client factory`, which holds the tls cert/key

pkg/chaosctl/common/common.go
AsterNighT: `grpcUtils.CreateGrpcConnection` is used elsewhere, I cannot simply delete it. But I can make it packed up around `grpcUtils.CreateGrpcConnectionFromRaw`

AsterNighT:
> > The implementation is ugly, very ugly. Do you have any ideas on that?
> 
> Actually, the codes are OK for me. 🙈 And I have no idea until making a refactor on `chaosctl`. About the refactor:
> 
> * create struct for each operation with the minimum, necessary states
> * let `DebugOption` only stands for configuration, and remove the logic from it
> * something like `chaos daemon client pool` or `chaos daemon client factory`, which holds the tls cert/key

So we'll just get by 🤦

AsterNighT:
Fix #1618

STRRL (commented):
Rest LGTM

I think it's time to clean my previous TODOs 🙈

pkg/grpc/utils.go
```diff
@@ -37,31 +37,44 @@ var log = ctrl.Log.WithName("util")
...
-		clientCert, err := tls.LoadX509KeyPair(certPath, keyPath)
+		cert, err = ioutil.ReadFile(certPath)
 		if err != nil {
 			return nil, err
 		}
+		key, err = ioutil.ReadFile(keyPath)
+		if err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
AsterNighT: I thought that would break some backward compatibility with secrue mode off. But I'm not sure.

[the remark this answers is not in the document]
STRRL: 🙈 I forget the situation without secure mode. And the current codes still could not be compatible without secure mode, it will return an error when parsing cert-key pair with cert and key as `nil`.😥

I think the better way is determining using secure mode at first, if we do not need secure mode, we create a connection by another function without cert things. And if we need secure mode, then load the certs (from file or from secrets in kubernetes) as []byte, then create a connection by `CreateGrpcConnectionFromRaw`. 

And I think each switch or flag at least could be explicitly controlled by the user, for example, if we could not determine using secure mode or not, just add a command flag. After finished the basic functionality, then we could do something easy-to-use.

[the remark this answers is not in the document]
AsterNighT: Ah yes I forgot the part. About the not-secure mode, the original implementation (I mean before this pr) is to call `CreateGrpcConnection` with no cert files specified. The function `CreateGrpcConnection` is used somewhere else in chaosmesh too. I thought it would be better if we would keep this feature.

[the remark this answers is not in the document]
STRRL: You are right. I am trying to refactor that in this new feature PR, that's bad.

AsterNighT:
Now `install.sh` set `SecureMode` to `false` by default, which make tls disabled in tests. Is it intended or simply no one has got a change to fix it?

AsterNighT:
@WangXiangUSTC And I see `install.sh` is indirectly tested by being called in `hack/local-up-chaos-mesh.sh`. Is it enough for us to delete `script_test.yml`?

STRRL (approved):

pkg/chaosctl/common/common.go
Yiyiyimu: ok then just add a todo 🤣

AsterNighT:
O some one come to merge this @YangKeao @cwen0

YangKeao (approved):
LGTM

cwen0:
@AsterNighT Can you cherry-pick this pr to release-1.2 branch manualy?

AsterNighT:
> @AsterNighT Can you cherry-pick this pr to release-1.2 branch manualy?

Done. And 1.1 does not seem to have #1504.

PR #1609, opened 2021-03-29
Revert workflow stuff

STRRL:

cleanup codes about workflow for the branch release-1.2

revert all the commits about workflow


STRRL:
I choose the bad branch, I will create another PR and close it now.

PR #1610, opened 2021-03-29
Revert workflow stuff

STRRL:

cleanup codes for release 1.2, which should not contain anything about workflow.

revert all the commits about the workflow


cwen0 (approved):
LGTM

PR #1616, opened 2021-03-29
[update] batch delete archives and experiments

fewdan:

Modify the interfaces for deleting experiments and archives.

Support batch selection.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

- [x] Need to modify the front end to send the request

```release-note
NONE
```


fewdan:
@g1eny0ung
After modifying these two interfaces, need to modify the front-end sending request. Please take a look at it when it is convenient for you.

pkg/apiserver/archive/archive.go
```diff
@@ -57,7 +58,7 @@ func Register(r *gin.RouterGroup, s *Service) {
 	endpoint.GET("", s.list)
 	endpoint.GET("/detail", s.detail)
 	endpoint.GET("/report", s.report)
-	endpoint.DELETE("/:uid", s.delete)
+	endpoint.DELETE("", s.delete)
```
g1eny0ung, line 61: Can we support both `DELETE /:uid` and `Delete /?uids=x,x,x` because I think It's better to keep the restful style.

g1eny0ung: I will test later in my local env.

g1eny0ung: After testing, I found that it can be done.

fewdan: Let me see, there are some minor changes.

fewdan:
@g1eny0ung 
Revised a bit, please take a look when it is convenient.

pkg/apiserver/archive/archive.go
```diff
@@ -278,38 +279,62 @@ func (s *Service) report(c *gin.Context) {
...
+	}
+	for _, uid := range uidSlice {
+		if exp, err = s.archive.FindByUID(context.Background(), uid); err != nil {
+			if gorm.IsRecordNotFoundError(err) {
+				_ = c.Error(utils.ErrInternalServer.WrapWithNoMessage(fmt.Errorf("delete archive uid (%s) error, because the archive is not found", uid)))
+			} else {
+				_ = c.Error(utils.ErrInternalServer.WrapWithNoMessage(fmt.Errorf("delete archive uid (%s) error, because %s", uid, err.Error())))
+			}
+			errFlag = true
+			continue
```
g1eny0ung, line 322: Should replace `continue` with `break`?

Suppose there are two errors that will occur. After the range loop end, we will set the context error twice.

g1eny0ung: The same goes for deleting experiments.

pkg/apiserver/archive/archive.go
```diff
@@ -278,38 +279,62 @@ func (s *Service) report(c *gin.Context) {
...
 	}
 
-	if err = s.archive.Delete(context.Background(), exp); err != nil {
-		c.Status(http.StatusInternalServerError)
-		_ = c.Error(utils.ErrInternalServer.WrapWithNoMessage(err))
-	} else {
+	if uid != "" {
+		uidSlice = append(uidSlice, uid)
+	}
+	for _, uid := range uidSlice {
```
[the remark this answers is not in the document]
g1eny0ung: After searching, if we delete a row that doesn't exist, then it will be treated as unaffected in SQLite. So the above I said about error handling can be ignored. 😆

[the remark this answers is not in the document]
fewdan: understand, let me modify it.

pkg/apiserver/archive/archive.go
g1eny0ung: https://pkg.go.dev/github.com/gin-gonic/gin#Context.Error

pkg/apiserver/archive/archive.go
```diff
@@ -278,38 +279,62 @@ func (s *Service) report(c *gin.Context) {
 // @Description Delete the specified archived experiment.
 // @Tags archives
 // @Produce json
-// @Param uid path string true "uid"
+// @Param uid path string false "uid"
```
AsterNighT, line 282: That is so weird. Can we delete this path later?

AsterNighT: Or make it `//@Router /archives [delete]` only, and split `//@Router /archives/{uid} [delete]` to another api

g1eny0ung: Agree to split them.

g1eny0ung (approved):
LGTM

pkg/apiserver/archive/archive.go
```diff
@@ -315,3 +317,42 @@ func (s *Service) delete(c *gin.Context) {
...
+		c.Status(http.StatusBadRequest)
+		_ = c.Error(utils.ErrInternalServer.WrapWithNoMessage(fmt.Errorf("uids cannot be empty")))
+		return
+	}
+	uidSlice = strings.Split(uids, ",")
+	errFlag := false
+
+	if err = s.archive.DeleteByUIDs(context.Background(), uidSlice); err != nil {
+		_ = c.Error(utils.ErrInternalServer.WrapWithNoMessage(err))
+		c.Status(http.StatusInternalServerError)
```
cwen0, line 346: Why not return directly?

pkg/apiserver/experiment/experiment.go
```diff
@@ -922,12 +923,108 @@ func (s *Service) getExperimentDetail(c *gin.Context) {
...
+			return
+		}
+
+		annotations := chaosMeta.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string)
+		}
+		annotations[common.AnnotationCleanFinalizer] = common.AnnotationCleanFinalizerForced
+		chaosMeta.SetAnnotations(annotations)
+		if err := kubeCli.Update(context.Background(), chaosKind.Chaos); err != nil {
```
AsterNighT, line 988: Why don't we need `retryOnConflict` here?

fewdan: @AsterNighT Let me see how to deal with it.

fewdan: I will submit a pr again to solve this problem later, because there is still a forced deletion bug to be dealt with, and it can be dealt with together at that time.

cwen0 (approved):
LGTM

cwen0:
@fewdan PTAL

PR #1620, opened 2021-03-30
Add a script to test chaosctl

### LINKED ISSUE #1600: add a basic script to test chaosctl
## Feature Request

Currently `chaosctl` only got basic unit tests with pretty low coverage, so whenever we make some changes on it, normally we need to do the manual test. 

If we don't want to spend time mocking too many things, I think a basic e2e test is enough for this situation. For example, we could add a basic test like integration test, create a network chaos, and test if the output of `chaosctl` is as expected.

AsterNighT:
Fix #1600 

Add a script to test chaosctl


.github/workflows/script_test.yml
```diff
@@ -1,34 +0,0 @@
-name: Script Test
```
WangXiangUSTC, line 1: why remove this file? it is used to test some script

AsterNighT: The file runs network test on master and release 1.0, while the job `integration_test` covers the same test on master.  Since we are no longer maintaining 1.0, it becomes useless.

WangXiangUSTC: In fact, this file is used to test `install.sh`,  make sure this script can install Chaos Mesh success

AsterNighT: oops, in that case that seems necessary. I'll revert this.

fewdan (approved):
LGTM

PR #1621, opened 2021-03-30
feat: apply multi type of chaos nodes with code generation

STRRL:

Make workflow supports all the type of chaos

- new method which mapping `templateType` and the CRD Kind, reuse the `chaosKindMap`
- spawn the chaos object from `EmbedChaos` by `templateType`
- tests on dependency of `chaosKindMap`
- appending the example workflow with stress-chaos

### Checklist

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


STRRL:
I think this PR is ready for review. 😁
/cc @chaos-mesh/committers 
/cc @chaos-mesh/maintainers

helm/chaos-mesh/values.yaml
```diff
@@ -249,7 +249,7 @@ dashboard:
         tlsSecret: dashboard.local-tls
 
 dnsServer:
-  create: false
+  create: true
```
Colstuwjx, line 252: It changed the default behavior of the chart installation. Why do we need to update this line inside this PR?

pkg/workflow/controllers/chaos_node_reconciler.go
```diff
@@ -51,51 +49,32 @@ func (it *ChaosNodeReconciler) Reconcile(request reconcile.Request) (reconcile.R
 		return reconcile.Result{}, client.IgnoreNotFound(err)
 	}
 
+	if !v1alpha1.IsChoasTemplateType(node.Spec.Type) {
+		return reconcile.Result{}, nil
```
Colstuwjx, line 53: Should we return an error here?

Colstuwjx (commented):
rest LGTM.

pkg/workflow/controllers/chaos_node_reconciler.go
STRRL: No, this reconciler would receive events about all workflownode, event type is not `Type<Some>Chaos`, and the `v1alpha1.IsChoasTemplateType()` plays as a filter.

STRRL:
Hi, as the GcpChaos has been merged https://github.com/chaos-mesh/chaos-mesh/pull/1526, this PR updates the definition of gcp chaos, so it need review again.

/cc @chaos-mesh/committers @chaos-mesh/reviewers

YangKeao (approved):
LGTM

cwen0 (approved):
LGTM

STRRL:
PTAL @AsterNighT @Colstuwjx

Issue #1597, opened 2021-03-25
Issues about applying kubernetes optimistic lock


AsterNighT:
Optimistic locks (like `retry.RetryOnConflict`) are adopted widely in chaos-mesh. But there are misuses and still somewhere in lack of locks. Take this as an example

https://github.com/chaos-mesh/chaos-mesh/blob/6ee0cb9f67c671eefb9110a99e021a260e2f5ac7/controllers/common/types.go#L112-L114

from my knowledge this retry will not work. And a lack of lock here

https://github.com/chaos-mesh/chaos-mesh/blob/6ee0cb9f67c671eefb9110a99e021a260e2f5ac7/controllers/common/types.go#L129-L132

I believe this is somewhat a reason for #1559, I will try to reproduce it locally.

AsterNighT:
@WangXiangUSTC PTAL

AsterNighT:
After a talk with @YangKeao, we decide this could be delayed and wait for a controller refactor in #1578

YangKeao:
Oops. The original issue was closed by the bot https://github.com/chaos-mesh/chaos-mesh/issues/791 :facepalm:

AsterNighT:
LOL

Issue #1559, opened 2021-03-12
miss the status information when getting details of chaos experiment


WangXiangUSTC:
## Bug Report

**What version of Chaos Mesh are you using?**

**What did you do?**

run integration test
https://github.com/chaos-mesh/chaos-mesh/runs/2092659499?check_suite_focus=true

get the details of chaos experiment
**What did you expect to see?**

can see the experiment's status 

**What did you see instead?**

the status is empty:

```log
***** get details of chaos experiments *****
all token can view the experiments under namespace busybox
0. use token *** to send GET request to /api/experiments/detail/991c2d5a-cd50-4645-81e6-cb49c19f5900?namespace=busybox, and save log in exp_detail.out, log should contains 'Running'
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   471  100   471    0     0  42818      0 --:--:-- --:--:-- --:--:-- 42818
TEST FAILED: exp_detail.out DOES NOT CONTAIN 'Running'
____________________________________
{"kind":"NetworkChaos","namespace":"busybox","name":"ci-test","uid":"991c2d5a-cd50-4645-81e6-cb49c19f5900","created":"2021-03-12T06:26:57Z","status":"","yaml":{"apiVersion":"chaos-mesh.org/v1alpha1","kind":"NetworkChaos","metadata":{"name":"ci-test","namespace":"busybox","labels":null,"annotations":null},"spec":{"action":"delay","mode":"one","value":"","selector":{"namespaces":["busybox"]},"delay":{"latency":"1ms","correlation":"0","jitter":"0ms"},"direction":"to"}}}^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

AsterNighT:
I ran into this https://github.com/chaos-mesh/chaos-mesh/runs/2180951187 and seem to fix it with a proper `/api/experiment/update` handler. But I don't think that is the reason.

PR #1622, opened 2021-03-30
Workaround for #1559

AsterNighT:
Signed-off-by: AsterNighT <klxjt99@outlook.com>

#1559, but it is not a fix.
This is a workaround for #1559, to make the integration test more stable. The final fix for this issue will land after #1578, as mentioned in #1597.


AsterNighT:
That does not seem to work

Issue #1354, opened 2021-01-06
chaos target support more than one container


WangXiangUSTC:
## Feature Request

**Is your feature request related to a problem? Please describe:**
https://github.com/chaos-mesh/chaos-mesh/issues

**Describe the feature you'd like:**

Now can only choose one container in Pod to inject chaos, need to support specify more than one container.

DarthBenro008:
Hey There!
I have noticed currently Chaos-Mesh takes in only a single container name in spec, the first container if none is mentioned in spec otherwise whatever ContainerName is mentioned in the spec

https://github.com/chaos-mesh/chaos-mesh/blob/6d045300ab546a60ac2ba6896032eb9331fb24cb/controllers/stresschaos/types.go#L191-L204

For this feature to be implemented, are we planning to just inject the stressors into all containers (a Boolean in the YAML configs)? or allow the user to specify a set of Container Names? (maybe an array of container names in YAML)

I would like to work on this issue 😄

WangXiangUSTC:
Thanks! I think we can implement these two features: 1. inject into all containers 2. specify a set of container names
@cwen0 @YangKeao @zhouqiang-cl @STRRL  how about you?

YangKeao:
> For this feature to be implemented, are we planning to just inject the stressors into all containers

Wow, really cool :beers: ! Some of these chaos are container-wise (like `StressChaos` and `IOChaos`) and some of them are pod-wise (like `NetworkChaos`). Could we have a unified way to call the implementation of them (rather than calling `SelectPods` or `SelectContainers` in the `Apply` functions.)

DarthBenro008:
@YangKeao maybe we could have a utility function, which takes in 3 parameters
- `Pod Name`
- `Type of Chaos to be injected`
- `List of Container names`

Then that function can apply that particular chaos to the required containers.
This util function can maybe be called from `types.go` of each Chaos Type that you had mentioned above.

YangKeao:
This issue has been fixed since June. Every chaos in container scope is selected with `containerNames: []string` to select multiple containers at the same time :smiley:.

PR #1635, opened 2021-04-02
target support more than one container added

Debanitrkl:
Two features implemented in stresschaos
1. inject into all containers
2. specify a set of container names
We solved the issue mentioned in #1354 

changed  `ContainerName *string` to `Containers []string`, it means we can now inject any containers as we want. Implemented in StressChaos
### Checklist

Tests


STRRL:
Hi @Debanitrkl, seem there are some extra changes should not be commited, like `helm/chaos-mesh/crds/chaos-mesh.org_helloworldchaos.yaml`, and the modification with `pkg/chaosdaemon/pb`, please cleanup that changes. 😁

STRRL:
And the modification of `Makefile` should also be cleaned up.

And yous should execute `make check` and `./hack/update_install_script.sh`, and commit the changes.

- `make check` will format your codes and take a static analyze
- `./hack/update_install_script.sh` will update the `install.sh` with the modification of crd.yaml and helm templates
-

Debanitrkl:
@STRRL I'm not getting the merge conflict where it is occcuring

PR #1641, opened 2021-04-05
target support more than one container added

Debanitrkl:
Two features implemented in stresschaos
1. inject into all containers
2. specify a set of container names
We solved the issue mentioned in #1354 

changed  `ContainerName *string` to `Containers []string`, it means we can now inject any containers as we want. Implemented in StressChaos
### Checklist

Tests


STRRL:
Hi @Debanitrkl, thanks for your contribution! As this PR will make break changes and it might conflict with the refactor of the selectors, it's hard to merge.

I am going to close the PR, feel free to comments or re-open it.

PR #1647, opened 2021-04-06
add dashboard hostNetwork option in helm

fewdan:

add dashboard hostNetwork option in helm

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [x] No code

Side effects

Related changes

```release-note
NONE
```


yeya24 (approved):
The change LGTM. But what's the use case of running the dashboard in host network mode?

STRRL:
> But what's the use case of running the dashboard in host network mode?

Same question for me.

fewdan:
> > But what's the use case of running the dashboard in host network mode?
> 
> Same question for me.

@yeya24 @STRRL 
I encountered a situation yesterday. It may be a problem with the CNI configuration (flannel) of the user cluster, which caused the pod to fail to connect to the api-server. Changing to hostNetwork can temporarily solve this user's problem.

STRRL:
That issue caused by the user's non-standard environment, and I think it's not necessary to change the default network mode. 🙈

fewdan:
> That issue caused by the user's non-standard environment, and I think it's not necessary to change the default network mode. 🙈

Yes, so when installing through helm, the default network mode is still "hostNetwork:false".

STRRL (approved):

PR #1661, opened 2021-04-09
Using builder for grpc connection

### LINKED ISSUE #1627: A better CreateGrpcConnection
## Feature Request

On a discussion with @STRRL we agrees on that grpc creation is getting awkward due to the secure mode of chaos-mesh. I think it may be better if we make it some `GrpcConnectionBuilder`.

AsterNighT:

Fix #1627 
Now there's a builder for grpc connection for more extensibility.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


pkg/grpc/utils.go
```diff
@@ -35,67 +35,136 @@ var RPCTimeout = DefaultRPCTimeout
...
+	return it
+}
+
+func (it *GrpcBuilder) TryTLSFromFiles(caCertPath string, certPath string, keyPath string) *GrpcBuilder {
+	if caCertPath == "" || certPath == "" || keyPath != "" {
+		return it.Insecure()
+	}
+	caCert, err := ioutil.ReadFile(caCertPath)
+	if err != nil {
+		return it.Insecure()
```
WangXiangUSTC, line 80: the error is not reported, should we save it in t.err?

pkg/grpc/utils.go
```diff
@@ -35,67 +35,136 @@ var RPCTimeout = DefaultRPCTimeout
...
+	it.port = port
+	return it
+}
+
+func (it *GrpcBuilder) Insecure() *GrpcBuilder {
+	it.options = append(it.options, grpc.WithInsecure())
+	return it
+}
+
+func (it *GrpcBuilder) TryTLSFromFiles(caCertPath string, certPath string, keyPath string) *GrpcBuilder {
```
WangXiangUSTC, line 74: what is the difference between `TryTLSFromFiles` and `TLSFromFiles`?

AsterNighT: `TryTLSFromFiles` is somewhat similar to the original behavior of our implementation. If the specified path is not alright, it falls back to use an insecure connection. `TLSFromFiles` is more strict, it reports the error and refuse to continue rather than perform a fall back.

STRRL (changes requested):
Please move the required parameters like `address`, `port` into the constructor of the builder, only leave the optional options as methods.

And I think we do not need `TryTLSFromFiles`, the logic about "if the path of certs is invalid, fallback to insecure mode" is out of this module, the caller should determine to use TLS or not before building a grpc connection.

Another thing is the `err error` in `GrpcBuilder`, I think it might be a little confusing in here. You store the error that happens before `Build()` into this field, I suggest do not actually do anything until calling `Build()`.

AsterNighT:
> Please move the required parameters like `address`, `port` into the constructor of the builder, only leave the optional options as methods.
> 
> And I think we do not need `TryTLSFromFiles`, the logic about "if the path of certs is invalid, fallback to insecure mode" is out of this module, the caller should determine to use TLS or not before building a grpc connection.
> 
> Another thing is the `err error` in `GrpcBuilder`, I think it might be a little confusing in here. You store the error that happens before `Build()` into this field, I suggest do not actually do anything until calling `Build()`.

Umm, so what do we do to the errors like `file does not exist` or `cert is invalid` during creation?

STRRL:
> > Please move the required parameters like `address`, `port` into the constructor of the builder, only leave the optional options as methods.
> > And I think we do not need `TryTLSFromFiles`, the logic about "if the path of certs is invalid, fallback to insecure mode" is out of this module, the caller should determine to use TLS or not before building a grpc connection.
> > Another thing is the `err error` in `GrpcBuilder`, I think it might be a little confusing in here. You store the error that happens before `Build()` into this field, I suggest do not actually do anything until calling `Build()`.
> 
> Umm, so what do we do to the errors like `file does not exist` or `cert is invalid` during creation?

The method `Build()` could return an error, do not actually read something at `TlsFromFile()`, do it at `Build（）`

AsterNighT:
> > > Please move the required parameters like `address`, `port` into the constructor of the builder, only leave the optional options as methods.
> > > And I think we do not need `TryTLSFromFiles`, the logic about "if the path of certs is invalid, fallback to insecure mode" is out of this module, the caller should determine to use TLS or not before building a grpc connection.
> > > Another thing is the `err error` in `GrpcBuilder`, I think it might be a little confusing in here. You store the error that happens before `Build()` into this field, I suggest do not actually do anything until calling `Build()`.
> > 
> > 
> > Umm, so what do we do to the errors like `file does not exist` or `cert is invalid` during creation?
> 
> The method `Build()` could return an error, do not actually read something at `TlsFromFile()`, do it at `Build()`

That'll make `Build()` a mess at last. I couldn't say that's a very good idea.

About the error storing method for now, I think it's ok to return the first error you met while building a connection?

STRRL:
> > > > Please move the required parameters like `address`, `port` into the constructor of the builder, only leave the optional options as methods.
> > > > And I think we do not need `TryTLSFromFiles`, the logic about "if the path of certs is invalid, fallback to insecure mode" is out of this module, the caller should determine to use TLS or not before building a grpc connection.
> > > > Another thing is the `err error` in `GrpcBuilder`, I think it might be a little confusing in here. You store the error that happens before `Build()` into this field, I suggest do not actually do anything until calling `Build()`.
> > > 
> > > 
> > > Umm, so what do we do to the errors like `file does not exist` or `cert is invalid` during creation?
> > 
> > 
> > The method `Build()` could return an error, do not actually read something at `TlsFromFile()`, do it at `Build()`
> 
> That'll make `Build()` a mess at last. I couldn't say that's a very good idea.

At least it could not be more mess than using the origin grpc library directly. 🤔

> About the error storing method for now, I think it's ok to return the first error you met while building a connection?

The current implementation does make the thing as "return the first error", I prefer to report the logs as soon as it occurs.

The mess or the quality of codes and the expected behaviors of the codes are not related, I think we could use tricks to abstract more things when you find it is too complex.

I have another idea about that:

```go

type TransportCredentialLoader interface {
	GetTransportCredentials() (grpc.DialOption, error)
}

// Insecure, TLSFromRaw and TLSFromFile will implement TransportCredentialLoader
type Insecure struct {

}

type TLSFromRaw struct {
	caCert []byte
	cert []byte
	key []byte
}

type TLSFromFile struct {
	caCertPath string
	certPath string
	keyPath string
}

type GrpcBuilder struct {
	// other fields
	transportCredentialLoader TransportCredentialLoader
}

func (it *GrpcBuilder) Build() (*grpc.ClientConn, error) {
	credentialOption, err := it.transportCredentialLoader.GetTransportCredentials()
	// if err!= nil; return nil, err
	options := append(it.options, credentialOption)
	return grpc.Dial(fmt.Sprintf("%s:%d", it.address, it.port), options...)
}

```

AsterNighT:
PTAL @STRRL

pkg/grpc/utils.go
```diff
@@ -35,42 +35,72 @@ var RPCTimeout = DefaultRPCTimeout
...
+	}
+	if it.fromType == RAW {
+		return it.TLSFromRaw()
 	}
-	options := []grpc.DialOption{grpc.WithUnaryInterceptor(TimeoutClientInterceptor)}
-	options = append(options, grpc.WithInsecure())
-	conn, err := grpc.Dial(fmt.Sprintf("%s:%d", address, port), options...)
+	if it.fromType == FILE {
+		return it.TLSFromFile()
+	}
```
STRRL, line 73: I prefer to use three implementations instead of if-else codes, and it is called "replace conditional with polymorphism". 

When more situations coming, we could just use appending without modifying to extend our application.

As the codes are not so complex, they're OK to me.

pkg/grpc/utils.go
```diff
@@ -80,22 +110,85 @@ func CreateGrpcConnectionFromRaw(address string, port int, caCert []byte, cert [
...
+
+func (it *GrpcBuilder) Address(address string) *GrpcBuilder {
+	it.address = address
+	return it
+}
+
+func (it *GrpcBuilder) Port(port int) *GrpcBuilder {
+	it.port = port
+	return it
+}
```
STRRL, line 145: How about removing these two methods?

pkg/grpc/utils.go
AsterNighT: Ahh I don't get it completely. Could you provide a small piece of example? `foobar` would make sense.

STRRL: Here is an example of that: https://refactoring.com/catalog/replaceConditionalWithPolymorphism.html, this example uses a field that is an instance of `Swallow` instead of `bird.type`.

"Refactoring" only improves efficiency and maintenance, it will not change the usage of the origin thing.

AsterNighT:
@STRRL I think you mean something like this?

pkg/grpc/utils.go
```diff
@@ -35,42 +35,65 @@ var RPCTimeout = DefaultRPCTimeout
...
+	Cert   string
+	Key    string
+}
+
+type TLSFromType = string
+
+const (
+	RAW  TLSFromType = "RAW"
+	FILE TLSFromType = "FILE"
+)
```
STRRL, line 55: > @STRRL I think you mean something like this?

Yes! And we do not need these lines anymore.

AsterNighT:
After the update it seems that `kubectl set env deployment/chaos-controller-manager SECURITY_MODE=true -n chaos-testing` causes trouble by tricking controller manager into supposing it is using tls. What is it intended to do originally?

STRRL:
The `SECURITY_MODE` represents we need to authorize each request with kubernetes RBAC, reference: https://github.com/chaos-mesh/chaos-mesh/blob/master/pkg/config/controller.go#L69

STRRL:
It seems we need another flag in configuration with using TLS on chaos-daemon, what do you think about it @YangKeao ?

YangKeao:
> It seems we need another flag in configuration with using TLS on chaos-daemon, what do you think about it @YangKeao ?

I agree that using the same flag to control TLS and RBAC is really a bad idea :crying_cat_face: .

Always turning on TLS between `chaos-daemon` and `controller-manager` could also be a choice, as it brings nothing bad, does it?

AsterNighT:
> > It seems we need another flag in configuration with using TLS on chaos-daemon, what do you think about it @YangKeao ?
> 
> I agree that using the same flag to control TLS and RBAC is really a bad idea 😿 .
> 
> Always turning on TLS between `chaos-daemon` and `controller-manager` could also be a choice, as it brings nothing bad, does it?

Couldn't agree more.

STRRL:
> Always turning on TLS between chaos-daemon and controller-manager could also be a choice, as it brings nothing bad, does it?

Agree, we could do that in another PR. But it will ask for more works while debugging, any suggestion?

AsterNighT:
> > Always turning on TLS between chaos-daemon and controller-manager could also be a choice, as it brings nothing bad, does it?
> 
> Agree, we could do that in another PR. But it will ask for more works while debugging, any suggestion?

It will cause trouble in chaosdashboard, but it seems to do no harm if you use kubectl only.

YangKeao:
> > > Always turning on TLS between chaos-daemon and controller-manager could also be a choice, as it brings nothing bad, does it?
> > 
> > 
> > Agree, we could do that in another PR. But it will ask for more works while debugging, any suggestion?
> 
> It will cause trouble in chaosdashboard, but it seems to do no harm if you use kubectl only.

Chaos Dashboard does nothing about `chaos-daemon`, why it would cause trouble in chaosdashboard?

AsterNighT:
> > > > Always turning on TLS between chaos-daemon and controller-manager could also be a choice, as it brings nothing bad, does it?
> > > 
> > > 
> > > Agree, we could do that in another PR. But it will ask for more works while debugging, any suggestion?
> > 
> > 
> > It will cause trouble in chaosdashboard, but it seems to do no harm if you use kubectl only.
> 
> Chaos Dashboard does nothing about `chaos-daemon`, why it would cause trouble in chaosdashboard?

Sorry I mistaken that for RBAC 🤦 . That way it seems does no harm at all.

AsterNighT:
@STRRL  PLTAL

STRRL (approved):
LGTM

pkg/grpc/utils.go
```diff
@@ -35,42 +35,58 @@ var RPCTimeout = DefaultRPCTimeout
...
+type FileProvider struct {
+	file TLSFile
+}
+
+type RawProvider struct {
+	raw TLSRaw
+}
+
+type InsecureProvider struct {
+}
```
cwen0, line 59: Here needs a blank line

pkg/grpc/utils.go
```diff
@@ -35,42 +35,58 @@ var RPCTimeout = DefaultRPCTimeout
...
-	return conn, nil
-}
 
-// CreateGrpcConnectionFromRaw create a grpc connection with given port and address, and use raw data instead of file path
-func CreateGrpcConnectionFromRaw(address string, port int, caCert []byte, cert []byte, key []byte) (*grpc.ClientConn, error) {
-	options := []grpc.DialOption{grpc.WithUnaryInterceptor(TimeoutClientInterceptor)}
+	creds := credentials.NewTLS(&tls.Config{
+		Certificates: []tls.Certificate{clientCert},
+		RootCAs:      caCertPool,
+		ServerName:   "chaos-daemon.chaos-mesh.org",
```
cwen0, line 80: Add a const to define it

cwen0 (commented):
Rest LGTM

pkg/grpc/utils.go
AsterNighT: fixed

pkg/grpc/utils.go
```diff
@@ -35,67 +35,147 @@ var RPCTimeout = DefaultRPCTimeout
...
+	if it.credentialProvider == nil {
+		return nil, fmt.Errorf("an authorization method must be specified")
+	}
+	option, err := it.credentialProvider.getCredentialOption()
 	if err != nil {
 		return nil, err
 	}
-	return conn, nil
+	it.options = append(it.options, option)
+	return grpc.Dial(fmt.Sprintf("%s:%d", it.address, it.port), it.options...)
```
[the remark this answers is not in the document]
AsterNighT: really?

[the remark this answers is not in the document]
TomMD: I've seen this semgrep warning a lot and wondered if it should be removed from muse-dev.  Do you have a strong opinion on it?  Perhaps this warning is noise to you because the use of `use.net.JoinHostPort` not materially different from `Sprintf()`, or perhaps `JoinHostPort` is an unfamiliar but appropriate solution?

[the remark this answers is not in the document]
TomMD: Looking at [the Go documentation](https://golang.org/pkg/net/#JoinHostPort) I see it mentions how JoinHostPort can help when the IP is IPv6. Supposing that is a legitimate consideration, maybe this error best remain but the message be improved with the reason JoinHostPort is preferred.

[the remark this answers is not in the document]
AsterNighT: Surely the message need improvement. It would be much better if I don't have to google it myself to find out why I'm wrong about this 😆 . This would apply to all muse bug reports. A brief explaination or a link to some reference would be helpful.

AsterNighT:
PTAL @YangKeao @cwen0 @WangXiangUSTC

YangKeao (approved):
LGTM

Issue #1679, opened 2021-04-13
Workflows are denied by webhook "vauth.kb.io"


STRRL:
## Bug Report

`Workflow` is not a type of Chaos, so it does not implement the interface `ChaosValidator`. So all the request about workflow will fail because of that webhook:
```
$ k apply -f ./examples/workflow/workflow.yaml                 
Error from server: error when creating "./examples/workflow/workflow.yaml": admission webhook "vauth.kb.io" denied the request: kind Workflow is not support
```

The current implementation about auth depends on the `SelectSpec`, which is not so well.

I think we need another exactly interface to describe the permissions, like:

```golang
type PermissionScope interface {
	ClusterScoped() bool
	AffectedNamespaces() []string
}
```

Each type of chaos could persist the current implementation, and workflow will always return `false` with `ClusterScoped()` and return it's namespace with `AffectedNamespaces`.

STRRL:
I think we could bypass this issue by just allowing all the requests about workflow and workflow nodes.

STRRL:
closed by #1681

PR #1681, opened 2021-04-13
fix: let workflow and workflownode bypass the vauth webhook

STRRL:

It could bypass the #1679 temporarily;

when the Kind is `Workflow` or `WorkflowNode`, just response Allow.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


fewdan (approved):
LGTM

Colstuwjx:
Hi @STRRL

I have a question about this PR, is it permanently bypass the auth or just temporary work? If we said it's permanently, how about limiting the `resources` target in the [webhook configuration](https://github.com/chaos-mesh/chaos-mesh/blob/fd8ebcca7ac75a60d1cb80291f017cb278937886/api/webhook/validate_auth.go#L33)?

See [webhook configuration rules](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-rules), we may need to think about the resource group construction between chaos resource, chaos subresource ( e.g. PodIoChaos, PodNetworkChaos ), workflow resource.

STRRL:
> Hi @STRRL
> 
> I have a question about this PR, is it permanently bypass the auth or just temporary work? If we said it's permanently, how about limiting the `resources` target in the [webhook configuration](https://github.com/chaos-mesh/chaos-mesh/blob/fd8ebcca7ac75a60d1cb80291f017cb278937886/api/webhook/validate_auth.go#L33)?

It just a temporary work. :D  All resources should be authorized in expected, so I think use `*` as resources is OK.

STRRL:
conflicts resolved :D

PR #1683, opened 2021-04-14
Refactoring: creating children nodes with labels instead of updating status directly

STRRL:

Refactoring current implementation, not a new feature.
It could reduce the number of conflict, and prevent the unexpected duplicated creation of child node;

Currently, we record the state in Status, for example, when the controller react with ChaosNode with NetworkChaos embedded:

- create an NetworkChaos CR
- update the Status field

These actions could not be supported as a transaction, it means it's so easy for bringing state inconsistent because there might be something changed between actions. The created node might be ignored because the Status is overwritten.

> The same thing also happens with chaos nodes, it should also be refactored in another PR because it related more complicated code generation things.

This PR will not change the design and expected behavior, only refactor the implementation. 

The major change are: 

- Make Status as subresource.
- SerialNodeReconciler using query from kube-api for each times instead of using Status directly, the inconsistent between "real-world" and "Status" should not affect the behavior of SerialNodeReconciler.
- Use labels on each children nodes, instead of changing references in Status.
- Status of WorkflowNode is only designed to show the state for users and other components.
- Change the registration of SerialNodeReconciler with `EnqueueRequestForOwner`, for responding the  changes of children.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


pkg/workflow/controllers/serial_node_reconciler.go
```diff
@@ -52,93 +65,228 @@ func (it *SerialNodeReconciler) Reconcile(request reconcile.Request) (reconcile.
...
+	if err != nil {
+		return err
 	}
+	var taskToStartup string
+	if len(activeChildrenNodes) == 0 {
+		for index, task := range node.Spec.Tasks {
+			if index < len(finishedChildrenNodes) {
+				// TODO: if the definition/spec of task changed, we should also respawn the node
+				// child node start with task name
+				if strings.Index(finishedChildrenNodes[index].Name, task) != 0 {
```
Hexilee, line 153: Why don't use [HasPrefix](https://golang.org/pkg/strings/#HasPrefix).

Hexilee (commented):
Rest LGTM

pkg/workflow/controllers/serial_node_reconciler.go
```diff
@@ -52,93 +65,228 @@ func (it *SerialNodeReconciler) Reconcile(request reconcile.Request) (reconcile.
...
+		nodeNeedUpdate := v1alpha1.WorkflowNode{}
+		err := it.kubeClient.Get(ctx, request.NamespacedName, &nodeNeedUpdate)
+		if err != nil {
+			return err
+		}
+
+		// un-synced expected children
+		if nodeNeedUpdate.Status.ExpectedChildrenNum == nil || *nodeNeedUpdate.Status.ExpectedChildrenNum != len(nodeNeedUpdate.Spec.Tasks) {
+			expected := len(nodeNeedUpdate.Spec.Tasks)
+			nodeNeedUpdate.Status.ExpectedChildrenNum = &expected
```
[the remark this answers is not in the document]
STRRL: Yes, that's totally unused, let me remove that.

pkg/workflow/controllers/serial_node_reconciler.go
STRRL: Fixed.

STRRL:
PTAL @Hexilee 😆
/cc @Hexilee

Hexilee (approved):
LGTM

AsterNighT (commented):
LGTM but for one question. `status` subresource does not update `generation`. What do we do to ensure consistency?

STRRL:
> LGTM but for one question. `status` subresource does not update `generation`. What do we do to ensure consistency?

Actually, it does not guarantee any "consistency". Firstly, we do not make operations based on the state in `status`, we will re-observe the real world (for example, list the children workflow nodes) each time. So the status is only for the human users and other controllers. And the controller will keep trying to make that `status` as expected.

And I use the so-called `noCacheClient` for avoiding the inconsistency between controller-runtime and kubernetes API, https://github.com/chaos-mesh/chaos-mesh/blob/fa40abc70252a8e9cd4bd91f46e53caa7fb23b36/pkg/workflow/controllers/bootstrap.go#L27

pkg/workflow/controllers/serial_node_reconciler.go
```diff
@@ -52,93 +65,224 @@ func (it *SerialNodeReconciler) Reconcile(request reconcile.Request) (reconcile.
 		return reconcile.Result{}, nil
 	}
 
+	it.logger.V(4).Info("resolve serial node", "node", request)
+
+	err = it.syncChildrenNodes(ctx, node)
```
cwen0, line 70: Can you add some comments for `syncChildrenNodes` function? It is difficult for me to understand the logic and meaning of this function. I don't know what to sync and why to sync. 😂

pkg/workflow/controllers/serial_node_reconciler.go
```diff
@@ -52,93 +65,224 @@ func (it *SerialNodeReconciler) Reconcile(request reconcile.Request) (reconcile.
...
-		expected := len(node.Spec.Tasks)
-		node.Status.ExpectedChildrenNum = &expected
+	activeChildrenNodes, finishedChildrenNodes, err := it.fetchChildrenNodes(ctx, node)
+	if err != nil {
+		return err
 	}
+	var taskToStartup string
+	if len(activeChildrenNodes) == 0 {
+		for index, task := range node.Spec.Tasks {
+			if index < len(finishedChildrenNodes) {
```
cwen0, line 145: Why need to check `index < len(finishedChildrenNodes)`?

pkg/workflow/controllers/serial_node_reconciler.go
```diff
@@ -52,93 +65,224 @@ func (it *SerialNodeReconciler) Reconcile(request reconcile.Request) (reconcile.
...
-					return client.IgnoreNotFound(err)
-				}
-				SetCondition(&nodeNeedUpdate.Status, v1alpha1.WorkflowNodeCondition{
-					Type:   v1alpha1.ConditionAccomplished,
-					Status: corev1.ConditionTrue,
-					Reason: v1alpha1.NodeAccomplished,
-				})
-				return it.kubeClient.Update(ctx, &nodeNeedUpdate)
-			})
+				if strings.HasPrefix(task, finishedChildrenNodes[index].Name) {
```
cwen0, line 149: Why need to check the prefix here?

pkg/workflow/controllers/serial_node_reconciler.go
```diff
@@ -39,6 +44,14 @@ func NewSerialNodeReconciler(kubeClient client.Client, eventRecorder record.Even
 }
 
 func (it *SerialNodeReconciler) Reconcile(request reconcile.Request) (reconcile.Result, error) {
```
cwen0, line 46: Can you add some comments to help me understand the logic of this function?

STRRL:
Hi @cwen0, I append some comments, and going to mark some review conversations that might be explained as resolved, PTAL. 😁

pkg/workflow/controllers/serial_node_reconciler.go
```diff
@@ -52,93 +65,224 @@ func (it *SerialNodeReconciler) Reconcile(request reconcile.Request) (reconcile.
...
 
-		var childrenNames []string
-		for _, item := range childrenNodes {
-			childrenNames = append(childrenNames, item.Name)
+	var activeChildren []v1alpha1.WorkflowNode
+	var finishedChildren []v1alpha1.WorkflowNode
+
+	for _, item := range sortedChildrenNodes {
+		childNode := item
+		if WorkflowNodeFinished(childNode.Status) {
```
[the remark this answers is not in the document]
STRRL: If a chaos-node failed to apply the chaos, the `Finished` Condition could never be true, so that will pending the progress of workflow, which is expected. 😁

By the way, another condition `DeadlineExceed` will be not treated as finished after merging `nivara`.

We have not resolved the failed logic explicitly with all the workflow, such as sending some events or updating status.🙈
I think it should be introduced in another PR with a unified implementation.

cwen0 (approved):
LGTM

PR #1708, opened 2021-04-15
Fix chaosctl

AsterNighT:
It won't work properly now. Since it is originally designed upon `twophase`. This pr only intends to fix compile errors.


STRRL (approved):
LGTM

fewdan (approved):
LGTM

PR #1713, opened 2021-04-16
Add Schedule

### LINKED ISSUE #1693: Add `Schedule` custom resource to represent cron job
As discussed in  #1520, a `Schedule` resource is needed to represent a cron chaos.

YangKeao:

Fix #1693 

### TODO

- [x] Add `ConcurrencyPolicy`
- [x] Don't delete running chaos


cmd/chaos-builder/schedule.go
```diff
@@ -0,0 +1,36 @@
...
+type scheduleCodeGenerator struct {
+	// name of each Kind of chaos, for example: PodChaos, IoChaos, DNSChaos
+	chaosTypes []string
+}
+
+func newScheduleCodeGenerator(types []string) *scheduleCodeGenerator {
+	return &scheduleCodeGenerator{chaosTypes: types}
+}
+
+func (it *workflowCodeGenerator) AppendTypes(typeName string) {
```
AsterNighT, line 28: `scheduleCodeGenerator`?

cmd/chaos-builder/schedule.go
```diff
@@ -0,0 +1,36 @@
...
+	it.chaosTypes = append(it.chaosTypes, typeName)
+}
+
+func (it *workflowCodeGenerator) Render() string {
```
AsterNighT, line 32: Ditto

controllers/schedule/active/controller.go
```diff
@@ -0,0 +1,126 @@
...
+	ctx := context.Background()
+
+	schedule := &v1alpha1.Schedule{}
+	err := r.Get(ctx, req.NamespacedName, schedule)
+	if err != nil {
+		r.Log.Error(err, "unable to get chaos")
+		return ctrl.Result{}, nil
+	}
+
+	kind, ok := v1alpha1.AllKinds()[string(schedule.Spec.Type)]
```
AsterNighT, line 51: `v1alpha1.AllKinds()` seems to contains only chaos type only, workflow type is not supported yet.

controllers/schedule/gc/controller.go
```diff
@@ -0,0 +1,108 @@
...
+	ctx := context.Background()
+
+	schedule := &v1alpha1.Schedule{}
+	err := r.Get(ctx, req.NamespacedName, schedule)
+	if err != nil {
+		r.Log.Error(err, "unable to get chaos")
+		return ctrl.Result{}, nil
+	}
+
+	kind, ok := v1alpha1.AllKinds()[string(schedule.Spec.Type)]
```
AsterNighT, line 46: And here

examples/burn-cpu.yaml
```diff
@@ -8,12 +8,10 @@ spec:
...
       "app.kubernetes.io/component": "tikv"
-  stressors: 
-    cpu: 
+  stressors:
+    cpu:
       workers: 1
       load: 100
       options: ["--cpu 2", "--timeout 600", "--hdd 1"]
   duration: "30s"
-  scheduler:
```
AsterNighT, line 17: So we would not test scheduler seperately in every chaos. Instead we could test it once and for all?

YangKeao: Yes. But we nearly didn't test scheduler in previous tests :facepalm: .

I tried to write an integration test for the scheduler, through the `sigs.k8s.io/controller-runtime/pkg/envtest`. Would you like to help and make it easy to define and add case ? (And I think it could be added in another PR, as this one is big enough :facepalm: )

controllers/schedule/active/controller.go
YangKeao: As the support for `Workflow`, I need another discussion with @STRRL (as it needs the finished status to GC the history). I prefer to include it in another PR as this one is too big.

examples/burn-cpu.yaml
AsterNighT: Sure. I'll check it out.

AsterNighT: umm, I'm a bit coinfused. Does that mean we would use `envtest` to create a virtual cluster, register crd and controllers, and run tests on it?

YangKeao: Yes. Some tests are already using it, e.g. `api/v1alpha1/suite_test.go`.

AsterNighT: It seems that `suite_test.go` is good to use on crds, but it has no controllers. Could we just reuse that?

YangKeao:
It seems no one want to review it yet. But we have to review it in the final PR about merging nirvana into the master :stuck_out_tongue: 

Merge it to not block the development roadmap.

PR #1719, opened 2021-04-20
add check for whether duration is nil

YangKeao:

Without this patch, `controller-manager` will panic if the `Duration` field is `nil`.


STRRL:
How about appending a test case in `controllers/twophase/twophase_suite_test.go`, with nil duration.

YangKeao:
> How about appending a test case in `controllers/twophase/twophase_suite_test.go`, with nil duration.

Done :)

STRRL (approved):
LGTM

cwen0 (approved):
LGTM

PR #1724, opened 2021-04-20
Chaosctl nirvana

### LINKED ISSUE #1711: Refactor `chaosctl` to work with `nirvana`
Now in `nirvana`, `chaosctl` won't work properly since it rely on the `two phase` logic. It needs to be fixed. I think it's time to give it a total review and see if improvements can be made.

Known issues would be:
- Not very straight forward or useful, there is no script or guideline on how to use it in a bug report.
- Poor parameter passing. Some parameters are passed among modules using global variables.
- Very few document or comments. It's not easy to understand what it does.
- And more. It's open for discussion.

AsterNighT:
Surprisingly, a little fix would be sufficient to make chaosctl work!
fix #1711


AsterNighT:
@YangKeao PTAL

YangKeao (commented):
LGTM

PR #1725, opened 2021-04-20
Changes in controller rbac template file to avoid certain confusions

anuragpaliwal80:

It update the controller rbac template to avoid certain confusions regarding `clusterScoped` parameter. 

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


anuragpaliwal80:
I think we can remove `Role` -> `{{ .Release.Name }}-chaos-controller-manager-control-plane` as we are already assigning this rights to service account via `ClusterRole`. Let me know.

helm/chaos-mesh/templates/controller-manager-rbac.yaml
```diff
@@ -144,7 +141,9 @@ kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: {{ .Release.Name }}-chaos-controller-manager-target-namespace
+{{- if not .Values.clusterScoped }}
   namespace: {{ .Values.controllerManager.targetNamespace | quote }}
+{{- end }}
```
STRRL, line 146: I think I could not accept the changes except these. As we have talked on slack, here is indeed the dirty thing that I want to clean up. 😆

In this controller-manager-rbac.yaml file,  here are three `Role` or `ClusterRole`:

- `{{ .Release.Name }}-chaos-controller-manager-control-plane` represents the permission for the namespace which installed the chaos-mesh; It requires permissions on `serivces` and `configmaps`; it always use `RoleBinding` with service acount
- `{{ .Release.Name }}-chaos-controller-manager-cluster-level` represents the permission for the cluser-level, no matter with the `.Value.clusterScoped`, it always require the permission on `nodes` and `subjectaccessreviews`; it always use `ClueterRoleBinding` with service acount
- `{{ .Release.Name }}-chaos-controller-manager-target-namespace` represents the permission for the namespace which will be injected chaos; if `.Value.clusterScoped = true` it bind service account by `ClusterRoleBinding`, other with with `RoleBinding`

So we could not remove the `{{ .Release.Name }}-chaos-controller-manager-control-plane` even with `.Value.clusterScope = true`

anuragpaliwal80: If `.Value.clusterScope = true` then what ever permissions we are giving in `{{ .Release.Name }}-chaos-controller-manager-control-plane`  `Role` to resources like `configmaps`, `services` and `subjectaccessreviews`, these we are already giving these permission by `ClusterRole` as well.

anuragpaliwal80: We are giving the following permission in this `Role`:

```
rules:
  - apiGroups: [ "" ]
    resources: [ "configmaps", "services" ]
    verbs: [ "get", "list", "watch" ]
  - apiGroups: [ "authorization.k8s.io" ]
    resources:
      - subjectaccessreviews
    verbs: [ "create" ]
```

and these permissions are also part of two cluster roles as well.

anuragpaliwal80: Am I missing some thing ? 🤔

anuragpaliwal80:
In my cluster, I have even combined these two cluster roles and there is **no** role/rolebinding:
```
kind: ClusterRole
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: chaos-mesh:chaos-controller-manager-target-namespace-and-cluster-level
  labels:
    app.kubernetes.io/name: chaos-mesh
    app.kubernetes.io/instance: chaos-mesh
    app.kubernetes.io/component: controller-manager
rules:
- apiGroups: [ "" ]
  resources: [ "endpoints" ]
  verbs: [ "get", "list", "watch" ]
- apiGroups: [ "" ]
  resources: [ "pods" ]
  verbs: [ "get", "list", "watch", "delete", "update" ]
- apiGroups:
  - ""
  resources:
  - events
  verbs:
  - patch
  - create
- apiGroups: [ "" ]
  resources: [ "configmaps" ]
  verbs: [ "*" ]
- apiGroups: [ "chaos-mesh.org" ]
  resources:
  - "*"
  verbs: [ "*" ]
- apiGroups: [ "" ] # For cluster-level
  resources:
  - nodes
  - namespaces
  - services
  verbs: [ "get", "list", "watch" ]
- apiGroups: [ "authorization.k8s.io" ]
  resources:
  - subjectaccessreviews
  verbs: [ "create" ]
---
kind: ClusterRoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: chaos-mesh:chaos-controller-manager-target-namespace-and-cluster-level
  labels:
    app.kubernetes.io/name: chaos-mesh
    app.kubernetes.io/instance: chaos-mesh
    app.kubernetes.io/component: controller-manager
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: chaos-mesh:chaos-controller-manager-target-namespace-and-cluster-level
subjects:
- kind: ServiceAccount
  name: chaos-controller-manager
  namespace: chaos-testing
```

Just want to know, do you think I am missing any permission which might cause me any prob ?

STRRL:
> Just want to know, do you think I am missing any permission which might cause me any prob ?

No, I do not think you miss any permissions.

But if you want to use chaos mesh in namespace scoped mode, that rbac configuration will leak some permission to the cluster-wide.

anuragpaliwal80:
> But if you want to use chaos mesh in namespace scoped mode, that rbac configuration will leak some permission to the cluster-wide.

Let me try to fix this issue as well so that we don't leak any permission cluster-wide. I will also try to remove `Role` and `RoleBinding` all together.

STRRL:
I think we should still keep these `Role` and `ClusterRole` for description different permission for different usage, instead of combining them.

anuragpaliwal80:
Sure @STRRL , it was my view as it creates few confusion when it cluster scoped and I was thinking to remove it as I did for my cluster.
Let me close this ticket as you are fine keeping both.

PR #1744, opened 2021-04-23
Fix stresschaos e2e

### LINKED ISSUE #1742: Unstable e2e test: memory stress
## Bug Report

The e2e test case about memory stress is not stable:
https://github.com/chaos-mesh/chaos-mesh/blob/master/test/e2e/chaos/stresschaos/memory.go

We have disabled that case as a temporary solution. https://github.com/chaos-mesh/chaos-mesh/pull/1738

By the logging of e2e tests, the usage of memory does not decrease after recovering the Chaos, and the memory usage of another cgroup which was the "Non-Experimental Group"  dose increase.

This issue could be reproduced on a high-level kernel like 5.x, so it might be no relation with kernel 3.10.

It might be caused by the inconsistent value with cgroup of memory usage: `/sys/fs/cgroup/memory/memory.usage_in_bytes`

AsterNighT:
Signed-off-by: AsterNighT <klxjt99@outlook.com>

fix #1742

`cgroup/memory/usage_in_bytes`, if I did not misunderstand, consists of three parts: kernal memory, cache, and working set.

While kernal memory and cache seem ought to be excluded.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


test/pkg/fixture/fixture.go
```diff
@@ -199,7 +199,7 @@ func NewStressTestDeployment(name, namespace string, extraLabels map[string]stri
 								},
 								Limits: corev1.ResourceList{
 									corev1.ResourceCPU:    resource.MustParse("1"),
-									corev1.ResourceMemory: resource.MustParse("100M"),
+									corev1.ResourceMemory: resource.MustParse("150M"),
```
STRRL, line 202: Why do we need to increase that limit?

AsterNighT: There are occations that the pod with stress chaos would use more than 100M memory and gets OOMed. I'm not sure why this would happen.

STRRL (approved):
I'm not sure about that could absolutely fix the stress chaos testcases, let's take a look. 😆

LGTM

AsterNighT:
> I'm not sure about that could absolutely fix the stress chaos testcases, let's take a look. 😆
> 
> LGTM

I would call it a workaround instead of a decent fix 🤦

fewdan (approved):
LGTM

PR #1753, opened 2021-04-28
refactor the implementation of chaos nodes

STRRL:

It is extremely like #1683, but for chaos node reconcile.

> Unfortunately, this PR has been blocked by #1683 because of the sub-resource status, it will not pass the manual test until the serial node works.

It makes some similar refactor about using creating resources with certain labels instead of updating status directly.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


AsterNighT (approved):

YangKeao (approved):
LGTM

STRRL:
The e2e test case of memory stress is still unstable. :(

/run-e2e-tests

pkg/workflow/controllers/chaos_node_reconciler.go
```diff
@@ -53,20 +62,122 @@ func (it *ChaosNodeReconciler) Reconcile(request reconcile.Request) (reconcile.R
...
+		err := it.kubeClient.Get(ctx, request.NamespacedName, &nodeNeedUpdate)
+		if err != nil {
+			return client.IgnoreNotFound(err)
+		}
+
+		chaosList, err := it.fetchChildrenChaosCustomResource(ctx, nodeNeedUpdate)
+		if err != nil {
+			return client.IgnoreNotFound(err)
+		}
+		if len(chaosList) > 1 {
```
cwen0, line 83: Why do we not need this log when `len(chaosList) == 1` or `len(chaosList) == 0`?

STRRL: because `len(chaosList) > 1` is not expected, and only happens when the cache of controller-runtime is oudated. It's kind of a WARN log.

PR #1759, opened 2021-05-03
Escape HTML Content

soniasingla:

It escapes HTML Content :)


STRRL:
HI @soniasingla, thanks for your contribution.

But I think we should not use `html/template` in chaos-builder, chaos-builder is a tool for code generation, and it does not need HTML escape.

I'm going to close this PR, feel free to reopen it if you have other opinions. 😁

PR #1763, opened 2021-05-04
use JoinHostPort instead of Sprintf for network address

nescohen:
Fix SemGrep complaint to use JoinHostPort instead of Sprintf for network address construction.

From SemGrep:
> opt.semgrep.sprintf-host-port
> use net.JoinHostPort instead of fmt.Sprintf(port, dnsServerIP)

Using Sprintf to concatenate a network host and port can cause issues if an IPv6 address is used. It is generally recommended to use the JoinHostPort function instead. Here is a small toy example if you are interested: https://play.golang.org/p/sBFbWXL7RKE

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code
code still builds, no functionality was changed

Side effects

Related changes

```release-note
NONE
```


STRRL (approved):

cwen0 (approved):
LGTM

PR #1771, opened 2021-05-04
Remove duplicate imported file

soniasingla:
Remove duplicate imported file


STRRL:
Hi @soniasingla, as you deleted the `cm`, you should rename the references with `cm` to `common`.

pkg/chaosctl/cmd/debug.go
```diff
@@ -26,7 +26,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/chaos-mesh/chaos-mesh/pkg/chaosctl/common"
```
soniasingla, line 28: In line 28, it has been declared twice, so removing `cm` should not impact, can you explain me little bit? :)

STRRL: for example:

https://github.com/chaos-mesh/chaos-mesh/blob/7e12dea08017445d11f95e5540df5a34d4916ea1/pkg/chaosctl/cmd/debug.go#L76

should be 


```
	clientset, err := common.InitClientSet()
```

STRRL (changes requested):
please rename the import alias `cm` to `common`

soniasingla:
Thanks @STRRL for the review, i have updated the references :)

AsterNighT (approved):
LGTM

STRRL (approved):
LGTM

PR #1774, opened 2021-05-04
Weak PRNG fixes

arcolife:

Although `math/rand` is used to generate non-essential RNGs for stuff like Service account names and random pod/namespace/index selection, but as a best practice `crypto/rand` is a [cryptographically-secured PRNG](https://en.wikipedia.org/wiki/Cryptographically-secure_pseudorandom_number_generator)

We were using `math/rand` instead of `crypto/rand`


arcolife:
cc/ @fewdan @Yiyiyimu 

please note I've introduced an additional function to minimize usage differences for RNG across the 2 files.

pkg/selector/selector.go
```diff
@@ -528,6 +529,11 @@ func IsAllowedNamespaces(ctx context.Context, c client.Client, namespace string)
 	return false, nil
 }
 
+func getRandomNumber(length int) int64 {
```
AsterNighT, line 532: `max` or `upperbound` seem to be a better name than `length`. And since you never get a negative value from `rand`, what about using `uint64` instead of `int64`?

arcolife: sure makes sense

PR #1777, opened 2021-05-04
shell checks: double quote expansion fixes

arcolife:
Signed-off-by: arcolife <archit.py@gmail.com>

Double quote array expansions to avoid re-splitting elements.


arcolife:
cc/ @Colstuwjx @YangKeao

PR #1798, opened 2021-05-06
Update init.go and selector.go by adding better imports

pankaj892:
To escape HTML content it would not be possible by using text import so changed it to html import to escape HTML content
In selector.go removed weak random number generator by adding better import
Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

<!-- 
If no, just leave the release note block below as is.

If yes, a release note is required:
Enter your extended release note in the block below. If the PR requires additional action from users switching to the new rele

### DIFF
--- a/cmd/chaos-builder/init.go
+++ b/cmd/chaos-builder/init.go
@@ -15,7 +15,7 @@ package main
 
 import (
 	"bytes"
-	"text/template"
+	"html/template"
 )
 
 const initTemplate = `
--- a/pkg/chaoskernel/pb/bpfki.pb.go
+++ b/pkg/chaoskernel/pb/bpfki.pb.go
@@ -393,7 +393,7 @@ func init() {
 	proto.RegisterType((*StatusResponse)(nil), "bpfki.StatusResponse")
 }
 
-func init() { proto.RegisterFile("bpfki.proto", fileDescriptor_62eed357eb71de0e) }
+func init() { protoregistry.GlobalFiles.RegisterFile("bpfki.proto", fileDescriptor_62eed357eb71de0e) }
 
 var fileDescriptor_62eed357eb71de0e = []byte{
 	// 518 bytes of a gzipped FileDescriptorProto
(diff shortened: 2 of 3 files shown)


cmd/chaos-builder/init.go
```diff
@@ -15,7 +15,7 @@ package main
 
 import (
 	"bytes"
-	"text/template"
+	"html/template"
```
STRRL, line 18: `chaos-builder` is a tool for code generation, we do not render HTML content.

STRRL (commented):
and the modification about `rand` is duplicated with #1774, if that PR be merged sooner, this PR will be closed.

cmd/chaos-builder/init.go
pankaj892: @STRRL so i need to revert the import in init.go and keep the file as it is or do you want me to make any more changes ?

pkg/chaoskernel/pb/bpfki.pb.go
```diff
@@ -393,7 +393,7 @@ func init() {
 	proto.RegisterType((*StatusResponse)(nil), "bpfki.StatusResponse")
 }
 
-func init() { proto.RegisterFile("bpfki.proto", fileDescriptor_62eed357eb71de0e) }
+func init() { protoregistry.GlobalFiles.RegisterFile("bpfki.proto", fileDescriptor_62eed357eb71de0e) }
```
YangKeao, line 396: Code generated by protoc-gen-go.  **DO NOT EDIT.**. The correct way to fix `deprecated` in this file may be updating the protoc / protobuf / grpc and regenerate this file. However, it could be a little complicated.

YangKeao (changes requested):

cmd/chaos-builder/init.go
STRRL: > so i need to revert the import in init.go and keep the file as it is

Well that file should be updated by code generation tool: protoc, you should not edit it manually.

> do you want me to make any more changes?

If you're interested in protobuf and protoc, you could take a look on `Makefile:375`, add bpfki.pb.go in that recipe.

STRRL:

PR #1801, opened 2021-05-07
PodNetworkChaos through reconciler

YangKeao:

Reconcile `podnetworkchaos` through reconciler.

1. [x] Add a `podnetworkchaos` controller to listen to the resource
2. [x] Add a customized status for `networkchaos` to record whether it has been modified and the generation of it.
3. [x] Reconcile and migrate to success status when the generation has reached.


controllers/common/controller.go
```diff
@@ -109,16 +120,36 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+
+			if strings.HasPrefix(string(originalPhase), string(v1alpha1.NotInjected)) {
+				operation = Apply
+			} else {
+				operation = Recover
+			}
+		}
 
+		if operation == Apply {
+			operation = Apply
```
Hexilee, line 152: Shall we delete this line?

controllers/podnetworkchaos/controller.go
```diff
@@ -29,76 +31,134 @@ import (
...
 			Cidrs: ipset.Cidrs,
 		})
 	}
-	return ipset.FlushIPSets(ctx, h.Client, pod, ipsets)
+	return ipset.FlushIPSets(ctx, r.Client, pod, ipsets)
 }
 
 // SetIptables sets iptables on pod
-func (h *Handler) SetIptables(ctx context.Context, pod *corev1.Pod, chaos *v1alpha1.PodNetworkChaos) error {
+func (r *Reconciler) SetIptables(ctx context.Context, pod *corev1.Pod, chaos *v1alpha1.PodNetworkChaos) error {
```
Hexilee, line 161: It seems we can manage the iptables rules by PodNetworkChaos instead of PodHttpChaos itself. ( However, we need more features here in that case.

controllers/utils/controller/ownerReferences.go
```diff
@@ -0,0 +1,80 @@
...
+		existingRefs[fi] = ref
+	}
+
+	// Update owner references
+	object.SetOwnerReferences(existingRefs)
+	return nil
+}
+
+// Returns true if a and b point to the same object
+func referSameObject(a, b metav1.OwnerReference) bool {
```
Hexilee, line 68: Why not just compare by the UID?

Hexilee (commented):
Rest LGTM

controllers/utils/controller/ownerReferences.go
YangKeao: Fixed

controllers/common/controller.go
YangKeao: Fixed

controllers/podnetworkchaos/controller.go
YangKeao: Yes. This structure design is very limited. Have you tested, whether `podhttpchaos` works well with `podnetworkchaos` (except the double delay problem)?

Hexilee (approved):
LGTM

controllers/podnetworkchaos/controller.go
Hexilee: Not yet, I will test it in implementing the `httpchaos`.

PR #1815, opened 2021-05-07
Revert "shell checks: double quote expansion fixes"

YangKeao:
Reverts chaos-mesh/chaos-mesh#1777

As it seems to have broke the e2e test.

It's amazing that the bot seems to merge this PR without running e2e test.


STRRL:
As #1816 has been merged, I think we could close this pr.

PR #1819, opened 2021-05-09
fix storage class var name in helm

saravanan30erd:
In helm, persistent volume claim for chaos mesh dashboard is not creating due to the issue in PVC storage class variable name.

Used the correct storage class variable name in PVC template.


cwen0 (commented):
LGTM

cwen0 (approved):

fewdan (approved):
LGTM

PR #1827, opened 2021-05-12
feat: base implementation of parallel node

STRRL:

The implementation of parallel node;

It's extremely similar to the serial node, with several small difference:
- parallel node will spawn **all** of the task at the same time
- when one of the Spec.Task changes, all of the children should be deleted and respawn

### Checklist

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


STRRL:
I think this PR is ready to review

/cc @YangKeao 
/cc @WangXiangUSTC 
/cc @AsterNighT 
/cc @Hexilee 
/cc @fewdan

pkg/workflow/controllers/parallel_node_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+
+	// only resolve parallel nodes
+	if node.Spec.Type != v1alpha1.TypeParallel {
+		return reconcile.Result{}, nil
+	}
+
+	it.logger.V(4).Info("resolve parallel node", "node", request)
+
+	// make effects, create/remove children nodes
+	err = it.syncChildrenNodes(ctx, node)
```
AsterNighT, line 76: umm shouldn't it be `ChildNodes` rather than `ChildrenNodes`?

pkg/workflow/controllers/parallel_node_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+func getTaskNameFromGeneratedName(generatedNodeName string) string {
+	index := strings.LastIndex(generatedNodeName, "-")
+	if index < 0 {
+		return generatedNodeName
+	}
+	return generatedNodeName[:index]
+}
+
+// relativeComplementSet return the set of elements which contained in former but not in latter
+func relativeComplementSet(former []string, latter []string) []string {
```
AsterNighT, line 228: That is call `set difference` not `complement`, if I haven't got it wrong.

pkg/workflow/controllers/parallel_node_reconciler.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	existsChildrenNodes := append(activeChildrenNodes, finishedChildrenNodes...)
+
+	var taskNamesOfNodes []string
+	for _, childNode := range existsChildrenNodes {
+		taskNamesOfNodes = append(taskNamesOfNodes, getTaskNameFromGeneratedName(childNode.GetName()))
+	}
+
+	var tasksToStartup []string
+	if len(relativeComplementSet(taskNamesOfNodes, node.Spec.Tasks)) > 0 {
```
AsterNighT, line 158: What if I only add one task? Would the entire node be restarted?

AsterNighT (commented):
Rest LGTM!

pkg/workflow/controllers/parallel_node_reconciler.go
STRRL: I want to present that: https://en.wikipedia.org/wiki/Complement_(set_theory)#Relative_complement

pkg/workflow/controllers/parallel_node_reconciler.go
STRRL: Yes. I think it might be better if all of the tasks will be spawned at the same time and run **parallelly**.

pkg/workflow/controllers/parallel_node_reconciler.go
STRRL: fixed, PTAL

pkg/workflow/controllers/parallel_node_reconciler.go
AsterNighT: Oh, mb.

AsterNighT: It seems relativeComplementSet indicate B-A while set difference is A-B, I thought you mean A-B?

pkg/workflow/controllers/parallel_node_reconciler.go
AsterNighT: But it seems that if I only add one task, `relativeComplementSet(taskNamesOfNodes, node.Spec.Tasks)` would be empty. Thus it would not be restarted.

pkg/workflow/controllers/parallel_node_reconciler.go
STRRL: Yeah, you are right!

pkg/workflow/controllers/parallel_node_reconciler.go
STRRL: I update that logic, PTAL

pkg/workflow/controllers/parallel_node_reconciler.go
STRRL: fixed, PTAL

STRRL:
conflicts resolved, PTAL @YangKeao @Hexilee @cwen0

cwen0 (approved):
LGTM

PR #1831, opened 2021-05-12
implement pause controller

### LINKED ISSUE #1817: Support `pause` in Schedule
`pause` has been supported by the `common` controller. However, it hasn't been supported by the `cron` controller yet. 

1. Skip the spawn when it was paused
2. Automatically pause the active objects

The second one could be done by a standalone controller. And the first one should be done inside the `cron` controller.

YangKeao:

Fix #1817

1. Skip the spawn when it was paused
2. Automatically pause the active objects

### DIFF
--- a/controllers/schedule/pause/controller.go
+++ b/controllers/schedule/pause/controller.go
@@ -18,9 +18,6 @@ import (
 	"reflect"
 	"strconv"
 
-	"github.com/chaos-mesh/chaos-mesh/api/v1alpha1"
-	"github.com/chaos-mesh/chaos-mesh/controllers/schedule/utils"
-	"github.com/chaos-mesh/chaos-mesh/controllers/types"
 	"github.com/go-logr/logr"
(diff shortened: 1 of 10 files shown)


controllers/schedule/pause/controller.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	list, err := r.ActiveLister.ListActiveJobs(ctx, schedule)
+	if err != nil {
+		r.Recorder.Eventf(schedule, "Warning", "Failed", "Failed to list active jobs: %s", err.Error())
+		return ctrl.Result{}, nil
+	}
+
+	items := reflect.ValueOf(list).Elem().FieldByName("Items")
+	for i := 0; i < items.Len(); i++ {
```
AsterNighT, line 61: Do we need to set pause for finished jobs?

YangKeao: It doesn't matter. We don't need to filter out finished jobs manually :smile_cat:

PR #1832, opened 2021-05-12
add condition controller

### LINKED ISSUE #1698: Use `Condition` to track the status of a chaos
Like a `Pod`, use `Condition` to track the current status of a chaos.

YangKeao:

Fix #1698


controllers/condition/controller.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}
+
+	if !reflect.DeepEqual(newConditionMap, conditionMap) {
+		conditions := make([]v1alpha1.ChaosCondition, 0, 5)
+		for k, v := range newConditionMap {
+			conditions = append(conditions, v1alpha1.ChaosCondition{
+				Type:   k,
+				Status: v.Status,
+				Reason: v.Reason,
```
Colstuwjx, line 117: How to fill out the `reason`?

We did have `FailedMessage` to store the error message while applied failed or recover failed. For now, it seems that we lost this message?

AsterNighT (approved):
LGTM, can you brief me about what these conditions are for?

controllers/condition/controller.go
YangKeao: There is only one reason in these condition, and they are in the `if` statement :yum:.

For example, in the conditions of a `Pod`, the reason is `ContainersNotReady` but not `ImagePullBackOff`. If we need to find a place to write the `FailedMessage`, a `recordStatuses` or adding status for every record is better.

Now, the error messages are only sent through events. I could add a status for every record later :smile_cat: .https://github.com/chaos-mesh/chaos-mesh/issues/1833 an issue has been recorded.

Colstuwjx: IMO, I think `reason` field is optional, rather than required as we defined [here](https://github.com/chaos-mesh/chaos-mesh/pull/1832/files#diff-1b59fa0b31ead22983a731ae3671d88df11dcfb6d0ba7bb84cdaeca7d50f9c4dR92).

It would be empty if the condition status is `true`. If the condition status is `false`, the `reason` should be sth like one of the preset enums, like Pod Conditions [consts](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/status/generate.go#L33).

The reason field should tell the general failed reason, and just like the `Pod` resource did, the details, as you mentioned, would be better in sth like `recordStatuses`.

Another question is why we need a separate condition controller? We did apply action at [here](https://github.com/chaos-mesh/chaos-mesh/blob/nirvana/controllers/common/controller.go#L123), and now we just send events while it fails, and why not update its conditions accordingly. Just like kubelet did [syncPod](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet.go#L1490) and it sends [event](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_container.go#L180) if ImagePullBackOff, [updates conditions](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kubelet.go#L1711) inside this sync loop as well.

controllers/condition/controller.go
```diff
@@ -0,0 +1,140 @@
...
+			})
+		}
+
+		updateError := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
STRRL, line 121: I think we'd better to wrap all the logic from L52 - L119 with `retry.RetryOnConflict`.

AsterNighT: +1

controllers/condition/controller.go
```diff
@@ -0,0 +1,140 @@
+// Copyright 2019 Chaos Mesh Authors.
```
fewdan, line 1: s/2019/2021/g  ?

AsterNighT:
I suppose gc and schedule controller would base on this. Waiting for this pr to merge before #1861

controllers/condition/controller.go
YangKeao: Fixed.

controllers/condition/controller.go
YangKeao: Fixed

STRRL (approved):
LGTM

PR #1836, opened 2021-05-13
PodIOChaos through reconciler

YangKeao:
This PR is based on #1801, and is implementing IOChaos through reconciler.


Hexilee (approved):
LGTM

STRRL (commented):
And each `PodIOChaos` has two `OwnerRefernce`, one is `IOChaos`, another is `Pod` which is injected. I'm not sure about it breaks the expected garbage collection/cascade deletion when Pod is deleted. 

I'm still digging into the package `garbadgecollector` in kubernetes, please tell me more if you already have a conclusion.

controllers/podiochaos/fx.go
```diff
@@ -0,0 +1,53 @@
...
+	err := ctrl.NewControllerManagedBy(mgr).
+		For(&v1alpha1.PodIoChaos{}).
+		Named("podiochaos").
+		WithEventFilter(predicate.Funcs{
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				oldObj := e.ObjectOld.(*v1alpha1.PodIoChaos)
+				newObj := e.ObjectNew.(*v1alpha1.PodIoChaos)
+
+				return !reflect.DeepEqual(oldObj.Spec, newObj.Spec)
+			},
```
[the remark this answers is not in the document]
YangKeao: If we don't set it as a SubResource, updating the `observedGeneration` in the status will also increase the `generation` in metadata, which is really wired (and will cause unnecessary reconcile)

controllers/podiochaos/controller.go
```diff
@@ -0,0 +1,161 @@
...
+		updateError := retry.RetryOnConflict(retry.DefaultBackoff, func() error {
+			obj := &v1alpha1.PodIoChaos{}
+
+			if err := r.Client.Get(context.TODO(), req.NamespacedName, obj); err != nil {
+				r.Log.Error(err, "unable to get chaos")
+				return err
+			}
+
+			obj.Status.FailedMessage = failedMessage
+			obj.Status.ObservedGeneration = observedGeneration
```
Hexilee, line 92: It seems the `failedMessage`, `observedGeneration`, `pid` and `startTime` will never change.

Hexilee (changes requested):

controllers/podiochaos/controller.go
YangKeao: `observedGeneration` is initialized with the newest generation, and `pid`, `startTime` are modified in the following part.

`failedMessage` is modified in line 80.

Hexilee: Yes, I see. But in function RetryOnConflict, the newest generation may change.

Hexilee (approved):
LGTM

YangKeao:
> And each `PodIOChaos` has two `OwnerRefernce`, one is `IOChaos`, another is `Pod` which is injected. I'm not sure about it breaks the expected garbage collection/cascade deletion when Pod is deleted.
> 
> I'm still digging into the package `garbadgecollector` in kubernetes, please tell me more if you already have a conclusion.

@STRRL PTAL again.

controllers/chaosimpl/iochaos/impl.go
```diff
@@ -172,6 +175,12 @@ func (impl *Impl) Recover(ctx context.Context, index int, records []*v1alpha1.Re
 		if err == podiochaosmanager.ErrPodNotFound || err == podiochaosmanager.ErrPodNotRunning {
 			return v1alpha1.NotInjected, nil
 		}
+
+		if k8sError.IsForbidden(err) {
+			if strings.Contains(err.Error(), "because it is being terminated") {
+				return v1alpha1.NotInjected, nil
+			}
+		}
```
STRRL: That confuses me. It seems this `err` is returned by `Commit`, I saw that there are several actions like "Create", "Update", and "Get". But why does the termination could affect the RBAC authorization of PodIoChaos?

api/v1alpha1/iochaos_types.go
```diff
@@ -92,10 +92,18 @@ type IoChaosSpec struct {
 // IoChaosStatus defines the observed state of IoChaos
 type IoChaosStatus struct {
 	ChaosStatus `json:",inline"`
+
+	// Instances always specifies podnetworkchaos generation or empty
```
Hexilee, line 96: Should be podiochaos?

controllers/chaosimpl/iochaos/impl.go
YangKeao: The full error is: "Forbid to create xxx in namespace xxx, because it is being terminated". When the namespace is being deleted, `api-server` will refuse to create new object in that namespace, so that the `podiochaos` cannot be created, which is needed for recovery. However, it can be ignored safely, because when the original `podiochaos` is deleted, these pods have been recovered (if they still exist).

This error is really common in e2e test, and it will block e2e test to delete the working namespace.

api/v1alpha1/iochaos_types.go
YangKeao: Fixed

Issue #1838, opened 2021-05-14
Seeing error.api.no_cluster_privilege: can't list resource in the cluster when "All" is selected in "Choose namespace"


alperkokmen:
## Bug Report

**What version of Kubernetes are you using?**
```
❯ kubectl version --short
Client Version: v1.19.0
Server Version: v1.18.9-eks-d1db3c
```

**What version of Chaos Mesh are you using?**
1.2.0

**What did you do?**
I installed chaos-mesh by running the install script with `--template` option and then making some changes to use `CLUSTER_MODE=false`. Dashboard and controller started up fine; but, then in the dashboard UI, I keep getting failed XHR requests (401) and an error message when "All" is selected from the namespaces dropdown.

<img width="553" alt="image" src="https://user-images.githubusercontent.com/1657578/118206187-d2921000-b416-11eb-8227-e2cf763de0e4.png">

It seems like `All` is always present in that dropdown.
https://github.com/chaos-mesh/chaos-mesh/blob/5f6c2f0ede375ee057d7375532b5e2b853910280/ui/src/components/TopContainer/Namespace.tsx#L48-L52

I was not expecting an error by default when `All` is selected from the dropdown. Alternatively, maybe `All` shouldn't be there when `CLUSTER_SCOPED=false` and there is a `TARGET_NAMESPACE` set. excerpt from my `chaos-dashboard` `env` –

```
                {
                  name: 'CLUSTER_SCOPED',
                  value: 'false',
                },
                {
                  name: 'TARGET_NAMESPACE',
                  value: config.ChaosMesh.TargetNamespace,
                },
```

Here's the response for the XHR request from the web UI –
```json
{
    "message": "error.api.no_cluster_privilege: can't list resource in the cluster",
    "code": "error.api.no_cluster_privilege",
[...]

PR #1854, opened 2021-05-17
dashboard api: adjust namespace with the target namespace when it is empty

WangXiangUSTC:

fix https://github.com/chaos-mesh/chaos-mesh/issues/1838.

while the Chaos Mesh is installed with a target namespace, the namespace should be the target namespace when the user chooses `All`.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


pkg/apiserver/archive/archive.go
```diff
@@ -53,7 +57,9 @@ type StatusResponse struct {
 // Register mounts our HTTP handler on the mux.
 func Register(r *gin.RouterGroup, s *Service) {
 	endpoint := r.Group("/archives")
-	endpoint.Use(utils.AuthRequired)
+	endpoint.Use(func(c *gin.Context) {
+		utils.AuthRequired(c, s.conf.ClusterScoped, s.conf.TargetNamespace)
```
g1eny0ung, line 61: Should we bump the auth middleware to the top of the group? This can let us set the `namespace` query in advance to omit subsequent processing.

For example:

```go
q := c.Request.URL.Query()
q.Add("test", "testtest")
c.Request.URL.RawQuery = q.Encode()

test := c.Query("test") // testtest
```

WangXiangUSTC: I try to update the query in `gin.Context`, but I found that it not work. Then I read the source code of gin, I find that there is a `queryCache` in context, and when we call `context.Query("text")` it is in fact read data from the cache, and I don't find some functions to update this cache 

https://github.com/gin-gonic/gin/blob/ee4de846a894e9049321e809d69f4343f62d2862/context.go#L447

g1eny0ung: 😢 After taking a look at the code, it seems that this cache is not very useful.

I think there should be a `Reset` method to reset the cache. So, we can solve this problem in the future once gin has the feature of cache reset.

g1eny0ung (approved):
LGTM

cwen0 (approved):
LGTM

PR #1863, opened 2021-05-19
meaningful events message in controllers

### LINKED ISSUE #1825: Investigate meaningful event message
Investigate structured-log and make it easier to parse for Dashboard.

YangKeao:

Add a parser for the events

Fix #1825


AsterNighT (approved):
Get reduced to inventing one more wheel :facepalm:

STRRL (commented):
Rest LGTM!

And I think I will just using json string for as the message when build event for Workflow, like:

```json
{
	"event": "WorkflowNodeCreated",
	"reason": "SerialNodeSchedule",
    "message": "serial node create child node",
    "parent": "serial-node-xxx",
    "task": "network-chaos-delay",
    "node": "network-chaos-delay-xxx"
}
```

Is that OK for you? :see_no_evil:  That will reduce many works on writing `Parse()`

pkg/chaosctl/common/common.go
```diff
@@ -24,9 +24,9 @@ import (
 
 	"google.golang.org/grpc"
 
+	"google.golang.org/grpc"
```
Colstuwjx, line 27: duplicate to L25?

controllers/desiredphase/controller.go
```diff
@@ -76,13 +75,15 @@ func (ctx *reconcileContext) GetCreationTimestamp() metav1.Time {
 	return ctx.obj.GetObjectMeta().CreationTimestamp
 }
 
-func (ctx *reconcileContext) CalcDesiredPhase() v1alpha1.DesiredPhase {
+func (ctx *reconcileContext) CalcDesiredPhase() (v1alpha1.DesiredPhase, []recorder.ChaosEvent) {
+	events := []recorder.ChaosEvent{}
+
```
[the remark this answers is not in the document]
YangKeao: If `desiredPhase` doesn't change, it will not emit any event, so that `Reconcile` multiple times will not create tons of events.

STRRL:
Please resolve the conflicts @YangKeao

YangKeao:
> Rest LGTM!
> 
> And I think I will just using json string for as the message when build event for Workflow, like:
> 
> ```json
> {
> 	"event": "WorkflowNodeCreated",
> 	"reason": "SerialNodeSchedule",
>     "message": "serial node create child node",
>     "parent": "serial-node-xxx",
>     "task": "network-chaos-delay",
>     "node": "network-chaos-delay-xxx"
> }
> ```
> 
> Is that OK for you? see_no_evil That will reduce many works on writing `Parse()`

No. Writing a JSON into event message is too ugly. It's not a huge burden to writing `Parse()`

pkg/chaosctl/common/common.go
YangKeao: Fixed

controllers/utils/recorder/recorder.go
```diff
@@ -0,0 +1,58 @@
...
+func (r *ChaosRecorder) Event(object runtime.Object, ev ChaosEvent) {
+	r.recorder.Event(object, ev.Type(), ev.Reason(), ev.Message())
+}
+
+type ChaosEvent interface {
+	Type() string
+	Reason() string
+	Message() string
+
+	Parse(message string) ChaosEvent
```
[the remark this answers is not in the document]
YangKeao: I didn't see any necessity to add an `error` in this situation. Comments have been added to show when it will return a `nil`.

STRRL:
> Writing a JSON into event message is too ugly

I mean, the whole message is a JSON. 

I do not think that is ugly, that pattern is widely used in structured logs.

https://cloud.google.com/logging/docs/structured-logging

zap also use JSON encoder in ProductionConfig

https://github.com/uber-go/zap/blob/374825111f7fc2fca60d36daeb8d656c3bc5d40a/config.go#L120

(Please not be so nervous about that embedded structured message. 😉)

YangKeao:
> > Writing a JSON into event message is too ugly
> 
> I mean, the whole message is a JSON.
> 
> I do not think that is ugly, that pattern is widely used in structured logs.
> 
> https://cloud.google.com/logging/docs/structured-logging
> 
> zap also use JSON encoder in ProductionConfig
> 
> https://github.com/uber-go/zap/blob/374825111f7fc2fca60d36daeb8d656c3bc5d40a/config.go#L120
> 
> (Please not be so nervous about that embedded structured message. wink)

The total `message` will be a tagged union (multiple different possible types), which is also hard to ser/des in golang grammar. What's more, it would be hard for the users to understand the event message (without seeing the document).

Colstuwjx:
@STRRL I think it's a different scenario for the examples you mentioned, both google structured logging and zap are designed for logging, but now we're sending events.

Events should be a more restricted structure with the logs, since our users may want to collect the events and do an analysis or monitoring job.

If the whole event message is a raw JSON, we'd even not sure about whether it contains `reason` or `message` such basic fields.

How about supporting the `eventf` back? Arbitrary messages could append to the `args`, just like k8s [Eventf](https://github.com/kubernetes/client-go/blob/master/tools/record/event.go#L103) did.

Colstuwjx:
@YangKeao Yes, I agree with you, but on the other side, besides the common fields, each controller may have its own fields, if we don't think `eventf` is a good idea, how about taking the `AnnotatedEventf ` method? Workflow node\parent\tasks could append into the annotations.

STRRL:
> The total `message` will be a tagged union (multiple different possible types), which is also hard to ser/des in golang grammar. 

Yeah, the JSON contains `"event": "WorkflowNodeCreated"` as the name of the struct, I think I would deserialize that as `map[string]interface{}` then compare the name, then rebuild the struct with mapstructure(with reflect). (It is so called "outer tag"? )

> What's more, it would be hard for the users to understand the event message (without seeing the document).

Yes, but I think it is not a big deal.


Actually, the reason why I want to use JSON string as message in Event is only lazy, I do not want to write `Parse()` for each struct. 🙈

STRRL:
> Workflow node\parent\tasks could append into the annotations.

Sounds great. How do you think about it @YangKeao?

an Event like
```golang
type SomeEvent struct {
	Reason string
	Message string
	FieldA string
	FieldB string
	FieldC int 
}
```
could be sterilized to
```yaml
apiVersion: v1
kind: Event
message: <the-message>
metadata:
  annotations:
    chaos-mesh.org/field/field-a: <value>
    chaos-mesh.org/field/field-b: <value>
    chaos-mesh.org/field/field-c: <value>
type: Normal
```

YangKeao:
> @YangKeao Yes, I agree with you, but on the other side, besides the common fields, each controller may have its own fields, if we don't think `eventf` is a good idea, how about taking the `AnnotatedEventf ` method? Workflow node\parent\tasks could append into the annotations.

:+1: Do you mean writing the `struct` into the annotations and writing the human readable message into the string format part?

For example, for the `Applied` event:

```go
type Applied struct {
    	Id string
}
```

It will convert the `struct` into `{id: "test"}`, and write `.AnnotatedEventf(obj, {id: "test"}, "Normal", "Applied", "Successfully apply chaos for %s", "test")`.

I think it's really good. It saves our effort on writing multiple `Parse` functions. @STRRL What's your opinion?

YangKeao:
> > Workflow node\parent\tasks could append into the annotations.
> 
> Sounds great. How do you think about it @YangKeao?
> 
> an Event like
> 
> ```go
> type SomeEvent struct {
> 	Reason string
> 	Message string
> 	FieldA string
> 	FieldB string
> 	FieldC int 
> }
> ```
> 
> could be sterilized to
> 
> ```yaml
> apiVersion: v1
> kind: Event
> message: <the-message>
> reason: <the-reason>
> metadata:
>   annotations:
>     chaos-mesh.org/field/field-a: <value>
>     chaos-mesh.org/field/field-b: <value>
>     chaos-mesh.org/field/field-c: <value>
> type: Normal
> ```

@Colstuwjx @STRRL 

I will have a try, really thanks.

Colstuwjx:
> 👍 Do you mean writing the struct into the annotations and writing the human readable message into the string format part?

Yes! I think the k8s may also meet this issue, and it did have support [AnnotatedEventf](https://github.com/kubernetes/client-go/blob/master/tools/record/event.go#L106).

STRRL:
> I think it's really good. It saves our effort on writing multiple Parse functions. @STRRL What's your opinion?

THATS GREAT!!!

My only concerned thing is please using an identical prefix for the annotation, like whatever names like `chaos-mesh.org/field`

YangKeao:
@STRRL  @Colstuwjx @fewdan  PTAL

controllers/utils/recorder/annotations.go
```diff
@@ -0,0 +1,114 @@
...
+import (
+	"encoding"
+	"errors"
+	"reflect"
+	"strconv"
+
+	"github.com/iancoleman/strcase"
+)
+
+var InvalidType = errors.New("invalid type of fields")
```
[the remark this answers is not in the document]
YangKeao: Fixed

controllers/utils/recorder/annotations.go
```diff
@@ -0,0 +1,114 @@
...
+var UknownType = errors.New("uknown type of fields")
```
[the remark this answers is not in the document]
YangKeao: Fixed

controllers/utils/recorder/common.go
```diff
@@ -0,0 +1,77 @@
...
+
+func (a Applied) Reason() string {
+	return "Applied"
+}
+
+func (a Applied) Message() string {
+	return fmt.Sprintf("Successfully apply chaos for %s", a.Id)
+}
+
+func (a Applied) Parse(message string) ChaosEvent {
```
Colstuwjx, line 37: I think we don't need these `Parse` anymore, right?

YangKeao: Yes. It seems that I left some `Parse`... They are removed now :beers:

controllers/utils/recorder/recorder_test.go
```diff
@@ -0,0 +1,102 @@
...
+
+		ev, err := FromAnnotations(c.annotations)
+		if err != nil {
+			fmt.Printf("fail")
+		}
+		g.Expect(ev).To(Equal(c.ev))
+	}
+}
+
+func TestParse(t *testing.T) {
```
Colstuwjx, line 68: So we don't need this as well, I think.

fewdan (approved):
LGTM

PR #1864, opened 2021-05-19
Move endpoints RBAC rules to the `control-plane` role

YangKeao:

Fix #1857 

The `endpoints` RBAC rules are moved to `control-plane` role, so that in namespace scoped mode, the controller-manager is still able to list endpoints in the `control-plane` namespace (but not target namespace).


fewdan (approved):
LGTM

YangKeao:
I will work on this issue in a new PR, as the nirvana has merged.

PR #1896, opened 2021-05-27
upgrade to apiextensions.k8s.io/v1 CRD

### LINKED ISSUE #1890: keeping CRD with both  apiextensions.k8s.io/v1beta1 and apiextensions.k8s.io/v1
## Feature Request

Yes. As we decided to support kubernetes from v1.15 to the latest (v1.21), and occurs some issue about the CRD.

- We could NOT use `apiextensions.k8s.io/v1` on kubernetes lower than v1.15(include), because it was introduced in v1.16.
- We could NOT use `apiextension.k8s.io/v1beta1` on kubernetes higher than v1.18(include), because of the new validation rules, related issue/PR: https://github.com/kubernetes-sigs/controller-tools/issues/444, https://github.com/kubernetes/kubernetes/pull/88076

So I think we should:

- only using `apiextension.k8s.io/v1beta1` for kubernetes v1.15,(and we need to append `--validate=false` while apply CRD, like `kubectl apply -f --validate=false crd.yaml`; but no need to change anything with installing by helm, still do not know why )
- using `apiextension.k8s.io/v1` for higher version
- ~using helm built-in object `Capabilities.KubeVersion` for switching~ the CRD part is not templated, users should manually import `apiextensions.k8s.io/v1beta1` with kubernetes v1.15

No other ideas for now.

It will not affect the usage of installation with helm on kubernetes v1.16 +.

It asks users for manual installation with `crd-v1beta1.yaml`

STRRL:

fix #1890

- use apiextensions.k8s.io/v1 as the default CRD version
- also generate `manifests/crd-v1beta1.yaml` with `apiextensions.k8s.io/v1beta1`
- select correct version within `install.sh`
- update CI, upload the artifact: `manifests/crd-v1beta1.yaml`
- bump the version of controller-runtime to @0.5.0, because controller-run@0.4.0 still depends kubernetes lower than 1.16, so we could not pass the unit test with controller-runtime@0.4.0
- bump the version of kubebuilder from 2.2.0 to 2.3.2 with the same former reason

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


TomMD:
You might want to add the `e2e-test` directory to the ignore list by changing `.muse/config.toml` to have:

```
ignoreFiles = """
e2e-test/**
"""
```

This would prevent the reporting of bugs such as the CWE from of that directory.

STRRL:
> You might want to add the `e2e-test` directory to the ignore list by changing `.muse/config.toml` to have:
> 
> ```
> ignoreFiles = """
> e2e-test/**
> """
> ```
> 
> This would prevent the reporting of bugs such as the CVE from of that directory.

Thanks! I am going to configure that!

cwen0 (approved):
LGTM

YangKeao (approved):
LGTM

Issue #1884, opened 2021-05-24
Network chaos unable to heal if previous container crashes


jaysunmah:
## Bug Report

**What version of Kubernetes are you using?**

```
$kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-16T00:04:31Z", GoVersion:"go1.14.4", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.14-airbnb0", GitCommit:"9931e7ec32e27c7ef21b3a734d8554a5ebadf00c", GitTreeState:"clean", BuildDate:"2020-12-02T19:00:47Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

**What version of Chaos Mesh are you using?**

(We're using a docker image that was built and stored within my company's internal docker registry)
```
$kc exec -it chaos-controller-manager-754c4c56db-qkvvk -- /usr/local/bin/chaos-controller-manager -version
Controller manager Version: version.Info{GitVersion:"v0.9.0-156+a7b83630d43439", GitCommit:"a7b83630d43439b43c091842b10ed6ed6166f69b", BuildDate:"2021-01-12T19:01:35Z", GoVersion:"go1.14.6", Compiler:"gc", Platform:"linux/amd64"}
```
**What did you do?**

We're testing out the `NetworkChaos` resource on one of our pods. For simplicity, this pod has two containers, A and B. Container A is set to auto crash itself in the event of a network partition. When we inject this network partition, we can verify that A will begin crashing. As a result, when we try to heal this network partition, it appears from the chaos-daemon logs that since the previous container process no longer exists, it is unable to revert the iptable changes in both containers A and B. 

**What did you expect to see?**
I believe our desired behavior is so that the chaos-daemon is able to heal this network partition even if the original container crashes. We should not have to manually delete this pod to clear the iptable changes.

**What did you see instead?**
The network chaos is unable to be healed, and as a result the pod is corrupted. In order to "heal" this pod's network changes, we had to delete the pod and restart it.

**Output of chaosctl**

Output logs from chaos-daemon. As you can see, eventually the proc pid gets reset to 0, which I assume is because the previous process pid is no longer valid
```
[...]

PR #1899, opened 2021-05-27
[podnetworkchaos] Fix networkchaos heal

jaysunmah:
Signed-off-by: Jason Ma <jason.ma@airbnb.com>

Attempting to fix this issue: https://github.com/chaos-mesh/chaos-mesh/issues/1884, where reconciling a networkchaos won't work when the **FIRST** container in the pod is crashed. 

The previous behavior was to simply take the first container in a pod, find its pid, and modify the iptable and ipsets via that process pid. Our current behavior will loop through each container, try to modify the network changes, and return on the first success.

### Checklist

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


controllers/podnetworkchaos/tc/tc.go
```diff
@@ -22,9 +22,14 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	ctrl "sigs.k8s.io/controller-runtime"
+
+	"github.com/chaos-mesh/chaos-mesh/controllers/utils"
```
YangKeao, line 27: Please remove this line to pass the test 😸

jaysunmah:
@YangKeao seems like this integration test is consistently failing due to:
```
Error from server (spec.scheduler: Invalid value: "null": schedule and duration should be omitted or defined at the same time): error when creating "./delay_chaos.yaml": admission webhook "vnetworkchaos.kb.io" denied the request: spec.scheduler: Invalid value: "null": schedule and duration should be omitted or defined at the same time
```
Despite it being a network chaos related test failure, I'm not sure if it's related to my changes 🤔

YangKeao:
> @YangKeao seems like this integration test is consistently failing due to:
> 
> ```
> Error from server (spec.scheduler: Invalid value: "null": schedule and duration should be omitted or defined at the same time): error when creating "./delay_chaos.yaml": admission webhook "vnetworkchaos.kb.io" denied the request: spec.scheduler: Invalid value: "null": schedule and duration should be omitted or defined at the same time
> ```
> 
> Despite it being a network chaos related test failure, I'm not sure if it's related to my changes 🤔

Don't mind. Script tests are expected to fail now. It has passed all tests and I will merge it asap

PR #1904, opened 2021-05-31
chore: remove unused codes

STRRL:

These files were designed for workflow without the kubernetes operator.

As we have implement workflow with kubernetes, and I think we will not use it in the short future. This PR will remove these codes.

They are totally unused and have already decoupled with Kubernetes controllers, so it's safe to delete directly.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

Related changes

```release-note
NONE
```


STRRL:
I'm going to raise another PR for more refactors.

PR #1910, opened 2021-06-01
Controller test unify

AsterNighT:
I could not come up with a very good idea upon making the concurrent granularity to per-spec. Starting multiple controller will cause unexpected behavior.


STRRL:

PR #1919, opened 2021-06-02
Added Controller for awsssmchaosrunner

Debanitrkl:
This is a controller for the integration of awsssmchaosrunner with chaos mesh according to this [rfc.](https://github.com/chaos-mesh/rfcs/pull/16)

Added a file under `controllers/awschaos` called `attack-configuration`

Tests

- [x] Unit test
- [ ] E2E test

Related changes

```release-note
Yes a new feature under `awschaos` is added
```


controllers/awschaos/attack-configuration/types.go
```diff
@@ -0,0 +1,82 @@
...
+			Containers: []v1.Container{
+				{
+					Name: "attackConfiguration",
+					Image: "",
+					Args: { },
+				},
+			},
+		},
+	}
+	e.Log.Info("Running the Container")
```
STRRL, line 63: After creating the Pod, you should update the status of AwsChaos. It means you should also create new fields to describe the status of that created Pod.

And that status will be also used when `Recover()`.

Debanitrkl: Yeah working on that

STRRL:

Debanitrkl:
Can we put this in WIP I'm on a break and I will get back to this as soon as possible. 
Sorry for the delay got caught up with lot of work in last two months.

PR #1928, opened 2021-06-03
remove remote build by moving file into a docker image

YangKeao:

Originally, for compiling on remote docker, Chaos Mesh will build a docker image with files copied from current file. However, this image is recursive, which means history ones cannot be deleted and will reach storage threshold or layer count threshold.

This feature was implemented because Mac OS and minikube uses virtual machine to run docker on a remote machine, so `--volume` would fail to mount the current work directory. However, @STRRL  told me the Mac OS docker virtual machine mounts some folder into the virtual machine, and the minikube also provides some technology to mount directory (with `minikube mount` or start with `start --mount --mount-string "$(pwd):$(pwd)"` on docker driver).

So, it's not needed to load folders into the virtual machine now :smile_cat:. Everything becomes simpler. 

It would be really grateful if any developer using mac os could test for me.


STRRL (approved):
LGTM

and more details about "File Sharing" docker are here if you have interests. https://docs.docker.com/docker-for-mac/#file-sharing

cwen0:
> LGTM
> 
> and more details about "File Sharing" docker are here if you have interests. https://docs.docker.com/docker-for-mac/#file-sharing


Is there a requirement for Docker version for `File Sharing`?

STRRL:
> > LGTM
> > and more details about "File Sharing" docker are here if you have interests. [docs.docker.com/docker-for-mac/#file-sharing](https://docs.docker.com/docker-for-mac/#file-sharing)
> 
> Is there a requirement for Docker version for `File Sharing`?

That feature has existed since the very old version in 2016.

cwen0 (approved):
LGTM

PR #1932, opened 2021-06-03
fix controller plane cache problem

### LINKED ISSUE #1847: chaos injection fails with: Endpoints "chaos-daemon" not found when TARGET_NAMESPACE is set (CLUSTER_SCOPED=false)
## Bug Report

```
❯ kubectl version --short
Client Version: v1.19.0
Server Version: v1.18.9-eks-d1db3c
```

1.2.0

**What did you do?**
I installed chaos-mesh by running the install script with `--template` option and then making some changes to use `CLUSTER_MODE=false`. After applying all the manifests, I confirmed that all pods were running and dashboard was available.

```
❯ kubectl -n chaos-testing get pods
NAME                                        READY   STATUS    RESTARTS   AGE
chaos-controller-manager-68d6677dd5-2n4g9   1/1     Running   0          100m
chaos-daemon-b7d9z                          1/1     Running   0          77m
chaos-dashboard-6f694ccdb6-lrdmt            1/1     Running   0          100m
```

chaos-mesh pods are scheduled in `chaos-testing` namespace. This is essentially the value for `namespace`-scoped resources in the manifests generated from `install.sh`. Target namespace is `dev` which is set for `chaos-dashboard` and `chaos-mesh` –

```
# chaos-dashboard container env
{
  name: 'CLUSTER_SCOPED',
  value: 'false',
},
{
  name: 'TARGET_NAMESPACE',
  value: 'dev',
},

# chaos-mesh container env
{
  name: 'TARGET_NAMESPACE',
  value: 'dev',
},
{
  name: 'CLUSTER_SCOPED',
  value: 'false',
},
```

When I try to schedule a `PodChaos`, I get the following error which is complaining about `chaos-daemon` not being found.

```
2021-05-14T23:18:00.025Z	DEBUG	controller-runtime.manager.events	Warning	{"object": {"kind":"PodChaos","namespace":"dev","name":"kill-container-3","uid":"8b6681bf-5d01-4d06-bab3-cdb58ff0a0bb","apiVersion":"chaos-mesh.org/v1alpha1","resourceVersion":"9887552512"}, "reason": "ChaosInjectFailed", "message": "Endpoints \"chaos-daemon\" not found"}
```

I have verified that `chaos-daemon` endpoint actually exists in `chaos-testing` namespace where the associated service is created.

```
❯ kubectl -n chaos-testing get endpoints/chaos-daemon
NAME           ENDPOINTS                                   AGE
chaos-daemon   100.92.138.177:31766,100.92.138.177:31767   7h25m
```

It seems `chaos-mesh` container (as part of `chaos-controller-manager` service account) is trying to find the `chaos-daemon` endpoint in `chaos-testing` namespace and failing to do so.

- `chaos-controller-manager-cluster-level` only allows `get`, `list`, `watch` on `nodes` ([src](https://github.com/chaos-mesh/chaos-mesh/blob/5f6c2f0ede375ee057d7375532b5e2b853910280/helm/chaos-mesh/templates/controller-manager-rbac.yaml#L43-L62)).
- `chaos-controller-manager-target-namespace` is only a `RoleBinding` for the target namespace so I don't think it helps here ([src](https://github.com/chaos-mesh/chaos-mesh/blob/5f6c2f0ede375ee057d7375532b5e2b853910280/helm/chaos-mesh/templates/controller-manager-rbac.yaml#L13-L40)).
- `chaos-controller-manager-control-plane` is only for `configmaps` and `services` ([src](https://github.com/chaos-mesh/chaos-mesh/blob/5f6c2f0ede375ee057d7375532b5e2b853910280/helm/chaos-mesh/templates/controller-manager-rbac.yaml#L66-L81)).

Things work fine when `CLUSTER_SCOPED=true` and `chaos-controller-manager-target-namespace` becomes a `ClusterRoleBinding`. Is it possible that `chaos-controller-manager-control-plane` needs to include `endpoints` as well?

YangKeao:

Fix #1847 

Add a partial cached reader to get the endpoint. If failed to create the cached reader, it will fallback to a no cache reader.


api/webhook/validate_auth.go
```diff
@@ -52,7 +52,6 @@ var authLog = ctrl.Log.WithName("validate-auth")
 type AuthValidator struct {
 	enabled bool
 	client  client.Client
-	reader  client.Reader
```
STRRL, line 55: both `reader` and `client` are not used in `AuthValidator`, it seems we could remove them

controllers/chaosimpl/fx.go
```diff
@@ -38,4 +39,6 @@ var AllImpl = fx.Options(
 	podchaos.Module,
 	stresschaos.Module,
 	jvmchaos.Module,
-	timechaos.Module)
+	timechaos.Module,
+
+	utils.Module)
```
STRRL, line 44:
```suggestion
	timechaos.Module,
	utils.Module,
)
```

STRRL (commented):
I notice that there are some dedicated constructors of struct **only** for fx, it works well now, but I still prefer to keep the origin simple constructor and write another function for fx usage.

rest LGTM

controllers/chaosimpl/fx.go
YangKeao: Fixed.

api/webhook/validate_auth.go
YangKeao: Fixed.

YangKeao:
> 
> 
> I notice that there are some dedicated constructors of struct **only** for fx, it works well now, but I still prefer to keep the origin simple constructor and write another function for fx usage.

In most cases, they shouldn't be constructed outside `fx`. So the simple constructor will always be a `unused function` 😢

STRRL (approved):
LGTM!

AsterNighT (approved):

PR #1954, opened 2021-06-09
set exclusive minimum for startingDeadlineSeconds

YangKeao:
Set `startingDeadlineSeconds` as `exclusiveMinimum` to avoid `0` value, which could be confusing and be misunderstood as a replacement of null.


fewdan (approved):
LGTM

cwen0 (approved):
LGTM

PR #1970, opened 2021-06-15
workflow: propagate deadline into children node

STRRL:

the `deadline` condition of serial, parallel, and task node should be propagated into their children node, so the deadline could effect chaos immediately.

What's Changed:

### Related changes

* PR to update `chaos-mesh/website`/`chaos-mesh/website-zh`: No.
* Need to update Chaos Dashboard component, related issue: No.
* Need to cheery-pick to the release branch: No.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```release-note
Please add a release note.
```


pkg/workflow/controllers/deadline_reconciler.go
```diff
@@ -15,8 +15,11 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"time"
 
+	"k8s.io/apimachinery/pkg/types"
```
fewdan, line 21: It seems that it should be "import grouped"?

STRRL: fixed PTAL

cwen0 (approved):
LGTM

YangKeao (approved):
LGTM

PR #2002, opened 2021-06-21
chaosRecorder: submit events without kubernetes event recorder

YangKeao:

No events will be aggregated now.

Submit every events manually, without the handling of kubernetes client event recorder.


STRRL (approved):
LGTM!

cwen0 (approved):
LGTM

PR #2027, opened 2021-06-24
setup another cache which would be updated if "Update" function is called

YangKeao:

There are some mechanism to avoid outdated fields in chaos resource. However, the following codes may actually timeout:

```go
// TODO: make the interval and total time configurable
// The following codes ensure the Schedule in cache has the latest lastScheduleTime
ensureLatestError := wait.Poll(100*time.Millisecond, 2*time.Second, func() (bool, error) {
	schedule = schedule.DeepCopyObject().(*v1alpha1.Schedule)

	if err := r.Client.Get(ctx, req.NamespacedName, schedule); err != nil {
		r.Log.Error(err, "unable to get schedule")
		return false, err
	}

	return schedule.Status.LastScheduleTime.Time == lastScheduleTime, nil
})
```

I developed a new client, which includes another cache. So that we could store the newest resource in the cache temporarily. And it's really hard to decide when it could be removed from the cache, so an LRU is used. 

It cannot 100% ensure the data is newest, because:

1. The resource may have been removed from the LRU
2. The user has modified it manually, and it hasn't been synced with the cache inside controller-runtime's client.

### DIFF
--- a/cmd/chaos-controller-manager/provider/controller.go
+++ b/cmd/chaos-controller-manager/provider/controller.go
@@ -17,6 +17,7 @@ import (
 	"math"
 
 	"github.com/go-logr/logr"
+	lru "github.com/hashicorp/golang-lru"
 	"go.uber.org/fx"
 
 	"github.com/chaos-mesh/chaos-mesh/api/v1alpha1"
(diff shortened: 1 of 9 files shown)


STRRL (approved):
LGTM!

STRRL:
does this PR need to be released in `v2.0.0-beta.2` ？ or next beta version? @YangKeao

YangKeao:
> does this PR need to be released in `v2.0.0-beta.2` ？ or next beta version? @YangKeao

I think it's not urgent. It's enough to merge when it's approved :smiley_cat:

cwen0 (approved):
LGTM

PR #2034, opened 2021-06-25
Run default before validate schedule chaos

YangKeao:

Some validations depend on the mutating webhook. For example, network chaos validating webhook will validate the `correlation`, which is allowed to be empty and will be set by mutating webhook (to zero)


STRRL (approved):
LGTM!

cwen0 (approved):
LGTM

STRRL:
/merge

e2e test failed because of unstable network environment

PR #2066, opened 2021-07-01
chore(api): use direct routes

### LINKED ISSUE #1500: Dashboard takes into account base URL for redirects
## Feature Request

The chaos-mesh dashboard is hardcoded to the URL path `/dashboard` and this prevents in some cases from exposing the dashboard through Ingress. i.e. because there is already something deployed in that path, or certain paths are not allowed. 
For example if you configure chaos-mesh with the following configuration:

```
dashboard:
  create: true
  securityMode: true
  service:
    type: ClusterIP
  ingress:
    enabled: true
    hosts:
      - name: localhost.localdomain
        paths: ["/admin/chaos-mesh"]
```

When you access `https://localhost.localdomain/admin/chaos-mesh` you get redirected to `https://localhost.localdomain/dashboard`. 

The dashboard will be exposed in the path that you set in the Ingress rule. If you take the mentioned configuration, the desired result would be that the dashboard is available in `https://localhost.localdomain/admin/chaos-mesh/dashboard`. This means that it will take into account the base url when doing the redirect to `/dashboard`

In my case I could use port forwarding to access the dashboard, but there are cases where the only available interface will be the GUI, and in those cases port forwarding is not an option.
It has been suggested too that some Ingress Controller have configurations to avoid this problem, although in my Ingress Controller of choice (Traefik) I couldn't get any configuration to prevent this, specially as the UI of Traefik was deployed in `https://localhost.localdomain/dashboard`.

This feature will be beneficial in the cases where kubectl is not available to do port forwarding, or in cases where, for any reason, the path `/dashboard` is not a valid option.

g1eny0ung:

Close #1500.

#### Problem Summary

Since the https://github.com/gin-gonic/gin/pull/2706 be merged. Now gin allows param and exact paths to existing on the same route level.

So all of the UI assets can be safely routed by `/` and `/:foo/*bar`. This is almost the same as the `/*any`.

Except for all descriptions above, this PR also introduces some bug fixes and so on:

- Fix the problem of duplicate `uid` indices in SQLite
- Use `gorm.Model` to replace dummy fields
- Bump API server and UI versions
- Some code formats

### Related changes

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```release-note
Please add a release note.
```

### DIFF
--- a/pkg/apiserver/server.go
+++ b/pkg/apiserver/server.go
@@ -49,7 +49,7 @@ func serverRegister(r *gin.Engine, conf *config.ChaosDashboardConfig) {
 	go r.Run(listenAddr)
 }
 
-func newEngine() *gin.Engine {
+func newEngine(config *config.ChaosDashboardConfig) *gin.Engine {
 	r := gin.Default()
 
 	// default is "/debug/pprof/"
@@ -70,20 +70,25 @@ func newEngine() *gin.Engine {
 		v.RegisterValidation("RequiredFieldEqual", apivalidator.RequiredFieldEqualValid, true)
 	}
 
-	moveToUIRoot := func(c *gin.Context) {
-		c.Redirect(http.StatusMovedPermanently, "/dashboard")
-	}
-
-	r.GET("/", moveToUIRoot)
 	ui := uiserver.AssetsFS()
 	if ui != nil {
-		newDashboardRouter(r, ui)
+		r.GET("/", func(c *gin.Context) {
+			c.FileFromFS("/", ui)
+		})
+		r.GET("/:foo/*bar", func(c *gin.Context) {
+			c.FileFromFS("/", ui)
+		})
+
+		renderStatic := func(c *gin.Context) {
+			c.FileFromFS(c.Request.URL.Path, ui)
+		}
+		r.GET("/static/*any", renderStatic)
+		r.GET("/favicon.ico", renderStatic)
 	} else {
-		r.GET("/dashboard", func(c *gin.Context) {
+		r.GET("/", func(c *gin.Context) {
 			c.String(http.StatusOK, "Dashboard UI is not built. Please run `UI=1 make`.")
 		})
 	}
-	r.NoRoute(moveToUIRoot)
 
 	return r
 }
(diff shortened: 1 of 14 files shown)


fewdan (approved):
LGTM

pkg/apiserver/server.go
```diff
@@ -70,20 +70,25 @@ func newEngine() *gin.Engine {
...
-	r.GET("/", moveToUIRoot)
 	ui := uiserver.AssetsFS()
 	if ui != nil {
-		newDashboardRouter(r, ui)
+		r.GET("/", func(c *gin.Context) {
+			c.FileFromFS("/", ui)
+		})
+		r.GET("/:foo/*bar", func(c *gin.Context) {
+			c.FileFromFS("/", ui)
+		})
```
STRRL, line 80: what does `/:foo/*bar` means? I do not found and reference in gin's doc.

STRRL, line 80: It seems these lines of codes are unnecessary. Only `renderStatic` is the useful func.

STRRL (commented):
rest LGTM

pkg/apiserver/server.go
g1eny0ung: > I do not found and reference in gin's doc.

It still lacks docs.

An example, `/experiments` will match `/:foo` and `experiments/xxx` will match `/:foo/*bar`.

Because gin does not permit to define root route wildcard (`/*any`) and before https://github.com/gin-gonic/gin/pull/2663 (This include in gin `1.7`) it's also not permitted to define mixed param and non-param paths (Like `/:foo ` and `/api`).

So with the latest gin, I can use `/:foo/*bar` to simulate all of the frontend routes. Without the changes above, we can only define the **exact** routes like below:

- `/experiments`
- `/experiments/:uid`
- `/schedules`
- and so on

This would be very redundant.

In the past, we have been limited by the [httprouter](https://github.com/julienschmidt/httprouter), you can found the history reason in https://github.com/julienschmidt/httprouter/pull/329, This pr still not merged. Roughly speaking, httprouter forgoes fuzzy matching for the sake of speed.

g1eny0ung: Not. Because the production build of the UI only has one entry, `index.html`, it's a single page application. So it's needed to redirect all requests to root.

STRRL: Oh, I get that, the `:foo` and `*bar` are only placeholders, not meaningful parameters.
This route will show `index.html` on all the paths (meaningful in frontend), like `/expriments/<uid>`, then let javascript parse/resolve the path for showing certain pages.

Could you append some comments on the `/:foo/*bar`, the name is ambiguous.

g1eny0ung: Got it. I will add some comments later.

STRRL (approved):
LGTM!

STRRL (approved):
> LGTM!

PR #2072, opened 2021-07-02
chore(api): use direct routes (#2066)

### LINKED ISSUE #1500: Dashboard takes into account base URL for redirects

ti-srebot:
cherry-pick #2066 to release-2.0
You can switch your code base to this Pull Request by using [git-extras](https://github.com/tj/git-extras):
```bash
# In chaos-mesh repo:
git pr https://github.com/chaos-mesh/chaos-mesh/pull/2072
```

After apply modifications, you can push your change to this PR via:
```bash
git push git@github.com:ti-srebot/chaos-mesh.git pr/2072:release-2.0-2c98e649556c
```

---

Close #1500.

#### Problem Summary

Since the https://github.com/gin-gonic/gin/pull/2706 be merged. Now gin allows param and exact paths to existing on the same route level.

So all of the UI assets can be safely routed by `/` and `/:foo/*bar`. This is almost the same as the `/*any`.

Except for all descriptions above, this PR also introduces some bug fixes and so on:

- Fix the problem of duplicate `uid` indices in SQLite
- Use `gorm.Model` to replace dummy fields
- Bump API server and UI versions
- Some code formats

### Related changes

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```release-note
Please add a release note.
```


STRRL (approved):

cwen0 (approved):
LGTM

PR #2098, opened 2021-07-09
Scope PSP more tightly, based on chart values

### LINKED ISSUE #2096: Daemon PSP needs to allow privileged
## Bug Report

**What did you do?**
Install chaos mesh using Helm with `--set chaosDaemon.podSecurityPolicy=true`, setting the target namespace as chaos-testing.

Daemon pods start

**What did you see instead?**

DaemonSet reports that the PSP was unable to admit the pod.

This is because the PSP doesn't have `privileged: true` set, but the daemon pods have `privileged: true` in their security context.

I'll submit a PR for this, this is an easy fix.

Niksko:
This causes the PSP that is created for the chaos daemon to both
correctly allow the pods to admit based on the default values, but also
to allow for the full range of security posture that is configurable in
this chart.

This adds the ability to:
* Set the privileged value based on chaosDaemon.privileged, but also
  bpfki.create
* Set hostNetwork based on chaosDaemon.hostNetwork
* Set allowed host ports based on bpfki.create and bfpki.grpcPort
* Set allowedHostPaths based on chaosDaemon.socketPath and bfpki.create
* Set allowedCapabilities based on chaosDameon.privileged

Issue Number: closes #2096 

Problem Summary:

Proposal: Scope PSP much more tightly, based on other values.

What's Changed:

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```release-note
PodSecurityPolicy is now more tightly scoped to the values specified in the chart
```


helm/chaos-mesh/templates/chaos-daemon-rbac.yaml
```diff
@@ -58,16 +58,45 @@ metadata:
...
+  allowPrivilegeEscalation: true
+  {{- if .Values.chaosDaemon.hostNetwork }}
   hostNetwork: true
+  {{- end }}
+  {{- if .Values.bpfki.create }}
   hostPorts:
-  - max: 65535
-    min: 0
+  - max: {{ .Values.bpfki.grpcPort }}
+    min: {{ .Values.bpfki.grpcPort }}
```
fewdan, line 95: Are both max and min set to ".Values.bpfki.grpcPort"?  In this case, is "max" equal to "min"?

helm/chaos-mesh/templates/chaos-daemon-rbac.yaml
```diff
@@ -58,16 +58,45 @@ metadata:
...
+    - NET_ADMIN
+    - MKNOD
+    - SYS_CHROOT
+    - SYS_ADMIN
+    - KILL
+    # CAP_IPC_LOCK is used to lock memory
+    - IPC_LOCK
+  {{- end }}
+  allowedHostPaths:
+    - pathPrefix: {{ .Values.chaosDaemon.socketPath | default "/var/run/docker.sock" }}
```
STRRL, line 76: that default value is not necessary, `.Values.chaosDaemon.socketPath` always exists.

helm/chaos-mesh/templates/chaos-daemon-rbac.yaml
```diff
@@ -58,16 +58,45 @@ metadata:
     {{- include "chaos-mesh.labels" . | nindent 4 }}
     app.kubernetes.io/component: chaos-daemon
 spec:
-  privileged: false
-  allowPrivilegeEscalation: true
+  {{- if .Values.chaosDaemon.privileged }}
```
STRRL, line 61: I think we should keep this line same with L99:
```
{{ or (eq .Values.chaosDaemon.privileged true) (eq .Values.bpfki.create true) }}
```

We could create another variable like `chaos-daemon-need-privileged` to reuse that expression.

STRRL (commented):
Hi @Niksko, thanks for your contribution!

There are several comments, PTAL! :heart:

helm/chaos-mesh/templates/chaos-daemon-rbac.yaml
Niksko: Yes. However this is the correct way to specify a single allowed host port.

helm/chaos-mesh/templates/chaos-daemon-rbac.yaml
Niksko: Yep, good idea, done!

STRRL (approved):
LGTM!

helm/chaos-mesh/templates/chaos-daemon-rbac.yaml
cwen0: This config should contain the ports of chaos-daemon.  https://github.com/chaos-mesh/chaos-mesh/blob/master/helm/chaos-mesh/values.yaml#L88-L89

PR #2099, opened 2021-07-09
Update PSP RoleBinding to respect release namespace

### LINKED ISSUE #2095: PSP RoleBinding has hardcoded namespace
## Bug Report

**What did you do?**
Install chaos mesh with `--set chaosDaemon.podSecurityPolicy=true`, setting the target namespace as `chaos-mesh`.

Helm chart installs properly

**What did you see instead?**

Error message from helm controller: `Helm upgrade failed: failed to create resource: namespaces "chaos-testing" not found`

This occurs because of the following chart configuration: https://github.com/chaos-mesh/chaos-mesh/blob/0ff6d685c98b2302ea28ae01f465eedd0aef583b/helm/chaos-mesh/templates/chaos-daemon-rbac.yaml#L19

The namespace of the PSP rolebinding is hardcoded to `chaos-testing`. This should either be configurable via `values.yaml`, or should just be set to the global `.Release.Namespace`

Niksko:
Currently, this value is hardcoded, which means that deploying with a
PSP to any namespace other than 'chaos-testing' will cause the daemonset
pods to fail to be admitted, because there is no valid PSP that matches
them.

Signed-off-by: Nikolas Skoufis <nskoufis@seek.com.au>

Issue Number: close #2095 

Problem Summary: Hardcoded namespace in PSP RoleBinding

What's Changed: Set the namespace of this RoleBinding to the same as the overall release namespace

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```release-note
Fixes the namespace of the PSP RoleBinding so that it matches the release namespace
```

### DIFF
--- a/helm/chaos-mesh/templates/chaos-daemon-rbac.yaml
+++ b/helm/chaos-mesh/templates/chaos-daemon-rbac.yaml
@@ -16,7 +16,7 @@ kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: {{ .Release.Name }}-chaos-daemon-target-namespace
-  namespace: {{ .Release.Namespace }}
+  namespace: {{ .Release.Namespace | quote }}
   labels:
     {{- include "chaos-mesh.labels" . | nindent 4 }}
     app.kubernetes.io/component: chaos-daemon
--- a/helm/chaos-mesh/templates/chaos-daemon-rbac.yaml
+++ b/helm/chaos-mesh/templates/chaos-daemon-rbac.yaml
@@ -16,7 +16,7 @@ kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: {{ .Release.Name }}-chaos-daemon-target-namespace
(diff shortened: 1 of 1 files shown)


helm/chaos-mesh/templates/chaos-daemon-rbac.yaml
```diff
@@ -16,7 +16,7 @@ kind: RoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
   name: {{ .Release.Name }}-chaos-daemon-target-namespace
-  namespace: chaos-testing
+  namespace: {{ .Release.Namespace }}
```
fewdan, line 19: Would it be better to change to "namespace: {{ .Release.Namespace | quote }}"  ?

Niksko: Sure, done, thanks!

fewdan (approved):
LGTM

cwen0 (approved):
LGTM

PR #2111, opened 2021-07-13
alert when no record has been selected

YangKeao:

Problem Summary:

In `PodSelector`, when no pod is selected, an error will be returned. However, in `ContainerSelector`, if no container is selected, there will be no error.

A zero-length check is added under the selection, so that when no records are selected, it will always submit an error event, even if there is no error returned.


cwen0 (approved):
LGTM

STRRL (approved):
LGTM!

PR #2122, opened 2021-07-16
Automatically generate webhook for a chaos type

YangKeao:

Redundant webhooks logic shouldn't be defined separately for every  single chaos type, as the developers are not reliable and may forget to do so.

All webhook of a chaos is replaced by a `Validate(obj interface{})` and `Default(obj interface{})` function call, and the bindings are generated automatically by `chaos-builder`.

The implementation of `Validate(obj interface{})` and `Default(obj interface{})` are provided by a field walker through `reflect`.

### Design details

The receiver of some validate implementation for a type is a pointer, because the field is `optional` and it could be nil, but shouldn't be nil at some time. We should allow the developers to define the logic to check whether a field could be nil under specific action (or other situation).

However, it leads to another problem. For example, the `PodChaosSpec` under `WorkflowSpec` could be nil, but `PodSelector` implements the `FieldValidator` interface, so the `Validate` method of `PodChaosSpec` will be called automatically, and panic :cry: .

The key of understanding this issue is that: some pointer receivers are allowed to be nil (as it's fine to be nil under some situation), but the others are not, as it's an embedded field of the parent type.

Confusing the struct embedding language feature with the json inline is really (another) a bad design of go standard library, and it hasn't been fixed yet.

In this PR, we add a special `validate:"optional"` to field, to specify that "even if it's nil, it's safe to call the `Validate`and `Default` method of it."

Tests

- [x] Unit test
- [x] E2E test
- [x] Manual test


STRRL:
> In this PR, we add a special validate:"optional" to field, to specify that "even if it's nil, it's safe to call the Validateand Default method of it."

How about using `nilable` instead of `optional`, it's a little confusing for `optional`, it leads people to think the validation for this field is "optional".

STRRL:
> However, it leads to another problem. For example, the PodChaosSpec under WorkflowSpec could be nil, but PodSelector implements the FieldValidator interface, so the Validate method of PodChaosSpec will be called automatically, and panic cry .

I think current work is enough good! We could generate the validator for each simple chaos. :clap: 

Validation for each template in `WorkflowSpec` depends on `field: TemplateType`, it might be a little harder for code generation.(same situation on Schedule)

As there will only be one struct `Schedule` and `Workflow`, I think it's ok to make validators for them manually.

api/v1alpha1/awschaos_types.go
```diff
@@ -54,11 +54,11 @@ type AWSChaosSpec struct {
 
 	// Duration represents the duration of the chaos action.
 	// +optional
-	Duration *string `json:"duration,omitempty"`
+	Duration *string `json:"duration,omitempty" webhook:"Duration"`
 
 	// SecretName defines the name of kubernetes secret.
 	// +optional
-	SecretName *string `json:"secretName,omitempty"`
+	SecretName *string `json:"secretName,omitempty" webhoo:",nilable"`
```
fewdan, line 61: "webhoo" is a typo here?

STRRL:
LGTM.

some validators like `Duration` and `Percent` is very helpful, it could be reused easily.

but some validator like `IOErrno` and `EbsVolume` is bonded with its root object(this field must be placed in certain type, for example, Errno must be in IOChaosSpec), I could not use these validators for fields that belong to other structs( it not happens now). For example, someday I write another Chaos called (IOChaosOnPhysicalMachine) with one field `Errno`, I can not resuse `webhook:"IOErrno"` directly.

STRRL (approved):

cmd/chaos-builder/impl.go
```diff
@@ -152,6 +161,43 @@ func (in *{{.Type}}) IsOneShot() bool {
 	return false
 	{{end}}
 }
+
+var {{.Type}}WebhookLog = logf.Log.WithName("awschaos-resource")
```
cwen0, line 165: s/awschaos-resource/{{.Type}}-resouce?

Hexilee:
> 
> 
> LGTM.
> 
> some validators like `Duration` and `Percent` is very helpful, it could be reused easily.
> 
> but some validator like `IOErrno` and `EbsVolume` is bonded with its root object(this field must be placed in certain type, for example, Errno must be in IOChaosSpec), I could not use these validators for fields that belong to other structs( it not happens now). For example, someday I write another Chaos called (IOChaosOnPhysicalMachine) with one field `Errno`, I can not resuse `webhook:"IOErrno"` directly.

I seem we need an interface to fit different root type for `IOErrno` and `EbsVolume`.

Hexilee (approved):
LGTM

cmd/chaos-builder/impl.go
STRRL: > s/awschaos-resource/{{.Type}}-resouce?

PTAL @YangKeao  :D

STRRL:
e2e failed because the Defaulter of `Direction` overwrites the direction from `Both` to `To`.

https://github.com/YangKeao/chaos-mesh/blob/7e00973ebd22b6873c959815bcb6b3837699b91e/api/v1alpha1/networkchaos_webhook.go#L36-L38

https://github.com/YangKeao/chaos-mesh/blob/7e00973ebd22b6873c959815bcb6b3837699b91e/api/v1alpha1/genericwebhook/defaulter.go#L44-L47

PR #2133, opened 2021-07-19
recover after failing to apply

YangKeao:

Without this PR, the former `stress-ng` instance created by controller but failed will never be collected and recoverd, because the next retry will overwrite the pid information.

Add a recover logic after failing to apply.


fewdan (approved):
LGTM

STRRL (approved):
LGTM

PR #2141, opened 2021-07-22
Check the length of parts while using ParseNamespacedName and ParseNamespacedNameContainer

### LINKED ISSUE #1933: Check the length of array before using it in `ParseNamespacedName` and  `ParseNamespacedNameContainer` functions
## Bug Report

`ParseNamespacedName` and  `ParseNamespacedNameContainer` functions don't check the length of the `parts` array before using it, which may panic the whole program when using these two functions  https://github.com/chaos-mesh/chaos-mesh/blob/master/controllers/utils/controller/key.go#L22-L36

FingerLeader:

Issue Number: close #1933

Problem Summary:

Before returning value, check the length of parts to avoid panic.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code


controllers/chaosimpl/httpchaos/impl.go
```diff
@@ -59,7 +59,11 @@ func (impl *Impl) Apply(ctx context.Context, index int, records []*v1alpha1.Reco
 
 	if phase == waitForApplySync {
 		podhttpchaos := &v1alpha1.PodHttpChaos{}
-		err := impl.Client.Get(ctx, controller.ParseNamespacedName(record.Id), podhttpchaos)
+		NamespacedName, err := controller.ParseNamespacedName(record.Id)
```
YangKeao, line 62: Please use little camel case variable name: `namespacedName`

controllers/chaosimpl/httpchaos/impl.go
```diff
@@ -59,7 +59,11 @@ func (impl *Impl) Apply(ctx context.Context, index int, records []*v1alpha1.Reco
...
+		if err != nil {
+			return waitForApplySync, err
+		}
+		err = impl.Client.Get(ctx, NamespacedName, podhttpchaos)
```
YangKeao, line 66: ditto

controllers/chaosimpl/httpchaos/impl.go
```diff
@@ -126,7 +133,12 @@ func (impl *Impl) Recover(ctx context.Context, index int, records []*v1alpha1.Re
 	phase := record.Phase
 	if phase == waitForRecoverSync {
 		podhttpchaos := &v1alpha1.PodHttpChaos{}
-		err := impl.Client.Get(ctx, controller.ParseNamespacedName(record.Id), podhttpchaos)
+		NamespacedName, err := controller.ParseNamespacedName(record.Id)
```
YangKeao, line 136: ditto

controllers/chaosimpl/httpchaos/impl.go
```diff
@@ -126,7 +133,12 @@ func (impl *Impl) Recover(ctx context.Context, index int, records []*v1alpha1.Re
...
+		if err != nil {
+			// This error is not expected to exist
+			return waitForRecoverSync, err
+		}
+		err = impl.Client.Get(ctx, NamespacedName, podhttpchaos)
```
YangKeao, line 141: ditto

controllers/chaosimpl/iochaos/impl.go
```diff
@@ -58,7 +58,11 @@ func (impl *Impl) Apply(ctx context.Context, index int, records []*v1alpha1.Reco
 
 	if phase == waitForApplySync {
 		podiochaos := &v1alpha1.PodIOChaos{}
-		err := impl.Client.Get(ctx, controller.ParseNamespacedName(record.Id), podiochaos)
+		NamespacedName, err := controller.ParseNamespacedName(record.Id)
```
YangKeao, line 61: ditto

controllers/chaosimpl/iochaos/impl.go
```diff
@@ -140,7 +147,12 @@ func (impl *Impl) Recover(ctx context.Context, index int, records []*v1alpha1.Re
 	phase := record.Phase
 	if phase == waitForRecoverSync {
 		podiochaos := &v1alpha1.PodIOChaos{}
-		err := impl.Client.Get(ctx, controller.ParseNamespacedName(record.Id), podiochaos)
+		NamespacedName, err := controller.ParseNamespacedName(record.Id)
```
YangKeao, line 150: ditto

controllers/chaosimpl/jvmchaos/impl.go
```diff
@@ -40,7 +40,11 @@ func (impl *Impl) Apply(ctx context.Context, index int, records []*v1alpha1.Reco
 	jvmchaos := obj.(*v1alpha1.JVMChaos)
 
 	var pod v1.Pod
-	err := impl.Client.Get(ctx, controller.ParseNamespacedName(records[index].Id), &pod)
+	NamespacedName, err := controller.ParseNamespacedName(records[index].Id)
```
YangKeao, line 43: ditto

controllers/chaosimpl/jvmchaos/impl.go
```diff
@@ -87,7 +91,12 @@ func (impl *Impl) Recover(ctx context.Context, index int, records []*v1alpha1.Re
...
```
YangKeao, line 94: ditto

controllers/chaosimpl/networkchaos/partition/impl.go
```diff
@@ -68,7 +68,11 @@ func (impl *Impl) Apply(ctx context.Context, index int, records []*v1alpha1.Reco
 
 	if phase == waitForApplySync {
 		podnetworkchaos := &v1alpha1.PodNetworkChaos{}
-		err := impl.Client.Get(ctx, controller.ParseNamespacedName(record.Id), podnetworkchaos)
+		NamespacedName, err := controller.ParseNamespacedName(record.Id)
```
YangKeao, line 71: ditto

controllers/chaosimpl/networkchaos/partition/impl.go
```diff
@@ -95,7 +99,11 @@ func (impl *Impl) Apply(ctx context.Context, index int, records []*v1alpha1.Reco
 	}
 
 	var pod v1.Pod
-	err := impl.Client.Get(ctx, controller.ParseNamespacedName(record.Id), &pod)
+	NamespacedName, err := controller.ParseNamespacedName(record.Id)
```
YangKeao, line 102: ditto

controllers/chaosimpl/networkchaos/partition/impl.go
```diff
@@ -244,7 +252,12 @@ func (impl *Impl) Recover(ctx context.Context, index int, records []*v1alpha1.Re
 
 	if phase == waitForRecoverSync {
 		podnetworkchaos := &v1alpha1.PodNetworkChaos{}
-		err := impl.Client.Get(ctx, controller.ParseNamespacedName(record.Id), podnetworkchaos)
+		NamespacedName, err := controller.ParseNamespacedName(record.Id)
```
YangKeao, line 255: ditto

controllers/chaosimpl/networkchaos/partition/impl.go
```diff
@@ -265,7 +278,12 @@ func (impl *Impl) Recover(ctx context.Context, index int, records []*v1alpha1.Re
 	}
 
 	var pod v1.Pod
-	err := impl.Client.Get(ctx, controller.ParseNamespacedName(record.Id), &pod)
+	NamespacedName, err := controller.ParseNamespacedName(record.Id)
```
YangKeao, line 281: ditto

controllers/chaosimpl/networkchaos/partition/impl.go
```diff
@@ -324,7 +342,12 @@ func (impl *Impl) SetDrop(ctx context.Context, m *podnetworkchaosmanager.PodNetw
 	targetPods := []v1.Pod{}
 	for _, record := range targets {
 		var pod v1.Pod
-		err := impl.Client.Get(ctx, controller.ParseNamespacedName(record.Id), &pod)
+		NamespacedName, err := controller.ParseNamespacedName(record.Id)
```
YangKeao, line 345: ditto

controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
```diff
@@ -71,7 +71,11 @@ func (impl *Impl) Apply(ctx context.Context, index int, records []*v1alpha1.Reco
 
 	if phase == waitForApplySync {
 		podnetworkchaos := &v1alpha1.PodNetworkChaos{}
-		err := impl.Client.Get(ctx, controller.ParseNamespacedName(record.Id), podnetworkchaos)
+		NamespacedName, err := controller.ParseNamespacedName(record.Id)
```
YangKeao, line 74: ditto

controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
```diff
@@ -98,7 +102,11 @@ func (impl *Impl) Apply(ctx context.Context, index int, records []*v1alpha1.Reco
 	}
 
 	var pod v1.Pod
-	err := impl.Client.Get(ctx, controller.ParseNamespacedName(record.Id), &pod)
+	NamespacedName, err := controller.ParseNamespacedName(record.Id)
```
YangKeao, line 105: ditto

controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
```diff
@@ -182,7 +190,12 @@ func (impl *Impl) Recover(ctx context.Context, index int, records []*v1alpha1.Re
 
 	if phase == waitForRecoverSync {
 		podnetworkchaos := &v1alpha1.PodNetworkChaos{}
-		err := impl.Client.Get(ctx, controller.ParseNamespacedName(record.Id), podnetworkchaos)
+		NamespacedName, err := controller.ParseNamespacedName(record.Id)
```
YangKeao, line 193: ditto

controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
```diff
@@ -203,7 +216,12 @@ func (impl *Impl) Recover(ctx context.Context, index int, records []*v1alpha1.Re
 	}
 
 	var pod v1.Pod
-	err := impl.Client.Get(ctx, controller.ParseNamespacedName(record.Id), &pod)
+	NamespacedName, err := controller.ParseNamespacedName(record.Id)
```
YangKeao, line 219: ditto

controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
```diff
@@ -267,7 +285,12 @@ func (impl *Impl) ApplyTc(ctx context.Context, m *podnetworkchaosmanager.PodNetw
 	targetPods := []v1.Pod{}
 	for _, record := range targets {
 		var pod v1.Pod
-		err := impl.Client.Get(ctx, controller.ParseNamespacedName(record.Id), &pod)
+		NamespacedName, err := controller.ParseNamespacedName(record.Id)
```
YangKeao, line 288: ditto

controllers/chaosimpl/podchaos/podfailure/impl.go
```diff
@@ -39,7 +39,11 @@ func (impl *Impl) Apply(ctx context.Context, index int, records []*v1alpha1.Reco
 	podchaos := obj.(*v1alpha1.PodChaos)
 
 	var origin v1.Pod
-	err := impl.Get(ctx, controller.ParseNamespacedName(records[index].Id), &origin)
+	NamespacedName, err := controller.ParseNamespacedName(records[index].Id)
```
YangKeao, line 42: ditto

controllers/chaosimpl/podchaos/podfailure/impl.go
```diff
@@ -92,7 +96,12 @@ func (impl *Impl) Recover(ctx context.Context, index int, records []*v1alpha1.Re
...
```
YangKeao, line 99: ditto

controllers/chaosimpl/podchaos/podkill/impl.go
```diff
@@ -31,7 +31,11 @@ func (impl *Impl) Apply(ctx context.Context, index int, records []*v1alpha1.Reco
 	podchaos := obj.(*v1alpha1.PodChaos)
 
 	var pod v1.Pod
-	err := impl.Get(ctx, controller.ParseNamespacedName(records[index].Id), &pod)
+	NamespacedName, err := controller.ParseNamespacedName(records[index].Id)
```
YangKeao, line 34: ditto

controllers/common/fx.go
```diff
@@ -94,7 +94,11 @@ func NewController(params Params) (types.Controller, error) {
 						for i := 0; i < items.Len(); i++ {
 							item := items.Index(i).Addr().Interface().(InnerObjectWithSelector)
 							for _, record := range item.GetStatus().Experiment.Records {
-								if controller.ParseNamespacedName(record.Id) == objName {
+								NamespacedName, err := controller.ParseNamespacedName(record.Id)
```
YangKeao, line 97: ditto

YangKeao (changes requested):

controllers/chaosimpl/httpchaos/impl.go
```diff
@@ -75,9 +79,12 @@ func (impl *Impl) Apply(ctx context.Context, index int, records []*v1alpha1.Reco
 		return waitForApplySync, nil
 	}
 
-	podId, _ := controller.ParseNamespacedNameContainer(records[index].Id)
+	podId, _, err := controller.ParseNamespacedNameContainer(records[index].Id)
```
YangKeao, line 82: Use `ParseNamespacedName` but not `ParseNamespacedNameContainer`

YangKeao (approved):

controllers/chaosimpl/httpchaos/impl.go
```diff
@@ -153,9 +165,13 @@ func (impl *Impl) Recover(ctx context.Context, index int, records []*v1alpha1.Re
 		return waitForRecoverSync, nil
 	}
 
-	podId, _ := controller.ParseNamespacedNameContainer(records[index].Id)
+	podId, _, err := controller.ParseNamespacedNameContainer(records[index].Id)
```
YangKeao, line 168: Use `ParseNamespacedName` but not `ParseNamespacedNameContainer`

YangKeao (changes requested):

fewdan:
it seems that CI failed.  you can exec "make check" before "git add ."   .

FingerLeader:
> it seems that CI failed. you can exec "make check" before "git add ." .

I can't run 'make check' and 'make test' on my Linux for some unknown errors, I'll try it later.

controllers/chaosimpl/utils/record.go
```diff
@@ -46,7 +46,12 @@ type DecodedContainerRecord struct {
 
 func (d *ContianerRecordDecoder) DecodeContainerRecord(ctx context.Context, record *v1alpha1.Record) (decoded DecodedContainerRecord, err error) {
 	var pod v1.Pod
-	podId, containerName := controller.ParseNamespacedNameContainer(record.Id)
+	podId, containerName, err := controller.ParseNamespacedNameContainer(record.Id)
+	if err != nil {
+		// TODO: organize the error in a better way
+		err = NewFailToFindContainer(pod.Namespace, pod.Name, containerName, err)
```
Colstuwjx, line 52: If it parses failed, both of the `containerName` and `podId` should be the empty strings, it's non-sense to define this error, I suggest just return it.

FingerLeader: I did just as you say, but httpchaos threw an error of can't find pod "", so I add this error.

FingerLeader: But I find that may be the wrong use of ```NewFailToFindContainer``` in httpchaos

Colstuwjx: The `ParseNamespacedNameContainer` returns error should be sth like `too few parts of namespacedname` as [L51](https://github.com/chaos-mesh/chaos-mesh/pull/2141/files#diff-fad6897b0163bb2450778c4bade5d28c847364e881670dd290b3d83b602dd925R51), it's different to `can't find pod ""`.

FingerLeader: That happended when I haven't add this error.

YangKeao: Why? `HTTPChaos` doesn't call `ParseNamespacedNameContainer` now. How could it get an error from the function which it doesn't call?

Colstuwjx: Which line would the httpchaos return that error `can't find pod ""`?

There is nothing magic about `NewFailToFindContainer` method, it just exposed `Error` function to make it works as an error. And the outside world even couldn't check its original fields..

FingerLeader: This is the e2e-test log of the past without error adding
```
[2021-07-26T04:36:51.037Z] [3] Jul 26 04:36:50.016: FAIL: http chaos doesn't work as expected

[2021-07-26T04:36:51.037Z] [3] Unexpected error:

[2021-07-26T04:36:51.037Z] [3]     <*errors.errorString | 0xc000353f00>: {

[2021-07-26T04:36:51.037Z] [3]         s: "timed out waiting for the condition",

[2021-07-26T04:36:51.037Z] [3]     }

[2021-07-26T04:36:51.037Z] [3]     timed out waiting for the condition

[2021-07-26T04:36:51.037Z] [3] occurred

[2021-07-26T04:36:51.037Z] [3] [JustAfterEach] [HTTPChaos]

[2021-07-26T04:36:51.037Z] [3]   /home/jenkins/agent/workspace/chaos-mesh-e2e/go/src/github.com/chaos-mesh/chaos-mesh/e2e-test/e2e/chaos/basic.go:251

[2021-07-26T04:36:51.037Z] [3] [AfterEach] [Basic]

[2021-07-26T04:36:51.037Z] [3]   /go/pkg/mod/k8s.io/kubernetes@v1.17.2/test/e2e/framework/framework.go:152

[2021-07-26T04:36:51.037Z] [3] STEP: Collecting events from namespace "chaos-mesh-3790".

[2021-07-26T04:36:51.037Z] [3] STEP: Found 25 events.

[2021-07-26T04:36:51.037Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:45 +0000 UTC - event for http-test: {deployment-controller } ScalingReplicaSet: Scaled up replica set http-test-6c4c75854b to 1

[2021-07-26T04:36:51.037Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:45 +0000 UTC - event for http-test-6c4c75854b: {replicaset-controller } SuccessfulCreate: Created pod: http-test-6c4c75854b-pfl7j

[2021-07-26T04:36:51.037Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:45 +0000 UTC - event for http-test-6c4c75854b-pfl7j: {default-scheduler } Scheduled: Successfully assigned chaos-mesh-3790/http-test-6c4c75854b-pfl7j to chaos-mesh-worker2

[2021-07-26T04:36:51.037Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:46 +0000 UTC - event for http-test-6c4c75854b-pfl7j: {kubelet chaos-mesh-worker2} Started: Started container http

[2021-07-26T04:36:51.037Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:46 +0000 UTC - event for http-test-6c4c75854b-pfl7j: {kubelet chaos-mesh-worker2} Created: Created container http

[2021-07-26T04:36:51.037Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:46 +0000 UTC - event for http-test-6c4c75854b-pfl7j: {kubelet chaos-mesh-worker2} Pulled: Container image "localhost:5000/pingcap/e2e-helper:latest" already present on machine

[2021-07-26T04:36:51.037Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:49 +0000 UTC - event for http-chaos: {desiredphase } Started: Experiment has started

[2021-07-26T04:36:51.037Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:49 +0000 UTC - event for http-chaos: {finalizer } FinalizerInited: Finalizer has been inited

[2021-07-26T04:36:51.037Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:50 +0000 UTC - event for http-chaos: {records } Failed: Failed to apply chaos: Pod "" not found

[2021-07-26T04:36:51.037Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:50 +0000 UTC - event for http-chaos: {records } Failed: Failed to apply chaos: Pod "" not found

[2021-07-26T04:36:51.037Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:50 +0000 UTC - event for http-chaos: {records } Failed: Failed to apply chaos: Pod "" not found

[2021-07-26T04:36:51.038Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:50 +0000 UTC - event for http-chaos: {records } Failed: Failed to apply chaos: Pod "" not found

[2021-07-26T04:36:51.038Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:50 +0000 UTC - event for http-chaos: {records } Failed: Failed to apply chaos: Pod "" not found

[2021-07-26T04:36:51.038Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:50 +0000 UTC - event for http-chaos: {desiredphase } Updated: Successfully update desiredPhase of resource

[2021-07-26T04:36:51.038Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:50 +0000 UTC - event for http-chaos: {finalizer } Updated: Successfully update finalizer of resource

[2021-07-26T04:36:51.038Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:50 +0000 UTC - event for http-chaos: {records } Updated: Successfully update records of resource

[2021-07-26T04:36:51.038Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:50 +0000 UTC - event for http-chaos: {records } Failed: Failed to apply chaos: Pod "" not found

[2021-07-26T04:36:51.038Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:50 +0000 UTC - event for http-chaos: {records } Failed: Failed to apply chaos: Pod "" not found

[2021-07-26T04:36:51.038Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:50 +0000 UTC - event for http-chaos: {records } Failed: Failed to apply chaos: Pod "" not found

[2021-07-26T04:36:51.038Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:51 +0000 UTC - event for http-chaos: {records } Failed: Failed to apply chaos: Pod "" not found

[2021-07-26T04:36:51.038Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:52 +0000 UTC - event for http-chaos: {records } Failed: Failed to apply chaos: Pod "" not found

[2021-07-26T04:36:51.038Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:35:55 +0000 UTC - event for http-chaos: {records } Failed: Failed to apply chaos: Pod "" not found

[2021-07-26T04:36:51.038Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:36:00 +0000 UTC - event for http-chaos: {records } Failed: Failed to apply chaos: Pod "" not found

[2021-07-26T04:36:51.038Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:36:10 +0000 UTC - event for http-chaos: {records } Failed: Failed to apply chaos: Pod "" not found

[2021-07-26T04:36:51.038Z] [3] Jul 26 04:36:50.022: INFO: At 2021-07-26 04:36:31 +0000 UTC - event for http-chaos: {records } Failed: Failed to apply chaos: Pod "" not found
```

Colstuwjx: There are many commits after that time e2e test, how about trigger a new CI test, and let's see what gonna happen.

Colstuwjx: Besides, we could see the latest e2e test is passed, so I think it's not a problem for you mentioned case.

FingerLeader: Which version of code do you referring to?  The one without error throwing?

Colstuwjx: I mean we could just return the error as I suggested, and let's see the ci test result. If there are still some errors, I would help you dive into the root cause.

STRRL:
You should not use `ParseNamespacedName` within `IOChaos`, because the selected records are like `<namespace>/<podname>/<contianerid>`, and the container id might be the empty string. 

After you add the restrict for `len(parts) == 2`, the prase of record could not work anymore... because there are always be three parts split by two slash

Colstuwjx:
> After you add the restrict for len(parts) == 2, the prase of record could not work anymore... because there are always be three parts split by two slash

Oh, that is the root cause. It's a special case, and hard to debug, how to avoid this issue happen again?

YangKeao:
> > After you add the restrict for len(parts) == 2, the prase of record could not work anymore... because there are always be three parts split by two slash
> 
> Oh, that is the root cause. It's a special case, and hard to debug, how to avoid this issue happen again?

A more hard to find problem also locates here:

```go
for _, record := range item.GetStatus().Experiment.Records {
    if controller.ParseNamespacedName(record.Id) == objName {
	    id := k8sTypes.NamespacedName{
		    Namespace: item.GetObjectMeta().Namespace,
		    Name:      item.GetObjectMeta().Name,
	    }
	    setupLog.Info("mapping requests", "source", objName, "target", id)
	    reqs = append(reqs, reconcile.Request{
		    NamespacedName: id,
	    })
    }
}
```

In `common/fx.go`, it maps the modification on `podxxxchaos` to the `xxxchaos` through the records. And it works for both `ControllerSelector` like `IOChaos`, and also for `PodSelector` like `NetworkChaos`. If it returns empty struct, then it will fail to update `IOChaos` when the related `podiochaos` is updated.

Colstuwjx:
> In common/fx.go, it maps the modification on podxxxchaos to the xxxchaos through the records. And it works for both ControllerSelector like IOChaos, and also for PodSelector like NetworkChaos. If it returns empty struct, it will fail to update IOChaos when the related podiochaos is updated.

For my understanding, we're using `record` to unify all kinds of chaos resources, and from now, the record id, also is the primary key of our system, could be several forms:

* <namespace>/<podname> ( most of cases? )
* <namespace>/<podname>/<containerid>

The key point is how to determine a chaos record by its id in the controller-manager. Now we're using `ParseNamespacedName` and `ParseNamespacedNameContainer` as such util functions to solve this problem.

Currently, I think we could just keep using these two functions in this PR, and assume that each record id should start with `<namespace>/<podname>`, but it may not be only two parts.

A more robust solution, in my opinion, should be that **give a way to determine a record by its ID and type**. When we're calling the `ParseNamespacedName`, it usually follows a Get operation, which means we know its object type, except the case @YangKeao mentioned, the fx entrypoint. Could we find out a way to determine the record **type**, and then parse the id according to its type?

YangKeao:
> > In common/fx.go, it maps the modification on podxxxchaos to the xxxchaos through the records. And it works for both ControllerSelector like IOChaos, and also for PodSelector like NetworkChaos. If it returns empty struct, it will fail to update IOChaos when the related podiochaos is updated.
> 
> For my understanding, we're using `record` to unify all kinds of chaos resources, and from now, the record id, also is the primary key of our system, could be several forms:
> 
> * \<namespace\>/\<podname\> ( most of cases? )
> * \<namespace\>/\<podname\>/\<containerid\>
> 
> The key point is how to determine a chaos record by its id in the controller-manager. Now we're using `ParseNamespacedName` and `ParseNamespacedNameContainer` as such util functions to solve this problem.
> 
> Currently, I think we could just keep using these two functions in this PR, and assume that each record id should start with `<namespace>/<podname>`, but it may not be only two parts.
> 
> A more robust solution, in my opinion, should be that **give a way to determine a record by its ID and type**. When we're calling the `ParseNamespacedName`, it usually follows a Get operation, which means we know its object type, except the case @YangKeao mentioned, the fx entrypoint. Could we find out a way to determine the record **type**, and then parse the id according to its type?
> 

Actually yes. It can be determined from the return type of `GetSelectors`. If it is a `PodSelector`, it should be parsed with `ParseNamespacedName`, and for `ContainerSelector`, it should be `PaeseNamespavedNameContainer`.

Only need to do some type assertion.

Colstuwjx:
Hi @FingerLeader ,

It's a long trip since I've seen this PR, I previously thought it may be just an issue-fix PR!

For a further plan, I think we need to make a better way to find out the target object by record-id, and I think we've had some ideas now, as @YangKeao mentioned, sth like type assertion, but I think it could be another PR's job!

From now, IMO, how about adding an error back, for the `ParseNamespacedName` and `ParseNamespacedNameContainer` ?

If we've used the function in the wrong way again, the returned error could help us clarify the issue.

FingerLeader:
> Hi @FingerLeader ,
> 
> It's a long trip since I've seen this PR, I previously thought it may be just an issue-fix PR!
> 
> For a further plan, I think we need to make a better way to find out the target object by record-id, and I think we've had some ideas now, as @YangKeao mentioned, sth like type assertion, but I think it could be another PR's job!
> 
> From now, IMO, how about adding an error back in the return parameters, for the `ParseNamespacedName` and `ParseNamespacedNameContainer` ?
> 
> If we've used the function in the wrong way again, the returned error could help us clarify the issue.

Should I rebase my commit to that error-added one?

Colstuwjx:
> Should I rebase my commit to that error-added one?

All roads lead to Rome, you could use a way that you're familiar with, e.g. rebase or cherry-pick the commit and correct the issue we've met.

Colstuwjx:
Hi @FingerLeader ,

I've been reverted the unexpected changes towards some files, and I have some suggestions for git operations:

1. check out a specific branch for changes, rather than directly using the master branch of your forked repo, and each branch should only do one thing, after the PR merged, the branch should be also deleted;
2. try `git diff xxx > patch` and `git apply patch` such techniques if you want to quickly apply some changes on the source commit/branch/tag/files;

Happy hack!

FingerLeader:
> Hi @FingerLeader ,
> 
> I've been reverted the unexpected changes towards some files, and I have some suggestions for git operations:
> 
> 1. check out a specific branch for changes, rather than directly using the master branch of your forked repo, and each branch should only do one thing, after the PR merged, the branch should be also deleted;
> 2. try `git diff xxx > patch` and `git apply patch` such techniques if you want to quickly apply some changes on the source commit/branch/tag/files;
> 
> Happy hack!

Thanks for your suggestion! I'll have a try next time

controllers/common/fx.go
```diff
@@ -94,7 +94,11 @@ func NewController(params Params) (types.Controller, error) {
 						for i := 0; i < items.Len(); i++ {
 							item := items.Index(i).Addr().Interface().(InnerObjectWithSelector)
 							for _, record := range item.GetStatus().Experiment.Records {
-								if controller.ParseNamespacedName(record.Id) == objName {
+								namespacedName, err := controller.ParseNamespacedName(record.Id)
+								if err != nil {
+									setupLog.Error(err, "fail to get pods")
```
Colstuwjx, line 99: I prefer `setupLog.Error(err, "failed to parse record", "record", record)`.

controllers/chaosimpl/utils/record.go
Colstuwjx: It's still an issue.

Colstuwjx (changes requested):

Colstuwjx:
The file mode changes are unexpected behavior, please revert them, @FingerLeader.

FingerLeader:
Actually, I don't know why the file modes are changed. After reseting my branch to ```084d57a```, the modes are still change to 10755 when I commit the branch.

Colstuwjx:
I wonder if it's due to sth like git index cache, you could try the following commands to rollback to the upstream branch state:

```
git remote add upstream git@github.com:chaos-mesh/chaos-mesh.git
git fetch upstream
git diff -p -R --no-ext-diff --no-color upstream/master| grep -E "^(diff|(old|new) mode)" --color=never > patch
git apply patch
```

For a simple solution about this, I think you could just delete your local repo and re-clone it?

FingerLeader:
I set ```filemode = false``` in ```.git/config```, and it seems work.

controllers/common/fx.go
```diff
@@ -94,7 +94,11 @@ func NewController(params Params) (types.Controller, error) {
 						for i := 0; i < items.Len(); i++ {
 							item := items.Index(i).Addr().Interface().(InnerObjectWithSelector)
 							for _, record := range item.GetStatus().Experiment.Records {
-								if controller.ParseNamespacedName(record.Id) == objName {
+								namespacedName, err := controller.ParseNamespacedName(record.Id)
+								if err != nil {
+									setupLog.Error(err, "failed to parse record", "record", record.Id)
```
Colstuwjx, line 99: It should follow a `continue`.

Colstuwjx (approved):
LGTM!

FingerLeader:
> LGTM!

It's really a long way to get this LGTM. Thanks for your help!

YangKeao:
So `len(parts) > 1` (which includes `xx/xx/xx`) are all regarded as a valid `NamespacedName`?

@STRRL suggested that the `xx/xx/xx` input should be illegal for the `ParseNamespacedName`. Will you modify it in the future PR or just prefer the current choice?

FingerLeader:
> So `len(parts) > 1` (which includes `xx/xx/xx`) are all regarded as a valid `NamespacedName`?
> 
> @STRRL suggested that the `xx/xx/xx` input should be illegal for the `ParseNamespacedName`. Will you modify it in the future PR or just prefer the current choice?

But IOChaos uses `ParseNamespacedName` with input in the format of `xx/xx/xx`, I have no idea about solving this now. I perfer to modify this PR first. And I'll try to find a method to select correct function according to Chaos' typy later.

YangKeao (approved):
LGTM

YangKeao:
> But IOChaos uses `ParseNamespacedName` with input in the format of `xx/xx/xx`, I have no idea about solving this now. I perfer to modify this PR first. And I'll try to find a method to select correct function according to Chaos' typy later.

Please also record an issue as the TODO (just in case you forget).

/merge

PR #2168, opened 2021-07-29
fix data race in client and daemon

YangKeao:

Fix two possible data race and add an `UpdatedStatusClient`:

1. The `blockingBuffer` is full of data race while reading and writing at the same time.
2. The object returned from the `cache` could be the same one, and the reference stored in it will be simply copied out. Run `DeepCopy` before assigning it to the new one.


Hexilee (approved):
LGTM

STRRL (approved):
LGTM!

STRRL:
does this PR need to be cherrypicked into release-2.0? @YangKeao

PR #2180, opened 2021-08-03
fix:status of Chaos should remain the same when an error is encountered

FingerLeader:

According to  the design, status of Chaos should remain the same when an error is encountered in `impl.Client.Get()`  and Pod is found. There are some implementations are wrong, including `httpchaos`, `iochaos` and `podfailure`

I change the status to their origin status.

Tests

- [x] Unit test
- [x] E2E test
- [ ] No code

Side effects


STRRL (approved):
LGTM!

fewdan (approved):
LGTM

PR #2183, opened 2021-08-04
fix:The patched annotation will overwrite the original annotation of the pod

wenliumeng:
fix：

The patched annotation will overwrite the original annotation of the pod.

To ensure deterministic patches sort the keys.

Problem Summary:

* Modified the ifelse logic so that the logic in the first if statement does not cover all annotations.

* Traverse the map instead of traversing the string array.

What's Changed:

### Related changes

Tests
- [√] Unit test
- [ ] E2E test
- [ ] No code

Side effects


STRRL:
Why do we need "To ensure deterministic patches sort the keys"?

`Labels` and `Annotations` are stored as `map[string]string` in golang, which is not guaranteed stable order of keys.

rest LGTM

wenliumeng:
> Why do we need "To ensure deterministic patches sort the keys"?
> 
> `Labels` and `Annotations` are stored as `map[string]string` in golang, which is NOT claimed as stable order of keys.
> 
> rest LGTM

map is unordered. Because the return value of this method is an array. Ensure that the sequence in the loop will be safer

STRRL:
> Ensure that the sequence in the loop will be safer

I do NOT think so. 😰 The order of patch operations is not required. If the order is required, we should specify it as an interface explicitly. Do not do unnecessary things. :) 

Too many bugs occur because of these "take the granted" things.

wenliumeng:
> > Ensure that the sequence in the loop will be safer
> 
> I do NOT think so. 😰 The order of patch operations is not required. If the order is required, we should specify it as an interface explicitly. Do not do unnecessary things. :)
> 
> Too many bugs occur because of these "take the granted" things.

I think i agree。👍

STRRL:
Hi @wenliumeng , let's move on!

How about removing that `sort.Strings(keys)`?

STRRL (approved):
LGTM!

YangKeao (approved):
LGTM

PR #2193, opened 2021-08-09
bump controller runtime to 0.9.5

### LINKED ISSUE #2179: bump the version of controller-runtime
We have not updated the version of several libraries for a long time, as most of those libraries stay at `0.x.x` version, and updating them means bringing many API changes and lots of handwork. And the "dependabot" seems not works well now(still need to profile the problem).

I am trying to bump the version of controller-runtime(and library about kubernetes) from v0.5.0 to the latest version (v0.9.5 at 2021-08-03). I have took a try on https://github.com/STRRL/chaos-mesh/tree/bump-controller-runtime-and-kuberntes. Some core interfaces in `controller-runtime`, like `Reconciler`, `client.Client`, `client.Object` DO make changes, but I think these changes do the great work! I am very glad to change our codes to fit the upgrading of controller-runtime.

I think we should keep eyes on the updating of the version of dependencies in the future, it brings many fantastic new features.

There are several things for completing the upgrading, I will track the progress in this issue:

- [ ] upgrade the controller-runtime to 0.9.5
- [ ] upgrade the kubernetes models to 1.21.3 (which used in controller-runtime 0.9.5) (I remember there is a script for auto-generation the "replace" in the go.mod for pinning the version, https://github.com/kubernetes/kubernetes/issues/79384#issuecomment-521493597)
- [ ] update the kubernetes models in e2e test(optional)

STRRL:

Issue Number: close #2179

This PR will NOT change any logic. Only change codes to be compatible with new APIs of `controller-runtime`, `client-go`, and other upstream libraries.

What's Changed:

- bump the version of controller-runtime from v0.4.1 to v0.9.5
- bump the version of k8s.io/* library from 0.17.x to 0.21.x
- bump go version from go1.15 to go1.16

Some API changed while bumping the version:

- `Reconciler` in `controller-runtime`, a `context.Context` is introduced as one new parameter in method `Reconcile()`.
- `controller-runtime` use `client.Object` instead of `runtime.Object`. `client.Object` is an interface embeds both `runtime.Object` and `metav1.Object`.
- clientsets in `clinet-go` also add `context.Context` in its methods

## Related changes

* PR to update `chaos-mesh/website`/`chaos-mesh/website-zh`: No.
* Need to update Chaos Dashboard component, related issue: No.
* Need to cheery-pick to the release branch: No.

### Checklist

Tests

- [ ] Unit test
- [x] E2E test
- [ ] No code

Side effects

### Release note 

```release-note
- bump controler-runtime to v0.9.5
- bump kubernetes model libraries to v0.21.3
- bump go version to go1.16
```


STRRL:
/run-e2e-tests/run-e2e-tests

WangXiangUSTC:
There are some vulnerabilities found by @sonatype-lift, do we need to update these versions?

STRRL:
> There are some vulnerabilities found by @sonatype-lift, do we need to update these versions?

I think I would upgrade dependencies in the next PRs.

There are soo many issues found by lift, and it seems part of them have no obviously resolving solution, it might take a long time to fix them all. :sob: 

I prefer to merge this PR first.

hack/update-kubernetes-library.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
```
WangXiangUSTC, line 1: Where is this script used?

e2e-test/e2e/e2e.go
```diff
@@ -158,11 +163,11 @@ var _ = ginkgo.SynchronizedBeforeSuite(func() []byte {
...
-var _ = ginkgo.SynchronizedAfterSuite(func() {
-	framework.CleanupSuite()
-}, func() {
-	framework.AfterSuiteActions()
-})
+//var _ = ginkgo.SynchronizedAfterSuite(func() {
+//	framework.CleanupSuite()
+//}, func() {
+//	framework.AfterSuiteActions()
+//})
```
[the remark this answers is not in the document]
STRRL: updated!

hack/update-kubernetes-library.sh
STRRL: I use this script for generating the `replace k8s.io/*` directive in `go.mod`.

It will only be used when updating kubernetes library.

api/v1alpha1/common_types.go
```diff
@@ -116,6 +120,7 @@ type MetaObject interface {
 
 // +kubebuilder:object:generate=false
 
+// TODO: refactor/cleanup ChaosInstance only used in common controller, and only field Name is used, I think it could be replace with other kind kie GenericChaos in this usage.
```
WangXiangUSTC, line 123: what is the meaning of `kie`? or a typo?

api/v1alpha1/common_types.go
```diff
@@ -134,5 +139,8 @@ type ChaosInstance struct {
 // ChaosList defines a common interface for chaos lists
 type ChaosList interface {
 	runtime.Object
+	metav1.ListInterface
+	// TODO: refactor: move this DeepCopyList() to upper level, rename to SpawnList()
```
WangXiangUSTC, line 143: what is the `upper level`? And why need to rename it?

api/v1alpha1/common_types.go
STRRL: yes it's a typo, it should be "like". :D

api/v1alpha1/common_types.go
STRRL: not only rename it, but this method should also be moved to the "upper level", because developers usually forget to use deepcopy, and the origin struct in the global map will be dirty. 

for example, the common usage about this `ChaosList` is mostly like:


https://github.com/chaos-mesh/chaos-mesh/blob/984b3b3140e9f9b11959c8ae922a66d5f6ab26cf/controllers/schedule/utils/list.go#L33-L39

we fetch the `ChaosKind` from the global map then get the `ChaosList`, then invoke the `DeepCopyList()` manually. If some developer forgets to invoke `DeepCopyList()`, bugs MIGHT BE created, but it's hard to find because no compiler error will be throw.

I want remove this `DeepCopyList()` for `ChaosList`, and add new method called `SpawnNewList()`( or `SpawnDeepCopiedList()`) for the struct `ChaosKind`, so developers do NOT need call `DeepCopyList()` manually.

Same manually `DeepCopy()` situation also happens on the `Chaos client.Object` in `ChaosKind`:

```golang
type ChaosKind struct {
	Chaos client.Object // <- I mean this
	ChaosList
}
```

STRRL: tldr; I want:

```golang
type ChaosKind struct {
	chaos client.Object // private
	list ChaosList // private
}

func (it *ChaosKind) SpawnObject() client.Object {
	// implementation
}

func (it *ChaosKind) SpawnList() ChaosList {
	// implementation
}
```

@FingerLeader , if you have interest in this thing, you could have a try and raise another PR after this PR merged. :heart:

FingerLeader: OK, I think I can do this.

WangXiangUSTC (approved):
LGTM

STRRL:
conflicts resolved!

STRRL:
PTAL

/cc @cwen0 
/cc @YangKeao 
/cc @Hexilee

Hexilee (approved):
LGTM

YangKeao (approved):
LGTM

PR #2217, opened 2021-08-12
Replace ChaosInstance and ChaosList with GenericChaos and GenericChaosList

FingerLeader:
### What problem does this PR solve?
`ChaosInstance` and `ChaosList` are similar to `GenericChaos` and `GenericChaosList`, We can get all the data which ChaosInstance has from `metav1.Object`, so I combine them to reduce redundancy.
### What is changed and how it works?

What's Changed:
* `ChaosInstance` and `ChaosList` are removed
* `InnerObject` doesn't have method `GetChaos()` anymore, you can get data by using `InnerObject.GetXXX()` ranther than `InnerObject.GetChaos().XXX`
* `cmd/chaos-builder/test.go` doesn't generate `TestXXXGetChaos()` anymore

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test (add detailed scripts or steps below)
- [ ] No code

Side effects

- [ ] Breaking backward compatibility

PR #1368, opened 2021-01-07
Fix: client.Get gets empty GVK(Group/APIVersion/Kind)

cwen0:
### What problem does this PR solve?

image

`Client.Get` gets empty GVK(Group/APIVersion/Kind). 

### What is changed and how does it work?
 
Use `schema` and object to get GVK

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [x] Manual test (add detailed scripts or steps below)
- [ ] No code

image

PR #2218, opened 2021-08-12
refactor: experiments related APIs

g1eny0ung:

This PR refactors all the experiment-related APIs to enable the creation of chaos-related objects in the form of YAML definitions.

For a long time, our definitions of `apiserver` and `api/v1alpha1` were fragmented. We have repeatedly defined a new set of structures for the API to create and present chaos objects.

This is very bad, bad because **having two sets of structures simultaneously makes our subsequent code have to be compatible with two different formats. And when an error occurs, we have to think about both parts of the code to check it**.

So the main change in this PR is to reuse `api/v1alpha1` to create chaos objects in a uniform form.

After the changes, the biggest benefits are:

- When a new chaos type is introduced, you don't need to write any code for it again to adapt to `apiserver`
- The dashboard front-end can apply a set of logic to achieve code reuse, which will relieve a great burden (no need to be compatible with any other format)

Other changes:

- Move `docs/` to `pkg/swaggerdocs` to avoid ambiguity, this should be a package, but it is placed at the top level of the project
- Fix `MWHandleErrors `, now use `func SetAPIError(c *gin.Context, err *errorx.Error)` to correct the err code and the stack trace

  > For a long time, it was surprising that we didn't see that there was a problem with this middleware. The problem was that gin will covariate the errors received as `c.Error`, so whatever error you threw would end up showing `ErrUnknown`, and the stack trace was reset in the `nil` judgment.

- Refactor `/api/events`, discard the legacy `constructQueryArgs` and bump the `Filter` to `core.Filter`, this can be used in every `ListFilter` like methods

  > Due to the refactoring of the API, the associated `store` unit tests have been updated as well.

- `newWorkflowFrontendCodeGenerator` => `frontendGenerator`, add type constraint in generated ts file

### Checklist

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

This PR has many incompatible changes:

#### API

- `/api/events/get` => `/api/events/:id`
- `/archives/detail` => `/api/archives/:uid`


STRRL:
Hi @g1eny0ung, I noticed this PR is labeled with `needs-cheer-pick-2.0`, but I do not think we should release it in v2.0.1. That brings new features about the chaos dashboard, only for 2.1.0.

How do you think about it?

g1eny0ung:
> Hi @g1eny0ung, I noticed this PR is labeled with `needs-cheer-pick-2.0`, but I do not think we should release it in v2.0.1. That brings new features about the chaos dashboard, only for 2.1.0.
> 
> How do you think about it?

Oh sorry, I realized I misread that, I thought it was a 2.1 tag. I'll remove it now.

STRRL:
types in `ui/src/components/NewExperiment/types.ts` still need maintenance manually, do we have plan to use the code generator for this in the future? by chaos-builder or swagger.

g1eny0ung:
> types in `ui/src/components/NewExperiment/types.ts` still need maintenance manually, do we have plan to use the code generator for this in the future? by chaos-builder or swagger.

Plan to use swagger to replace it.

Ref doc: https://redux-toolkit.js.org/rtk-query/usage/code-generation.

STRRL:
> Plan to use swagger to replace it.

In this PR or next several PRs?

g1eny0ung:
> > Plan to use swagger to replace it.
> 
> In this PR or next several PRs?

Probably done in the new PR, as I don't have a more concrete plan yet.

STRRL:
I have a suggestion:

Could we move all the dashboard related packages into one new sub-directory `pkg/dashboard`?

some packages like `pkg/core`, `pkg/store` and `pkg/status` (and other packages) are only designed for chaos dashboard, putting it at `pkg` bring a mess.

g1eny0ung:
> I have a suggestion:
> 
> Could we move all the dashboard related packages into one new sub-directory `pkg/dashboard`?
> 
> some packages like `pkg/core`, `pkg/store` and `pkg/status` (and other packages) are only designed for chaos dashboard, putting it at `pkg` bring a mess.

Sounds great. I agree with u. Will move later.

g1eny0ung:
> > I have a suggestion:
> > Could we move all the dashboard related packages into one new sub-directory `pkg/dashboard`?
> > some packages like `pkg/core`, `pkg/store` and `pkg/status` (and other packages) are only designed for chaos dashboard, putting it at `pkg` bring a mess.
> 
> Sounds great. I agree with u. Will move later.

@STRRL I realize that if we make this change in this PR, it may cause a lot of difficulties for the review. 🤔 How about we do it after this PR? I'll open an issue to track it.

STRRL:
> > > I have a suggestion:
> > > Could we move all the dashboard related packages into one new sub-directory `pkg/dashboard`?
> > > some packages like `pkg/core`, `pkg/store` and `pkg/status` (and other packages) are only designed for chaos dashboard, putting it at `pkg` bring a mess.
> > 
> > 
> > Sounds great. I agree with u. Will move later.
> 
> @STRRL I realize that if we make this change in this PR, it may cause a lot of difficulties for the review.  How about we do it after this PR? I'll open an issue to track it.

Great! That's alright! Looking forward to that PR！

pkg/apiserver/utils/gin.go
```diff
@@ -0,0 +1,34 @@
+// Copyright 2020 Chaos Mesh Authors.
```
fewdan, line 1: s/2020/2021/g  ?

pkg/apiserver/archive/archive.go
```diff
@@ -137,106 +123,71 @@ func (s *Service) list(c *gin.Context) {
...
+			Created:   meta.StartTime.Format(time.RFC3339),
 		})
 	}
 
 	c.JSON(http.StatusOK, archives)
 }
 
-// @Summary Get the detail of an archived chaos experiment.
-// @Description Get the detail of an archived chaos experiment.
+// @Summary Get a archived chaos experiment.
```
fewdan, line 133: "an archived xxxx"

fewdan (commented):
just few typos, rest LGTM

STRRL (approved):
And here are a lot of codes working with reflect, like:

```go
reflect.ValueOf(chaos).Elem().FieldByName("APIVersion").String()
```

we could use type assert with `metav1.Object` and `schema.ObjectKind`, then use getter or setter. And this PR has been enough huge, I think I could help for improving that in the next PRs.

STRRL (commented):
And here are a lot of codes working with reflect, like:

```go
reflect.ValueOf(chaos).Elem().FieldByName("APIVersion").String()
```

we could use type assert with `metav1.Object` and `schema.ObjectKind`, then use getter or setter. And this PR has been enough huge, I think I could help for improving that in the next PRs.

STRRL (changes requested):
I could only use "request changes" for revoking the `LGT1` label

pkg/apiserver/archive/archive.go
g1eny0ung: Yep, this and the above are two typos. I'll fix it right away. 🍺

pkg/apiserver/schedule/schedule.go
```diff
@@ -25,618 +24,239 @@ import (
...
-		_ = c.Error(utils.ErrInvalidRequest.WrapWithNoMessage(err))
+		u.SetAPIError(c, u.ErrBadRequest.WrapWithNoMessage(err))
+
 		return
 	}
 
-	name := c.Query("name")
-	ns := c.Query("namespace")
+	var sch v1alpha1.Schedule
+	u.ShouldBindBodyWithJSON(c, &sch)
```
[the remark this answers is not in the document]
g1eny0ung: Good catch! I forgot that return is required here. I will fix it right away.

g1eny0ung:
@fewdan @STRRL Fix in b154cfe. PTAL

pkg/apiserver/experiment/experiment.go
```diff
@@ -938,381 +159,437 @@ func (s *Service) listExperiments(c *gin.Context) {
...
+				Name:      reflect.ValueOf(chaos).Elem().FieldByName("Name").String(),
+				Kind:      getChaosResult.FieldByName("Kind").String(),
+				UID:       getChaosResult.FieldByName("UID").String(),
+				Created:   getChaosResult.FieldByName("StartTime").Interface().(time.Time).Format(time.RFC3339),
+			},
+			Status: status.GetChaosStatus(chaos.(v1alpha1.InnerObject)),
+		},
+		KubeObject: core.KubeObjectDesc{
+			TypeMeta: metav1.TypeMeta{
+				APIVersion: v1alpha1.GroupVersion.String(),
```
g1eny0ung, line 279: > And here are a lot of codes working with reflect, like:
> 
> ```go
> reflect.ValueOf(chaos).Elem().FieldByName("APIVersion").String()
> ```
> 
> we could use type assert with `metav1.Object` and `schema.ObjectKind`, then use getter or setter. And this PR has been enough huge, I think I could help for improving that in the next PRs.

@STRRL I currently hardcoded the `APIVersion` this way, but I don't know if that's a good idea since it's not dynamically fetched, but I guess even if it's not dynamic, there shouldn't be an error here? Because if we change the version in the future, `v1alpha1.GroupVersion` will be updated as well.

STRRL: let me dig into the kubernetes machinery to findout 🤔

g1eny0ung: I consulted while solving this problem: https://github.com/kubernetes/kubernetes/pull/63972. Hope it helps.

STRRL: I think we could just use:

```go
			TypeMeta: metav1.TypeMeta{
				APIVersion: chaos.GetObjectKind().GroupVersionKind().GroupVersion().String(),
				Kind:       chaos.GetObjectKind().GroupVersionKind().Kind,
			},
```

STRRL: because `runtime.Object` contains the method:

```go
type Object interface {
	GetObjectKind() schema.ObjectKind
	DeepCopyObject() Object
}
```

and `json.Unmarshal` could resolve fields: `APIVersion` and `Kind` properly into `runtime.TypeMeta`

STRRL:
this PR has conflicts with https://github.com/chaos-mesh/chaos-mesh/pull/2218, I think I could help to resolve these conflicts.

pkg/apiserver/experiment/experiment.go
g1eny0ung: I thought so too, but you can check my PR link above and the related issues, and while I didn't dig into the details, it looks like GVK is not serialized out in the k8s API because it would omit a lot of duplicate GVK fields?

g1eny0ung: Another way is using https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/scheme to retrieve the GVK dynamically. But I think if we can use the existing method to do the GVK retrieving, we can do it without introducing new dependencies.

g1eny0ung: I looked at #2217 and it looks like we can only use the second method to get the GVK currently after it is merged.

g1eny0ung:
> this PR has conflicts with #2217, I think I could help to resolve these conflicts.

Already done but leave the GVK problem described above.

pkg/apiserver/experiment/experiment.go
STRRL: > I thought so too, but you can check my PR link above and the related issues, and while I didn't dig into the details, it looks like GVK is not serialized out in the k8s API because it would omit a lot of duplicate GVK fields?

I find one related PR: https://github.com/chaos-mesh/chaos-mesh/pull/1368.

I only tried `chaos.GetObjectKind().GroupVersionKind().GroupVersion().String()` in the `archive.go`, which definitely have `APIVersion` and `Kind` in the serialized json.

Let me take a look at other files like `experiment.go` :D

g1eny0ung: Ok. 👌🏽

STRRL: @g1eny0ung, in PR #1368, we use `apiutil.GVKForObject()` to fetch the GVK each time. 

I think we could also use `apiutil.GVKForObject()` as a temporary solution in this PR.

STRRL: > I only tried chaos.GetObjectKind().GroupVersionKind().GroupVersion().String() in the archive.go, which definitely have APIVersion and Kind in the serialized json.
> 
> Let me take a look at other files like experiment.go :D

fetching GVK with reflection DOES NOT WORK in `experiment.go` and `schedule.go`, because the GVK info could only referenced by `runtime.Schema` in kubernetes apimachinary.


> @g1eny0ung, in PR #1368, we use apiutil.GVKForObject() to fetch the GVK each time.
> 
> I think we could also use apiutil.GVKForObject() as a temporary solution in this PR.

Maybe we could make the `v1alpha1.AllKinds()` better: filling the `TypeMeta` of each Chaos when `register` the `ChaosKind`. But this operation needs a `runtime.Schema`, and we have not provide one in package `api/v1alpha1`. And I think provide `runtime.Schema` in `api/v1alpha1` is not good. :/

I will raise an issue about that for discussing the next step of refactoring.

g1eny0ung: > I think we could also use apiutil.GVKForObject() as a temporary solution in this PR.

I think so too, and this could be a temporary solution. Looking forward to making this part easier in the future.

g1eny0ung: Already done.

g1eny0ung:
CI failing because: https://github.com/kubernetes-sigs/kubebuilder/issues/2311.

STRRL:
hi @g1eny0ung , please resolve the conflicts.

STRRL:
Hi @g1eny0ung, I meet one issue when I trying to create a `Schedule` of `PodChaos`, I think it might be a bug:

photo_2021-08-25_14-26-44

When I click the button `Submit`, nothing happens, no logs on console, no request on network.

g1eny0ung:
> Hi @g1eny0ung, I meet one issue when I trying to create a `Schedule` of `PodChaos`, I think it might be a bug:
> 
> photo_2021-08-25_14-26-44
> 
> When I click the button `Submit`, nothing happens, no logs on console, no request on network.

@STRRL Thx for finding this bug. It's a validation problem and should be fixed now.

g1eny0ung:
@STRRL Thx for finding this bug. It's a validation problem and should be fixed now.

pkg/core/common.go
```diff
@@ -14,20 +14,92 @@
...
-	Name        string            `json:"name"`
 	Namespace   string            `json:"namespace"`
+	Name        string            `json:"name"`
 	Labels      map[string]string `json:"labels,omitempty"`
 	Annotations map[string]string `json:"annotations,omitempty"`
 }
+
+type Filter struct {
+	ObjectID  string `json:"object_id"`
+	Start     string
```
WangXiangUSTC, line 48: how about adding ``json:"-"`` for `Start`, `End` and `Limit`?

pkg/apiserver/experiment/experiment.go
```diff
@@ -27,1290 +26,579 @@ import (
...
-		c.Status(http.StatusInternalServerError)
-		_ = c.Error(utils.ErrInternalServer.WrapWithNoMessage(err))
+	if err = patchExperiment(kubeCli, exp, annotations); err != nil {
+		u.SetAPImachineryError(c, err)
+
 		return
 	}
 
-	c.JSON(http.StatusOK, StatusResponse{Status: "success"})
+	c.JSON(http.StatusOK, u.Response{Status: "success"})
```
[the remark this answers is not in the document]
g1eny0ung: Got it! Already replaced. 🍺

pkg/core/common.go
g1eny0ung: This will always omit the fields, but we need them.

pkg/apiserver/experiment/experiment.go
```diff
@@ -27,1290 +26,579 @@ import (
...
-				Namespace:   chaos.Namespace,
-				Labels:      chaos.Labels,
-				Annotations: chaos.Annotations,
-			},
-			Spec: chaos.Spec,
-		},
-	}, nil
-}
+	if chaosKind, ok := v1alpha1.AllKinds()[kind]; ok {
+		reflect.ValueOf(chaosKind.Chaos).Elem().FieldByName("ObjectMeta").Set(reflect.ValueOf(metav1.ObjectMeta{}))
```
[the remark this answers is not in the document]
g1eny0ung: Because the previous chaos' `ObjectMeta` is still in `chaosKind.Chaos`, this needs to be cleared to prevent errors from occurring (The key thing is to remove the `ResourceVersion`).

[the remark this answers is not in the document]
STRRL: using DeepCopy could prevent this “previous object“ issue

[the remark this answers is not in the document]
STRRL: Actually, each usage of `v1alpha1.AllKind()` should be wrapped with `DeepCopyObject`. I think we should provide the easy-to-use and more obviously method name.

ref: https://github.com/chaos-mesh/chaos-mesh/pull/2193#discussion_r690915555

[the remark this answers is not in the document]
STRRL: or call `DeepCopyObject` or `DeepCopyList` inside of `clone()`, but it is not so "obvious".

WangXiangUSTC (approved):
LGTM

PR #2242, opened 2021-08-20
use `host-sys` in fusedev device priviledge granting

YangKeao:

As the `chaos-daemon` tried to write `device.allow` in cgroup, but used the `/sys/fs/cgroup` as the root path, it will give an error. The root path of host cgroup should be `/host-sys/fs/cgroup`.

Modify `sys` to `host-sys`.


cwen0 (approved):
LGTM

fewdan (approved):
LGTM

PR #2245, opened 2021-08-20
fix(StressChaos): containerNames problem

g1eny0ung:

Unify the `containerNames` field on front-end and back-end.

Proposal: [xxx](url) 

What's Changed:

### Related changes

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```release-note
Please add a release note.
```

### DIFF
--- a/pkg/apiserver/experiment/experiment.go
+++ b/pkg/apiserver/experiment/experiment.go
@@ -373,14 +373,11 @@ func (s *Service) createStressChaos(exp *core.ExperimentInfo, kubeCli client.Cli
 					Mode:     v1alpha1.PodMode(exp.Scope.Mode),
 					Value:    exp.Scope.Value,
 				},
+				ContainerNames: exp.Target.StressChaos.ContainerNames,
 			},
 			Stressors:         stressors,
 			StressngStressors: exp.Target.StressChaos.StressngStressors,
 		},
-	}
-
-	if exp.Target.StressChaos.ContainerName != nil {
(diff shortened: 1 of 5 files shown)


WangXiangUSTC (approved):
LGTM

pkg/apiserver/experiment/experiment.go
```diff
@@ -373,16 +373,13 @@ func (s *Service) createStressChaos(exp *core.ExperimentInfo, kubeCli client.Cli
 					Mode:     v1alpha1.PodMode(exp.Scope.Mode),
 					Value:    exp.Scope.Value,
 				},
+				ContainerNames: exp.Target.StressChaos.ContainerNames,
```
cwen0, line 376: Panic here if `exp.Target.StressChaos.CintainerName` is nil.

g1eny0ung: Will be an empty array because of the `omitempty` option?

cwen0: Got it

cwen0: you are right

Muscule:
When this fix will be available?

cwen0:
> When this fix will be available?

v2.0.1 included this fix. You can upgrade to v2.0.1 https://github.com/chaos-mesh/chaos-mesh/releases/tag/v2.0.1

PR #2289, opened 2021-09-09
chaos-daemon: fix nil pointer dereference in chaos-daemon

YangKeao:

When the `toda` exited unexpectedly, the process will dereference an `nil` error.

### DIFF
--- a/pkg/bpm/bpm.go
+++ b/pkg/bpm/bpm.go
@@ -148,8 +148,7 @@ func (m *BackgroundProcessManager) StartProcess(cmd *ManagedProcess) (*process.P
 	go func() {
 		err := cmd.Wait()
 		if err != nil {
-			err, ok := err.(*exec.ExitError)
-			if ok {
+			if err, ok := err.(*exec.ExitError); ok {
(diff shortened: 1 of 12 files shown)


pkg/bpm/bpm.go
```diff
@@ -148,8 +148,7 @@ func (m *BackgroundProcessManager) StartProcess(cmd *ManagedProcess) (*process.P
 	go func() {
 		err := cmd.Wait()
 		if err != nil {
-			err, ok := err.(*exec.ExitError)
-			if ok {
+			if err, ok := err.(*exec.ExitError); ok {
```
STRRL, line 151: I think we should change the name of `err` in `if err, ok := err.(*exec.ExitError); ok `, this re-assignment of `err` still affects the next `else` branch

STRRL: PTAL @YangKeao

YangKeao: Fixed.

STRRL (approved):
LGTM!

WangXiangUSTC (approved):
LGTM

Issue #2247, opened 2021-08-21
Support for multiple pod networks


Cerebus:
## Feature Request

Kubernetes can have multiple data planes connecting pods with multiple interfaces (multiple CNIs or instances of a CNI can run in parallel).   Each data plane connects to a pod as an additional interface (eth1 &etc).  It would be great to be able to apply network chaos on these networks.

WangXiangUSTC:
So maybe we need to add a config for the network interface, does this meet your requirement？

Cerebus:
Yes, as long as the experiment doesn't fail if the interface doesn't exist in a Pod.  The number of interfaces may vary across pods even in the same app.  

Configuring by qdisc by CIDR might be useful in addition (but not as a replacement).  Again, no failure if a selected Pod doesn't have an interface in that range.

Cerebus:
Oh, and it should work w/ containerd.  Litmus experiments in this area rely on pumba, but pumba + containerd it giving me heartache.

WangXiangUSTC:
Why pumba + containerd giving you heartache?

YangKeao:
> Why pumba + containerd giving you heartache?

Because `Pumba` is chaos testing tool for Docker :smile_cat:

YangKeao:
Are you using CNI-Genie to adapt multiple CNI plugin in the same kubernetes cluster?

Cerebus:
I'm actually using [meshnet-cni](https://github.com/networkop/meshnet-cni) to run routed network emulations b/c it only emulates point to point links, which is 90% of what I need, but the Topology CRD is a pain to work with.  I started to look at multus.  I'll check out cni-genie.

I *think* multus will actually annotate Pods with all the attachment data.

PR #2290, opened 2021-09-09
networkchaos: support select different interface in networkchaos

YangKeao:

Issue Number: close https://github.com/chaos-mesh/chaos-mesh/issues/2247

Problem Summary:

`NetworkChaos` doesn't support selecting different interfaces yet. All `NetworkChaos` will operate on interface `eth0`. However, with `Multus` or some other CNI, it's possible to have multiple different interfaces.

The `device` argument was added to the protobuf protocol, and the `tc` command/`iptables` command will use  corresponding arguments to add netem/iptables rules on the specific interface.

### Related changes

* PR to update `chaos-mesh/website`/`chaos-mesh/website-zh`: TODO
* Need to update Chaos Dashboard component, related issue: TODO
* Need to cheery-pick to the release branch: No

### Checklist

Tests

- [x] Manual test 

I created a cluster with `Multus`, and test the `NetworkChaos` on it.

### Release note 

```release-note
support select different interface in networkchaos
```


g1eny0ung:
Please add the `targetDevice` field to the docs. @YangKeao

g1eny0ung:
Also, it's better to add an example (with explanation) to tell people how to use it. :)

PR #2299, opened 2021-09-13
build only when dockerfile is modified

YangKeao:

The `dev-env` and `build-env` is modified too frequently! We should only upload them when the `Dockerfile` is modified.


fewdan (approved):
LGTM

STRRL (approved):
LGTM!

PR #2303, opened 2021-09-14
fix: persist event with UTC Time

g1eny0ung:

Problem Summary:

This PR persists event with UTC time (`CreatedAt` field). Using UTC time helps us to better filter data by time range without considering the time zone.

What's Changed:

#### main

- before persist and query events, parse time to UTC

#### other

- fix a condition typo in `ConstructQueryArgs`
- remove unused imports

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```release-note
Please add a release note.
```

### DIFF
--- a/pkg/dashboard/apiserver/event/event.go
+++ b/pkg/dashboard/apiserver/event/event.go
@@ -80,18 +80,8 @@ func (s *Service) list(c *gin.Context) {
 		log.V(1).Info("Replace query namespace with", ns)
 	}
 
-	start, err := time.Parse(time.RFC3339, c.Query("start"))
-	if err != nil {
-		u.SetAPIError(c, u.ErrBadRequest.WrapWithNoMessage(err))
-
-		return
-	}
-	end, err := time.Parse(time.RFC3339, c.Query("end"))
-	if err != nil {
-		u.SetAPIError(c, u.ErrBadRequest.WrapWithNoMessage(err))
-
-		return
-	}
+	start, _ := time.Parse(time.RFC3339, c.Query("start"))
+	end, _ := time.Parse(time.RFC3339, c.Query("end"))
 
 	filter := core.Filter{
 		ObjectID:  c.Query("object_id"),
--- a/pkg/dashboard/apiserver/event/event.go
+++ b/pkg/dashboard/apiserver/event/event.go
@@ -80,8 +80,18 @@ func (s *Service) list(c *gin.Context) {
 		log.V(1).Info("Replace query namespace with", ns)
 	}
 
-	start, _ := time.Parse(time.RFC3339, c.Query("start"))
-	end, _ := time.Parse(time.RFC3339, c.Query("end"))
+	start, err := time.Parse(time.RFC3339, c.Query("start"))
+	if err != nil {
+		u.SetAPIError(c, u.ErrBadRequest.WrapWithNoMessage(err))
+
+		return
+	}
+	end, err := time.Parse(time.RFC3339, c.Query("end"))
+	if err != nil {
+		u.SetAPIError(c, u.ErrBadRequest.WrapWithNoMessage(err))
+
+		return
+	}
 
 	filter := core.Filter{
 		ObjectID:  c.Query("object_id"),
--- a/pkg/dashboard/apiserver/event/event.go
+++ b/pkg/dashboard/apiserver/event/event.go
@@ -17,6 +17,7 @@ import (
 	"context"
 	"net/http"
 	"strconv"
+	"time"
 
 	"github.com/gin-gonic/gin"
 	"github.com/jinzhu/gorm"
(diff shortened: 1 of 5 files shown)


pkg/dashboard/apiserver/event/event.go
```diff
@@ -77,10 +80,13 @@ func (s *Service) list(c *gin.Context) {
 		log.V(1).Info("Replace query namespace with", ns)
 	}
 
+	start, _ := time.Parse(time.RFC3339, c.Query("start"))
+	end, _ := time.Parse(time.RFC3339, c.Query("end"))
```
WangXiangUSTC, line 84: I think it's better to judge the error, otherwise, an invalid request can make the dashboard panic.

pkg/dashboard/core/common.go
```diff
@@ -62,6 +62,8 @@ func (f *Filter) toMap() map[string]interface{} {
 	return fMap
 }
 
+const zeroTime = "0001-01-01 00:00:00"
```
WangXiangUSTC, line 65: do we need to use this to initialize `Start` and `End` in `Filter`?

pkg/dashboard/apiserver/event/event.go
g1eny0ung: Agreed! I will update soon.

pkg/dashboard/core/common.go
```diff
@@ -82,19 +84,19 @@ func (f *Filter) ConstructQueryArgs() (string, []interface{}) {
...
 		args = append(args, f.Start)
-	} else if f.Start == "" && f.End != "" {
+	} else if f.Start == zeroTime && f.End != zeroTime {
 		startEnd = "created_at <= ?"
 		args = append(args, f.End)
 	}
 
 	if startEnd != "" {
-		if len(args) > 0 {
+		if len(query) > 0 {
```
fewdan, line 99: I think there is no need to modify it here? That is to say the same as line 76.

pkg/dashboard/core/common.go
g1eny0ung: I think it's no need to initialize them in the `Filter` struct because if we don't pass start or end, the `time.Parse` will always return zero time.

pkg/dashboard/core/common.go
g1eny0ung: This is because if we only pass start or end, the `args` with bigger than zero. So the query result will be like `AND created_at >= xxx`, this will generate an incorrect statement that will cause the SQL query to fail.

g1eny0ung: > I think there is no need to modify it here? That is to say the same as line 76.

The previous is fine because we append `query` and `args` at the same time but at the `startEnd` we have pre-append `args`. This causes us to need to determine whether the `query` is empty or not.

pkg/dashboard/apiserver/event/event.go
g1eny0ung: Done in ddd08ef.

g1eny0ung: It looks like parse will always return zero time if the second param is incorrect. 🔮 I will reset the previous commit.

pkg/dashboard/core/common.go
fewdan: okay

WangXiangUSTC (approved):
LGTM

YangKeao (approved):
LGTM

STRRL:
Should we use UTC time in all the persistent entities (Schedule, Workflow, Experiment)? 

And I do not get the point of the original issue, sqlite would store the timezone info, and `/api/events` always accept parameters `start/end` with format RFC3339.  🤔🤔🤔🤔

g1eny0ung:
@STRRL

> Should we use UTC time in all the persistent entities (Schedule, Workflow, Experiment)?

If we need to query by time range, here's what I suggest, depositing UTC time eliminates the need to account for time zone effects.

> And I do not get the point of the original issue, sqlite would store the timezone info, and `/api/events` always accept parameters `start/end` with format RFC3339. It seems every things should works. 🤔🤔🤔🤔

The first question is related to the above said.

For example:

If now there is a cluster whose local time is different from your browser's local time. Then if you want to query in the correct time range, you need the code on your browser to convert according to the cluster's time zone, which requires an extra step, **the browser needs to know what time zone it is first**.

But with UTC time storage, the browser code just needs to convert any local time to UTC.

Regarding the second question, I didn't quite understand what you meant. 🤔 We need a time format for unifying the user input internally. The `RFC3339` can be well identified `ISO 8601` standard. We can use it to receive input with time zone information very well, then we convert the input to UTC.

STRRL:
So it's an enhancemnet or a bugfix? 🙈

g1eny0ung:
I think it's more of a bugfix? Without this PR, querying events with time range will only succeed when the local and cluster are in the same time zone.

STRRL:
> Without this PR, querying events with time range will only succeed when the local and cluster are in the same time zone.

I think the root cause is using `string` but not `time.Time` in `core.Filter`, and losing timezone in somewhere. Using UTC time everywhere also works, LGTM. :)

PR #2318, opened 2021-09-17
Makefile,ci: Build arm image for `build-env` and `dev-env` and upload to `ghcr.io`

YangKeao:

Build arm image for `build-env` and `dev-env`. 

Problem Summary:

I have encountered these problem while trying to build arm image and upload:

1. Kustomize don't have arm image in an older version.
2. `docker manifest` and `docker buildx` doesn't fix in the matrix construct model.

### Solution

1. Kustomize don't have arm image in an older version.

I update the `kustomize` to a higher version, which releases with an arm binary.

2. `docker manifest` and `docker buildx` doesn't fix in the matrix construct model.

The suggested way to use `docker buildx build` is to tell all platforms to it, which means I should pass `--push --platform linux/arm64,linux/amd64`. This method is fine for `build-env` and `dev-env`, but will fail for other images, because we build the binary at the outside of `Dockerfile`.

So this PR uses another way: build an arm image with `:tag-arm64` and an amd64 image with `:tag-amd64`, then reference them from a manifest.


.github/workflows/upload_env_image.yml
```diff
@@ -0,0 +1,99 @@
...
+    strategy:
+      matrix:
+        arch: [amd64, arm64]
+        image: [dev, build]
+    outputs:
+      image_tag: ${{ steps.image_tag.outputs.image_tag }}
+    steps:
+      - uses: actions/checkout@master
+        with:
+          # Must use at least depth 2!
```
STRRL, line 24: why?

STRRL:
It seems this PR should trigger `Upload Env Images`, but it did not.

.github/workflows/upload_env_image.yml
YangKeao: I don't know. It's copied from other yaml in this probject. Maybe it will affect the behavior of `GITHUB_REF`.

I will remove this line and try whether it works well.

YangKeao:
> It seems this PR should trigger `Upload Env Images`, but it did not.

The CI was triggered by label. If the label is not set, it will skip. I have opened another issue https://github.com/chaos-mesh/chaos-mesh/issues/2322 to use a better way to judge whether we should skip.

This time, I will add the label and rerun the CI.

STRRL:
I mean, these workflows defined in `upload_build_env_image.yml` and `upload_dev_env_image.yml` (not in `ci.yml`) should also be triggered within this PR, because of some changes in dockerfiles 🤔

YangKeao:
> I mean, these workflows defined in `upload_build_env_image.yml` and `upload_dev_env_image.yml` (not in `ci.yml`) should also be triggered within this PR, because of some changes in dockerfiles thinking

They will only be triggered on master branch, as it will upload the image to the `ghcr.io`.

.github/workflows/upload_env_image.yml
YangKeao: Fixed.

STRRL (approved):
LGTM!

iguoyr:
https://github.com/chaos-mesh/chaos-mesh/blob/a2c9ba138e5cb27506002ae631f1038332500c01/images/dev-env/Dockerfile#L15 

need to replace with arm version.

images/build-env/Dockerfile
```diff
@@ -20,8 +20,11 @@ RUN npm install yarn -g
 
 RUN if [[ -n "$HTTP_PROXY" ]]; then yarn config set proxy $HTTP_PROXY; fi
 
-ARG TARGET_PLATFORM=amd64
-RUN curl https://dl.google.com/go/go1.16.7.linux-$TARGET_PLATFORM.tar.gz | tar -xz -C /usr/local
+RUN echo -n $(uname -m) > /arch
+RUN if [ $(uname -m) = aarch64 ]; then echo -n arm64 > /arch; fi
+RUN if [ $(uname -m) = x86_64 ]; then echo -n amd64 > /arch; fi
+
+RUN curl -vvv https://dl.google.com/go/go1.16.7.linux-$(cat /arch).tar.gz | tar -xz -C /usr/local
```
iguoyr, line 27: It would be better to use a local variable.

YangKeao:
> https://github.com/chaos-mesh/chaos-mesh/blob/a2c9ba138e5cb27506002ae631f1038332500c01/images/dev-env/Dockerfile#L15
> 
> need to replace with arm version.

Fixed.

iguoyr (approved):
LGTM!

PR #2335, opened 2021-09-23
Dockerfile: use kubebuilder-tools but not kubebuilder

YangKeao:

Download the `kubebuilder-tools` but not `kubebuilder`. The chaos mesh doesn't depend on `kubebuilder` now, so I removed `kubebuilder` from the `Dockerfile`, and downloaded the `kubebuilder-tools` (which includes `etcd`, `apiserver` and `kubectl`) for the envtest.

The very first motivation of this PR is that, the older version of `kubebuilder` arm64 tar package doesn't include `etcd`, `apiserver` and `kubectl` :cry: .

The url of tools is found in https://github.com/kubernetes-sigs/controller-runtime/blob/v0.7.2/hack/setup-envtest.sh#L87

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -1,21 +1,20 @@
-# Project Maintainers
-
 This file lists the maintainers and committers of the Chaos Mesh project.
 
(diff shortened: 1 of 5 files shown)


iguoyr (approved):
LGTM!

cwen0 (approved):
LGTM

PR #2336, opened 2021-09-23
Better way to gen command.

Andrewmatilde:
Signed-off-by: andrewmatilde <davis6813585853062@outlook.com>

Give a better way to Gen command instead of many  fmt.sprintf and empty judgments.

I am a contributor of chaos-mesh/chaosd and chaosd use chaosmesh/pkg/chaosdeamon/iptables_server as a lib to complete network attack.
I found it is hard to use , add extension, debug or read code of iptables_server ,because there are both many codes wrote to generate iptables and many logical codes to use iptables to complete to task like drop web packet ...
I think iptables_server may designed for just using it in chaos-mesh , but it is hard to get a further use , so I try to give a better way to generate iptables commands first .

What's Changed:
Nothing changed except adding a cmmand generator package in utils.
If we use it to refactor iptables_server, it will reduce many code in generate codes and may make it easier to be understand.

Example (Before):
```
	protocolAndPort := ""
	if len(chain.Protocol) > 0 {
		protocolAndPort += fmt.Sprintf("--protocol %s", chain.Protocol)

		if len(chain.SourcePorts) > 0 {
			if strings.Contains(chain.SourcePorts, ",") {
				protocolAndPort += fmt.Sprintf(" -m multiport --source-ports %s", chain.SourcePorts)
			} else {
				protocolAndPort += fmt.Sprintf(" --source-port %s", chain.SourcePorts)
			}
		}

		if len(chain.DestinationPorts) > 0 {
			if strings.Contains(chain.DestinationPorts, ",") {
				protocolAndPort += fmt.Sprintf(" -m multiport --destination-ports %s", chain.DestinationPorts)
			} else {
				protocolAndPort += fmt.Sprintf(" --destination-port %s", chain.DestinationPorts)
			}
		}

		if len(chain.TcpFlags) > 0 {
			protocolAndPort += fmt.Sprintf(" --tcp-flags %s", chain.TcpFlags)
		}
	}

	rules := []string{}

	if len(chain.Ipsets) == 0 {
		rules = append(rules, strings.TrimSpace(fmt.Sprintf("-A %s -j %s -w 5 %s", chain.Name, chain.Target, protocolAndPort)))
	}

	for _, ipset := range chain.Ipsets {
		rules = append(rules, strings.TrimSpace(fmt.Sprintf("-A %s -m set --match-set %s %s -j %s -w 5 %s",
			chain.Name, ipset, matchPart, chain.Target, protocolAndPort)))
	}
	err := iptables.createNewChain(&iptablesChain{
		Name:  chain.Name,
		Rules: rules,
	})
	if err != nil {
		return err
	}
```
Example(After):
```
type Iptables struct {
	Exec           `exec:"iptables"`
	Tables         string `para:"-t"`
	Command        string `para:""`
	Chain          string `para:""`
	JumpTarget     string `para:"-j"`
	Protocol       string `para:"--protocol"`
	MatchExtension string `para:"-m"`
	SPorts         string `para:"--source-ports"`
	DPorts         string `para:"--destination-ports"`
	SPort          string `para:"--source-port"`
	DPort          string `para:"--destination-port"`
	TcpFlags       string `para:"--tcp-flags"`
}

func TestUnmarshalExample(t *testing.T) {
	n := Iptables{
		Exec:           NewExec(),
		Command:        "-A",
		Chain:          "Chaos_Chain",
		JumpTarget:     "Chaos_Target",
		Protocol:       "tcp",
		MatchExtension: "multiport",
		SPorts:         "2021,2022",
		TcpFlags:       "SYN",
	}
	path, args := Unmarshal(n)
	assert.Equal(t, "iptables -A Chaos_Chain -j Chaos_Target --protocol tcp -m multiport --source-ports 2021,2022 --tcp-flags SYN",
		path+" "+strings.Join(args, " "))
}
```
How it works:

### Checklist

Tests

- [ ] Unit test

Side effects

### DIFF
--- a/pkg/utils/command.go
+++ b/pkg/utils/command.go
@@ -1,0 +1,84 @@
+package utils
+
+import (
+	"os/exec"
+	"reflect"
+	"strings"
+)
+
+const ExecTag = "exec"
+const SubCommandTag = "sub_command"
+const ParaTag = "para"
+
+type Exec struct {
+	option string
+}
+
+func NewExec() Exec {
+	return Exec{option: "OK"}
+}
+
+func ToCommand(i interface{}) *exec.Cmd {
+	path, args := Unmarshal(i)
+	return exec.Command(path, args...)
+}
+
+func Unmarshal(i interface{}) (string, []string) {
+	value := reflect.ValueOf(i)
+	return unmarshal(value)
+}
+
+func unmarshal(value reflect.Value) (string, []string) {
+	//var options []string
+	if path, ok := SearchKey(value); ok {
+		// Field(0).String is Exec.Path
+
+		if path == "" {
+			return "", nil
+		}
+		args := make([]string, 0)
+		for i := 0; i < value.NumField(); i++ {
+			if _, ok := value.Type().Field(i).Tag.Lookup(SubCommandTag); ok {
+				subPath, subArgs := unmarshal(value.Field(i))
+				if subPath != "" {
+					args = append(args, subPath)
+				}
+				args = append(args, subArgs...)
+			}
+			if paraName, ok := value.Type().Field(i).Tag.Lookup(ParaTag); ok {
+				if value.Type().Field(i).Type.Name() == "string" {
+					if value.Field(i).String() != "" {
+						if paraName != "" {
+							args = append(args, paraName)
+						}
+						args = append(args, value.Field(i).String())
+					}
+				}
+				if value.Field(i).Kind() == reflect.Slice {
+					if slicePara, ok := value.Field(i).Interface().([]string); ok {
+						if strings.Join(slicePara, "") != "" {
+							args = append(args, paraName)
+							args = append(args, slicePara...)
+						}
+
+					}
+				}
+			}
+		}
+		return path, args
+	} else {
+		return "", nil
+	}
+}
+
+func SearchKey(value reflect.Value) (string, bool) {
+	for i := 0; i < value.NumField(); i++ {
+		if path, ok := value.Type().Field(i).Tag.Lookup(ExecTag); ok {
+			if value.Field(i).Field(0).String() == "" {
+				return "", false
+			}
+			return path, ok
+		}
+	}
+	return "", false
+}
--- a/pkg/utils/command_test.go
+++ b/pkg/utils/command_test.go
@@ -1,0 +1,74 @@
+package utils
+
+import (
+	"github.com/stretchr/testify/assert"
+	"strings"
+	"testing"
+)
+
+type IptablesTest struct {
+	Exec   `exec:"iptables"`
+	Port   string   `para:"-p"`
+	Ports  []string `para:"--ports"`
+	EPort  string   `para:"-ep"`
+	EPorts []string `para:"--e_ports"`
+	Match  `sub_command:""`
+	Match_ `sub_command:""`
+}
+
+type Match struct {
+	Exec   `exec:"-m"`
+	Helper string `para:"--helper"`
+}
+
+type Match_ struct {
+	Exec   `exec:"-m"`
+	Helper string `para:"--helper"`
+}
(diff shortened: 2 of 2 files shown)


Andrewmatilde:
PTAL, @cwen0 @WangXiangUSTC

pkg/utils/command.go
```diff
@@ -0,0 +1,123 @@
+// Copyright 2021 Chaos Mesh Authors.
```
cwen0, line 1: How about using a single pkg instead of `utils`?

pkg/utils/command.go
```diff
@@ -0,0 +1,123 @@
...
+// If the tag value like "-p" is empty string ,
+// the para will just add the field value into the command just as some single value parameter in command.
+// If the value of field is empty string or empty string slice or empty slice, the field and tag will all be skipped.
+const ParaTag = "para"
+
+type Exec struct {
+	option string
+}
+
+func NewExec() Exec {
```
cwen0, line 55: Add some comments to describe when to use it and why need to use it.

pkg/utils/command.go
```diff
@@ -0,0 +1,123 @@
...
+// 	Exec
+//	Port string `para:"-p"`
+// }
+// If the field is not string type or []string type , it will be skipped.
+// If the tag value like "-p" is empty string ,
+// the para will just add the field value into the command just as some single value parameter in command.
+// If the value of field is empty string or empty string slice or empty slice, the field and tag will all be skipped.
+const ParaTag = "para"
+
+type Exec struct {
```
cwen0, line 51: Can we just use an empty struct instead of it?

Andrewmatilde: I use option string to stand for exec existing or not existing here.

pkg/utils/command.go
```diff
@@ -0,0 +1,123 @@
...
+
+package utils
+
+import (
+	"os/exec"
+	"reflect"
+	"strings"
+)
+
+// ExecTag stand for the path of executable file in command.
```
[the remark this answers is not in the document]
Andrewmatilde: Fixed.

pkg/utils/command.go
```diff
@@ -0,0 +1,123 @@
...
+// we must add Exec in the struct and use NewExec() to initialize it,
+// because the default way to initialize Exec means None in code.
+const ExecTag = "exec"
+
+// SubCommandTag stand for the sub command in common command.
+// We can use it in struct fields as a tag.
+// Just like MatchExtension below
+// type Iptables Struct {
+// 	Exec
+//	MatchExtension string `sub_command:""`
```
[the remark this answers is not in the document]
Andrewmatilde: Fixed.

pkg/utils/command.go
Andrewmatilde: Fixed.

pkg/utils/command.go
Andrewmatilde: Fixed.

cwen0 (approved):
LGTM

pkg/command/command.go
```diff
@@ -0,0 +1,136 @@
...
+
+// Exec is the interface of a command.
+// We need to inherit it in the struct of command.
+// User must add ExecTag as the tag of Exec field.
+// Example:
+// type Iptables struct {
+//	Exec           `exec:"iptables"`
+//	Tables         string `para:"-t"`
+// }
+type Exec struct {
```
YangKeao, line 64: It seems that the only function of `Exec` is to set `exec` tag? Then why it has an option field (and the field is verified to be not empty in `SearchKey`)?

Andrewmatilde: In the case of sub command , we need to know if the sub command is used.
Such as the Match subcommand in IptablesTest below.

PR #2366, opened 2021-10-09
JVMChaos: refine with byteman

### LINKED ISSUE #2365: refine JVMChaos with byteman
## Feature Request

Now JVMChaos implement with [chaosblade-exec-jvm](https://github.com/chaosblade-io/chaosblade-exec-jvm), which lack 
detailed documentation, and require sidecar, all together make it difficult for users to use.

refine it with byteman, which has a detailed document https://downloads.jboss.org/byteman/4.0.14/byteman-programmers-guide.html, and no need to use sidecar.

WangXiangUSTC:

Issue Number: close #2365 

Problem Summary: Now JVMChaos implement with chaosblade-exec-jvm, which lack
detailed documentation, and require sidecar, all together make it difficult for users to use.

What's Changed:
1. refactor the structure of JVMChaos
2. install byteman in chaos-daemon

### Related changes

* PR to update `chaos-mesh/website`/`chaos-mesh/website-zh`: Yes. issue here: https://github.com/chaos-mesh/website/issues/162

### Checklist

Tests

- [x] Unit test
- [x] E2E test
- [ ] No code

Side effects

### Release note 

```release-note
Please add a release note.
```


WangXiangUSTC:
@chaos-mesh/maintainers @chaos-mesh/committers PTAL

WangXiangUSTC:
> > Is it possible that use the application's JVM instead of preparing JVM in the chaos-daemon? 🤔
> 
> Yeah, after discussion with @STRRL, we found that there was a problem in using Java in chaos-daemon: For example, The Java version in chaos-daemon was different from that in the target container, so attach might fail. So I agree with this.

I tried to execute Byteman using the environment(Java) in the target container, but ran into a problem that byteman's script would use `/bin/bash`, but the target container might not contain it. There may be additional problems due to the environment of the target container.

In addition, I have tested using Java 16.0.1 to attach version 1.8, and using Oracle JDK 16 to attach OpenJDK 14, and all without any problem. So I think that probably in most cases different versions of Java will not cause attach Agent to fail

How about keeping the java in chaos-daemon, and still use it? @STRRL

api/v1alpha1/jvmchaos_webhook.go
```diff
@@ -17,560 +17,67 @@ package v1alpha1
...
-		var value = ""
-		var exist = false
-		if parameters != nil {
-			value, exist = parameters[rule.Name]
-		}
-		if rule.Required && !exist {
-			errorMsg := fmt.Sprintf("with %s: %s, %s: %s", target, in.Target, action, in.Action)
-			allErrs = append(allErrs, field.Required(innerField, errorMsg))
+	switch in.Action {
+	case JVMStressAction:
```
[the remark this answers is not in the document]
WangXiangUSTC: OK, I will update it in Chaosd, and then update it in Chaos Mesh in another pr

STRRL:
> > > Is it possible that use the application's JVM instead of preparing JVM in the chaos-daemon? thinking
> > 
> > 
> > Yeah, after discussion with @STRRL, we found that there was a problem in using Java in chaos-daemon: For example, The Java version in chaos-daemon was different from that in the target container, so attach might fail. So I agree with this.
> 
> I tried to execute Byteman using the environment(Java) in the target container, but ran into a problem that byteman's script would use `/bin/bash`, but the target container might not contain it. There may be additional problems due to the environment of the target container.
> 
> In addition, I have tested using Java 16.0.1 to attach version 1.8, and using Oracle JDK 16 to attach OpenJDK 14, and all without any problem. So I think that probably in most cases different versions of Java will not cause attach Agent to fail
> 
> How about keeping the java in chaos-daemon, and still use it? @STRRL

It's OK to me. :D

pkg/chaosdaemon/pb/chaosdaemon.proto
```diff
@@ -240,3 +244,17 @@ message SetDNSServerRequest {
   bool enable = 3;
   bool enterNS = 4;
 }
+
+message InstallJVMRulesRequest {
+  string container_id = 1;
+  string rule = 2;
+  int32 port = 3;
+  bool enterNS = 4;
```
STRRL, line 252: I noticed that we always set `enterNs` to `true`. And `jvm_server.go` only works with `enterNs` being `ture`. It there any other scenarios using `false`? Or remove this field. Same situation in UninstallJVMRulesRequest.

STRRL (commented):
The related examples should be updated or purged.

rest LTGM!

pkg/chaosdaemon/pb/chaosdaemon.proto
WangXiangUSTC: In chaosd, there are some similar codes. I'm planning to reuse the code in Chaos Mesh, and then need to set `enterNS` to false.

WangXiangUSTC:
@STRRL updated the examples, PTAL again

STRRL (approved):
LGTM!

PR #2380, opened 2021-10-12
Use SHA256 rather than SHA1

anuragpaliwal80:
As SHA1 is insecure, let's use SHA256

Issue Number: close #xxx 

Problem Summary:

Proposal: [xxx](url) 

What's Changed:

### Related changes

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```release-note
Please add a release note.
```

### DIFF
--- a/controllers/podnetworkchaos/netutils/len_test.go
+++ b/controllers/podnetworkchaos/netutils/len_test.go
@@ -35,6 +35,6 @@ func Test_compressName(t *testing.T) {
 
 		name = CompressName(name, 13, "test")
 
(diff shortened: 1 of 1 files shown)


anuragpaliwal80:
Thanks for pointing this out. My question is why this test is failing as the change is not impacting this test ?

WangXiangUSTC:
> Thanks for pointing this out. My question is why this test is failing as the change is not impacting this test ?

I think it's because that you change the `hash` function from `SHA1` to `SHA256`, so the `hashValue` changed

PR #2417, opened 2021-10-26
feat: add PhysicalMachineChaos UI

g1eny0ung:

What's Changed:

This PR allows the manipulation of `PhysicalMachineChaos` experiments, schedules, and workflows in the dashboard interface.

### Related changes

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```release-note
Please add a release note.
```


ui/src/components/NewExperimentNext/data/types.ts
```diff
@@ -534,14 +541,430 @@ const data: Record<Kind, Definition> = {
...
 
-const AwsChaosCommonSchema = Yup.object({
+const networkPhysicCommon: Spec = {
+  correlation: {
+    field: 'text',
+    label: 'Correlation',
+    value: '',
+    helperText: 'The correlation of corrupt',
+  },
+  device: {
```
iguoyr, line 551: @WangXiangUSTC `device` is a required field, it needs to be validated by webhook.

ui/src/components/NewExperimentNext/data/types.ts
```diff
@@ -534,14 +541,430 @@ const data: Record<Kind, Definition> = {
...
+  },
+}
+
+const diskPhysicCommon: Spec = {
+  size: {
+    field: 'text',
+    label: 'Size',
+    value: '',
+    helperText:
+      'The supported formats of the size are: c(=1), w(=2), kB(=1000), K(=1024), MB(=1024), M(=1024x1024), GB and so on.',
```
iguoyr, line 606: > MB(=1024)

0.0

ui/src/components/NewExperimentNext/data/types.ts
```diff
@@ -534,14 +541,430 @@ const data: Record<Kind, Definition> = {
...
+  },
+  path: {
+    field: 'text',
+    label: 'Path',
+    value: '',
+    helperText: 'Specify the file path of the data to be read/written',
```
iguoyr, line 612: @WangXiangUSTC Dashboard could only get events which is like `HTTP status is not ok` now, If I want to see more information about this error, I need to check logs on the physical machine. I wonder if we could return the error information to chaos-controller-manager, and then we can see more detailed events from dashboard.

WangXiangUSTC:
@iguoyr Thanks, since this doesn't have much to do with UI, I will update them in another pr.

iguoyr (approved):
LGTM

ui/src/components/NewExperimentNext/data/types.ts
g1eny0ung: Very messed here. 😵 Will update all mentioned descriptions in a new PR. Good catch!

PR #2429, opened 2021-10-28
refactor(apiserver): adapting /api/archives to the new utils

### LINKED ISSUE #2418: Swagger API Semantic error
## Bug Report

none

latest

**What did you do?**

Swagger API

Swagger API correct

**What did you see instead?**
Swagger API wrong

**Output of chaosctl**

...

image

https://github.com/chaos-mesh/chaos-mesh/blob/e79e27b81c03afe3e9d7a2ed690a0f5f1db41f5a/pkg/dashboard/apiserver/archive/archive.go#L299

https://github.com/chaos-mesh/chaos-mesh/blob/e79e27b81c03afe3e9d7a2ed690a0f5f1db41f5a/pkg/dashboard/apiserver/archive/archive.go#L476

`query` should be `path` .

g1eny0ung:

Issue Number: close #2418

Problem Summary:

Some swagger comments have errors.

What's Changed:

Except adapt `utils`, some other changes were introduced:

- Update `core.Filter` to parse time `Start` and `End` automatically. It means you needn't convert them to UTC time before using the filter.
- Replace `StartTime` and `FinishTime` with `CreatedAt` and `DeletedAt`.
- **Replace `Archived` with `DeletedAt`. We can borrow the power of gorm (soft delete) to simplify our logic.**

### Checklist

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```release-note
Please add a release note.
```

### DIFF
--- a/pkg/dashboard/collector/collector.go
+++ b/pkg/dashboard/collector/collector.go
@@ -85,7 +85,7 @@ func (r *ChaosCollector) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	// Ignore errors because logging is already done in the function
 	r.createOrUpdateExperiment(req, obj)
 
-	if obj.IsDeleted() && !isManaged {
+	if obj.IsDeleted() && isManaged {
 		if err = r.event.DeleteByUID(ctx, string(chaosMeta.GetUID())); err != nil {
 			r.Log.Error(err, "failed to delete experiment related events")
 		}


g1eny0ung:
> hi @g1eny0ung, how's going on about this PR? Is it ready for review? What could I do for help? :D

I'm fixing the bugs that happened because of the changes, and I think it'll be ready for review once all the tests are passed. 🔧

g1eny0ung:
> What could I do for help? :D

@STRRL The most important changes to this PR are in `collector.go`. I'm glad if you have free time to help focus a review of the code in this file

g1eny0ung:
> What could I do for help? :D

@STRRL The most important changes to this PR are in `collector.go`. I'm glad if you have free time to help focus a review of the code in this file.

STRRL:
> - Replace StartTime and FinishTime with CreatedAt and DeletedAt.
> - Replace Archived with DeletedAt. We can borrow the power of gorm (soft delete) to simplify our logic.

- IMO, using soft deletion might not be proper for this situation. I think soft deletion usually works for data auditing features for a system manager. I think using the soft deletion feature at this time might bring trouble when we need data auditing one day.
- As our common chaos CRDs have field `duration`, so the `FinishTime` and `DeleteAt` are NOT actually one thing. One experiment might already exceed the `duration` but it still exists on kube-apiserver. Then the user clicks `Archive` button on the dashboard. So the time the chaos recovered and the time the user clicks `Archive` button are obviously not same.

I prefer to add 2 fields with names like `FinishAt` and `ArchiveAt` (or other names).  Using `time.Time` for each type is great!

STRRL:
IMO, `CreateAt`(time), `UpdatedAt`(time), `DeletedAt`(time), (also `CreateBy`(user), `UpdatedBy`(user), `DeletedBy`(user)) are very common-to-used fields for data auditing in some software system. I think we should avoid these names if we do not actually develop some auditing features.

Actually, I still have concerns about using `CreateAt` as the start time. Because it's not the same in definition, `CreateAt` is usually used for the time created this database record, usually using `NOW()` function provided by the database(or current time provided by ORM framework). It is NOT the same as the time chaos affected(the time this object created in Kubernetes, by currently defined in our logic).  And currently we "consider" that the time creation the ChaoObject in kube-apiserver, is the affected time, but it's not extremely correct. Maybe one day we could control/report the accurate time the object was injected. :D

(So I also prefer to add `StartAt`. In other words:

- StartTime time.Time => StartAt time.Time
- FinsishTime time.Time => FinishAt time.Time
- Archived bool => ArchiveAt time.Time

As `CreateAt`, `UpdatedAt`, `DeletedAt`, only using them for database records' "metadata". And do not use Object's CreationTimestamp from kube-apiserver as the value of `CreateAt`, because they are actually two persistent systems.

g1eny0ung:
@STRRL Thx for the tips. After I finished reading, I think I can summarize it as follows:

> Such as `CreatedAt`, `DeleteAt` fields are database-sensitive. Chaos' creation and deletion times are not exactly the same as theirs.

I agree with u. But we use them for a long time and everywhere 😢. Regarding the changes in this PR, my initial intention was to **reduce unnecessary fields** and **reduce cognitive costs**.

We use an `Archived` field to distinguish whether an experiment is archived or not and delete it permanently if we don't want to record it. But with `gorm` we can do this simply by using soft delete. So I removed it.

For cognitive costs, I prefer https://gorm.io/docs/models.html#Conventions. This is the reason I don't wanna add too much more fields such as `StartAt` or `ArchiveAt`, etc. We can't mark every step of our thinking (With comments) on the code, so I chose to try to go with a generic approach.

**And I think it's just a question of how to think about it.** Just like, all archives, including schedules and workflows, are just a record, they are not created by the database, they are just saved in the database. When we choose to archive an experiment, **it does get deleted**, so we fill the delete field, it's independent of whether it's finished or not, and it doesn't have an archive time, it just gets deleted.

PR #2455, opened 2021-11-03
overwrite GORPOXY in the Makefile

Hexilee:

Problem Summary:

Now, the default `GOPROXY="https://proxy.golang.org,direct"` is not accessible from China mainland, and we also cannot pass through it by `HTTPS_PROXY`.

What's Changed:

Pass user-defined GORPOXY into build-env and dev-env.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [x] No code

Side effects

### Release note 

```release-note
Please add a release note.
```


YangKeao:
Why not keep it null (or off) by default?

Hexilee:
> Why not keep it null (or direct) by default?

Yes, now set it direct by default.

Makefile
```diff
@@ -23,14 +23,15 @@ IMAGE_DEV_ENV_BUILD ?= 0
...
-GOARCH := $(if $(GOARCH),$(GOARCH),"")
-GOENV  := GO15VENDOREXPERIMENT="1" CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH)
-CGOENV := GO15VENDOREXPERIMENT="1" CGO_ENABLED=1 GOOS=$(GOOS) GOARCH=$(GOARCH)
-GO     := $(GOENV) go
-CGO    := $(CGOENV) go
-GOTEST := USE_EXISTING_CLUSTER=false NO_PROXY="${NO_PROXY},testhost" go test
-SHELL  := bash
+GOOS   	:= $(if $(GOOS),$(GOOS),"")
+GOARCH 	:= $(if $(GOARCH),$(GOARCH),"")
+GOPROXY := $(if $(GOPROXY),$(GOPROXY),"direct")
```
YangKeao, line 28: Or `https://proxy.golang.org,direct`?

Sorry for wrong opinion before. I just prefer the real default value of `go`. It seems that the default value of `go` is `https://proxy.golang.org,direct`

YangKeao: Or could we unset the `GOPROXY` env, if it's not set by the user?

Hexilee: OK, we can just set `GOPROXY=""`, the go binary will use `GOPROXY="https://proxy.golang.org,direct"`.

Hexilee: It seems `GOPROXY=""` will cause some errors in the go code.

Makefile
```diff
@@ -145,7 +146,7 @@ ifeq ($(TARGET_PLATFORM),)
 	endif
 endif
 
-BUILD_INDOCKER_ARG := --env IN_DOCKER=1 --volume $(ROOT):/mnt --user $(shell id -u):$(shell id -g) --platform=linux/$(TARGET_PLATFORM)
+BUILD_INDOCKER_ARG := --env IN_DOCKER=1 --env HTTP_PROXY=${HTTP_PROXY} --env HTTPS_PROXY=${HTTPS_PROXY} --env GOPROXY=${GOPROXY} --volume $(ROOT):/mnt --user $(shell id -u):$(shell id -g) --platform=linux/$(TARGET_PLATFORM)
```
STRRL, line 149: Do we really need `HTTP_PROXY` and `HTTPS_PROXY`? I think it should not be considered by us.

Hexilee: `HTTP_PROXY` and  `HTTPS_PROXY` are the most common solution to access web resources that cannot be accessed by the local networks. 
For our use cases, we need  `HTTPS_PROXY` to download binaries from the GitHub release.

STRRL: I mean the network connectivity should not consider by the project itself. 🙈

`GOPROXY` makes sense because it's a part of `go env`, and actually I prefer to use something like `LD_FLAGS` for passing all the configurable flags or ENVs into go command.

Hexilee: Building in containers makes it difficult to pass HTTP proxy configurations, so I think we have the duty to make it easier (as router proxies or transparent proxies are more difficult to set up).

YangKeao (approved):
LGTM

Makefile
STRRL: > For our use cases, we need HTTPS_PROXY to download binaries from the GitHub release.

It seems we only need to download binaries when we build dockerimage. And HTTP_PROXY/HTTPS_PROXY have already been set in `DOCKER_BUILD_ARGS` at line 7. Do I miss something?

Hexilee: It seems you are right, we do not need to download binaries in this step. Let me clear my opinion, if we develop and build binaries on the host machine, the network connectivity should not consider by the project itself. However, now all the dev tools and build tools are running in containers, so we have the duty to provide ways to configure HTTP proxies in the container.

For pure golang projects, env `HTTPS_PROXY` and `GOPROXY` both works, and developers can choose one of them to solve network problems.

YangKeao: > > For our use cases, we need HTTPS_PROXY to download binaries from the GitHub release.
> 
> It seems we only need to download binaries when we build dockerimage. And HTTP_PROXY/HTTPS_PROXY have already been set in `DOCKER_BUILD_ARGS` at line 7. Do I miss something?

A lot of dependencies are downloaded in the container while we are building the binary inside the container.

STRRL: > A lot of dependencies are downloaded in the container while we are building the binary inside the container.

I do not get it, I think the `dev-env` already contains the binary tools so we do not need to download them anymore. 🤔

STRRL: > It seems you are right, we do not need to download binaries in this step. Let me clear my opinion, if we develop and build binaries on the host machine, the network connectivity should not consider by the project itself. However, now all the dev tools and build tools are running in containers, so we have the duty to provide ways to configure HTTP proxies in the container.
> 
> For pure golang projects, env `HTTPS_PROXY` and `GOPROXY` both works, and developers can choose one of them to solve network problems.


I get your point. LGTM!

STRRL (approved):
LGTM

PR #2472, opened 2021-11-05
add pre-finalizer to protect chaos from being deleted straightly

### LINKED ISSUE #2449: applied chaos cannot be recovered if the CR is deleted before adding finalizers
## Bug Report

1.2.x and 2.0.x

**What did you do?**

Delete an XXXChaos instance.

All injected faults are recovered.

**What did you see instead?**

CR is deleted, by faults are still injected.

**Possible Reasons**

In the 1.2.x case, if the chaos is being reconciled and is being applied, and users delete the chaos, the chaos will be deleted without recovering (because the finalizer hasn't been set). The way to fix this problem is to add a finalizer before applying the chaos.

In the 2.0.x case, this problem is not so obvious, as the chaos will not be applied multiple times (so that, only the very beginning and resuming from pause will run into this bug), but running the finalizer controller before the others will also make sure it doesn't happen.

Hexilee:

Close #2449

Take common controllers into a pipeline in order.

### Checklist

Tests

- [ ] Unit test
- [x] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DIFF
--- a/controllers/common/types.go
+++ b/controllers/common/types.go
@@ -18,9 +18,11 @@ import (
 	"time"
 
 	"github.com/chaos-mesh/chaos-mesh/api/v1alpha1"
+	"github.com/chaos-mesh/chaos-mesh/pkg/finalizer"
 	ctx "github.com/chaos-mesh/chaos-mesh/pkg/router/context"
 	endpoint "github.com/chaos-mesh/chaos-mesh/pkg/router/endpoint"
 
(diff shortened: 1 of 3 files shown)


YangKeao:
When the objects are deleted, new finalizers cannot be added. This PR cannot solve the problem. I will try to propose a new one.

Hexilee:
Observed by #2509, close.

Issue #2477, opened 2021-11-05
helm charts: unexpected image name splicing  


STRRL:
## Feature Request

**Is your feature request related to a problem? Please describe:**

Yes.

`make install` could not works well on the current master (bfee6297966fb7c38aa28e38db97f3670ea46135), after #2462 merged.

After we migrated the docker images from dockerhub to `ghcr.io`,  the way to render the image name is not suitable anymore, because the `ghcr.io` repository domain has been brought by default: 

Ref: https://github.com/chaos-mesh/chaos-mesh/blob/0d683c66c83874ff2e05f57b66cb30de3470c6fb/helm/chaos-mesh/templates/chaos-dashboard-deployment.yaml#L56

We are now using  ` {{template "registry-prefix" .}}{{ .Values.dashboard.image }}`, but the registry-prefix should already be a part of image name.

Our official image is `ghcr.io/chaos-mesh/chaos-mesh/chao-mesh/:latest`, if the user has a self-hosted mirror with domain `mirror.self.hosted`, and the user setup `registry` in the helm values, the final result is `mirror.self.hosted/ghcr.io/chaos-mesh/chaos-mesh/chao-mesh/:latest`, which is incorrect.

**Describe the feature you'd like:**

Updated:

**Pattern B with `registry`, `repository`, `tag` is chosen to use.**

 I'm not sure about which is the most fluency way to configure the image. I would give some available solutions in the comments. :)

Feel free to leave any comments!

**Describe alternatives you've considered:**

No other alternatives.

**Teachability, Documentation, Adoption, Migration Strategy:**

STRRL:
## Pattern A:
Remove `registry` directly, only use `image`.

If user use 
```
helm upgrade --set registry: mirrors.self.hosted
```
before, he should use
```
helm upgrade --set controllerManager.image ghcr.io/chaos-mesh/chaos-mesh:latest \
  --set controllerManager.image ghcr.io/chaos-mesh/chaos-mesh:latest \
  --set chaosDaemon.image ghcr.io/chaos-mesh/chaos-daemon:latest \
  --set dashboard.image ghcr.io/chaos-mesh/chaos-dashboard:latest \
  --set bpfki.image ghcr.io/chaos-mesh/chaos-kernel:latest
```
after the updates.

Many flags need to be set, but if the user using a `values.yaml` to setup the helm installation, the operation is not so different.

Well-known projects use this pattern:
[...]

PR #2496, opened 2021-11-09
chore: restructure image for primiary components in helm

### LINKED ISSUE #2476: bad images left in install.sh
## Question

Although there is a pr for migrating from dockerhub to ghcr.io in helm charts, there are still some old images left in instasll.sh.

https://github.com/chaos-mesh/chaos-mesh/blob/0d683c66c83874ff2e05f57b66cb30de3470c6fb/install.sh#L632-L634

These images are used to load images into the local kind cluster.

I am not sure it still makes sense but the appearance of the old `pingcap/chaos-mesh:latest` is definitely a bug.

STRRL:

close #2476
close https://github.com/chaos-mesh/chaos-mesh/issues/2477
> Problem Summary:

As #2477 says, we'd better use a clear way to render/struct the final image name.

> What's Changed:

- refine the image name with three parts `registry`, `repository`, and `tag`
- update the related scripts for CI
- update `install.sh` and `hack/update_install_script.sh` for the latest updates(that would fix #2476)

### Related changes

- [ ] Need to update `chaos-mesh/website`. No, I walk through the documents, the changed fields are not introduced at the document site. So I think we do not need update some related docs

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```


helm/chaos-mesh/values.yaml
```diff
@@ -43,7 +43,7 @@ enableProfiling: true
 # enableCtrlServer is a flag to enable ctrlserver in controller-manager and chaos-daemon.
 enableCtrlServer: true
 
-registry: ""
+registry: "ghcr.io"
```
iguoyr, line 46: Could we add a global `tag` here, and if user specifies the component `tag`, it will be used in preference, if not, the global `tag` will be used.

STRRL: OK!

STRRL: > Could we add a global `tag` here, and if user specifies the component `tag`, it will be used in preference, if not, the global `tag` will be used.

Do we need the same thing on `registry`?

iguoyr: I think it's a great idea :D

STRRL: > I think it's a great idea :D

Updated! PTAL

helm/chaos-mesh/templates/_helpers.tpl
```diff
@@ -113,5 +113,30 @@ Define the webhook's name
...
+{{- end -}}
+
+{{/*Define the image for chaos-controller-manager*/}}
+{{- define "chaos-controller-manager.image" -}}
+{{.Values.controllerManager.image.registry | default .Values.images.registry}}/{{.Values.controllerManager.image.repository}}:{{.Values.controllerManager.image.tag |  default .Values.images.tag}}
+{{- end -}}
+
+{{/*Define the image for chaos-daemon*/}}
+{{- define "chaos-daemon.image" -}}
+{{.Values.images.registry | default .Values.images.registry}}/{{.Values.chaosDaemon.image.repository}}:{{.Values.chaosDaemon.image.tag | default .Values.images.tag}}
```
iguoyr, line 126: typo, it should starts with `{{ .Values.xxx.images.registry .....` here 👀

helm/chaos-mesh/templates/_helpers.tpl
```diff
@@ -113,5 +113,30 @@ Define the webhook's name
...
+{{- end -}}
+
+{{/*Define the image for chaos-dashboard*/}}
+{{- define "chaos-dashboard.image" -}}
+{{.Values.images.registry | default .Values.images.registry}}/{{.Values.dashboard.image.repository}}:{{.Values.dashboard.image.tag | default .Values.images.tag}}
```
iguoyr, line 131: ditto

helm/chaos-mesh/templates/_helpers.tpl
```diff
@@ -113,5 +113,30 @@ Define the webhook's name
...
+{{- end -}}
+
+{{/*Define the image for chaos-kernel*/}}
+{{- define "chaos-kernel.image" -}}
+{{.Values.images.registry | default .Values.images.registry}}/{{.Values.bpfki.image.repository}}:{{.Values.bpfki.image.tag| default .Values.images.tag}}
```
iguoyr, line 136: ditto

helm/chaos-mesh/templates/_helpers.tpl
```diff
@@ -113,5 +113,30 @@ Define the webhook's name
...
+{{- end -}}
+
+{{/*Define the image for chaos-dlv*/}}
+{{- define "chaos-dlv.image" -}}
+{{.Values.images.registry | default .Values.images.registry}}/{{.Values.chaosDlv.image.repository}}:{{.Values.chaosDlv.image.tag| default .Values.images.tag}}
```
iguoyr, line 141: ditto

helm/chaos-mesh/templates/_helpers.tpl
```diff
@@ -113,5 +113,30 @@ Define the webhook's name
 Define the prefix of
 */}}
 {{- define "registry-prefix" -}}
-{{if .Values.registry}}{{.Values.registry}}/{{end}}
+{{if .Values.images.registry}}{{.Values.images.registry}}/{{end}}
```
iguoyr, line 116: As [this best practice](https://helm.sh/docs/chart_best_practices/templates/#formatting-templates) suggested, we'd better to format these templates with some whitespaces :)

helm/chaos-mesh/templates/_helpers.tpl
STRRL: fixed in 23d2e2ff

helm/chaos-mesh/templates/_helpers.tpl
STRRL: updated in 6af6e041, PTAL! :)

iguoyr (approved):
LGTM

YangKeao (approved):
LGTM

STRRL:
e2e test failed, I'm profiling it

Issue #2397, opened 2021-10-19
Add monitoring metrics about Chaos Mesh


TangliziGit:
## Feature Request

**Is your feature request related to a problem? Please describe:**

More metrics are needed for improving the observability of chaos-controller-manager, chaos-daemon, and chaos-dashboard. Now, we already have several metrics in chaos-controller-manager and chaos-daemon, but it's not enough.

**Describe the feature you'd like:**

There is an RFC describing the feature in detail: https://github.com/chaos-mesh/rfcs/pull/23
Below is a list of metrics which needs to be implemented:
- [x] chaos_controller_manager_chaos_schedules: #2402
- [x] chaos_controller_manager_chaos_workflows: #2402
- [x] chaos_controller_manager_emitted_event_total: #2435
- [x] chaos_controller_manager_grpc_client_handling_seconds: #2458
- [x] chaos_daemon_grpc_server_handling_seconds: #2543
- [x] chaos_daemon_bpm_controlled_process_total: #2497
- [x] chaos_daemon_bpm_controlled_processes: #2497
- [x] chaos_daemon_iptables_chains: #2540
- [x] chaos_daemon_iptables_rules: #2540
- [x] chaos_daemon_iptables_packets: #2540
- [x] chaos_daemon_iptables_packet_bytes: #2540
- [x] chaos_daemon_ipset_members: #2540
- [x] chaos_daemon_tcs: #2540
- [x] chaos_dashboard_http_request_duration_seconds: #2543
- [x] chaos_dashboard_archived_objects: #2568

**Describe alternatives you've considered:**

No.

**Teachability, Documentation, Adoption, Migration Strategy:**

These metrics will be exposed on `/metrics` HTTP endpoint which Prometheus can collect. Users should use these metrics to make dashboards through G to show the operation of the system.

PR #2497, opened 2021-11-09
feat(metrics): add bpm controlled processes metrics

TangliziGit:

> Problem Summary: a part of #2397

> What's Changed:
1. add `bpm.GetControlledProcessIDs` method to query the process IDs controlled by background process manager, and add a testcase for `bpm.GetControlledProcessIDs`.
2. add `bpmControlledProcessTotal` and `bpmControlledProcesses` metrics.

### Related changes

### Checklist

Tests

- [x] Unit test
- [ ] E2E test
 - create HTTPChaos to trigger `bpm.StartProcess`, then observe the `/metrics` endpoint to see if there are corresponding metrics.
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```


yeya24 (commented):
I am not sure why do we need bpm process metrics? Can you please give an example? I don't see the benefits or usecase of having it.

TangliziGit:
> I am not sure why do we need bpm process metrics? Can you please give an example? I don't see the benefits or usecase of having it.

In order to visually observe the health status of `chaos-daemon`. For example, when a Schedule containing IOChaos is submitted and the IOChaos executes, if the target folder does not exist, the execution of IOChaos will fail due to mounting failure. At the same time, the reconciler will always retry to execute the same error. If there are other types of chaos that need to be injected into the same container using BPM at the same time, it may cause a mess because the BPM identifiers lock is not released.

Of course, you can use the log to notice the execution failure, but more intuitive metrics may be more helpful: `chaos_daemon_bpm_controlled_process_total` will grow rapidly over a period of time, while `chaos_daemon_bpm_controlled_processes` will not change.

pkg/bpm/bpm.go
```diff
@@ -301,6 +315,26 @@ func (m *BackgroundProcessManager) Stdio(pid int, startTime int64) *Stdio {
...
+	pid := os.Getpid()
+	for _, p := range processes {
+		if pid == p.PPid() {
+			res = append(res, p.Pid())
+		}
+	}
+
+	log.Info("all controlled processes has been scanned", "PID", res)
+	return res, nil
+}
```
STRRL, line 336: This implementation may introduce more processes that might be not controller by this bpm instance. You are finding that all the processes that controller by **this process**, but one process might use more than one instance of `bpm` and has other logic that spawns new processes. 

I suggest that use `identifiers *sync.Map` in struct `BackgroundProcessManager`, the key of this map store the pid.

pkg/metrics/chaos_daemon.go
```diff
@@ -0,0 +1,70 @@
...
+type ChaosDaemonMetricsCollector struct {
+	backgroundProcessManager  bpm.BackgroundProcessManager
+	bpmControlledProcessTotal prometheus.Counter
+	bpmControlledProcesses    prometheus.Gauge
+}
+
+// NewChaosDaemonMetricsCollector initializes metrics for each chaos daemon
+func NewChaosDaemonMetricsCollector() *ChaosDaemonMetricsCollector {
+	return &ChaosDaemonMetricsCollector{
+		backgroundProcessManager: bpm.NewBackgroundProcessManager(nil),
```
STRRL, line 36: As the former comments say, you need to use another way to assembly this `ChaosDaemonMetricsCollector` with the certain in-use `bpm` instance. :D

STRRL (changes requested):
PTAL @TangliziGit

pkg/bpm/bpm.go
TangliziGit: OK, I will modify it later.

pkg/metrics/chaos_daemon.go
TangliziGit: OK, I got it.

pkg/bpm/bpm.go
TangliziGit: Done.

STRRL (approved):

STRRL:
/cc @YangKeao @Hexilee @iguoyr @Andrewmatilde 

PTAL

cwen0 (commented):
Rest LGTM

YangKeao (approved):
LGTM

Issue #1659, opened 2021-04-08
dashboard: can't insert experiment in MySQL NO_ZERO_IN_DATE mode


g1eny0ung:
## Bug Report

**What version of Kubernetes are you using?**

> Client Version: version.Info{Major:"1", Minor:"20+", GitVersion:"v1.20.4-dirty", GitCommit:"e87da0bd6e03ec3fea7933c4b5263d151aafd07c", GitTreeState:"dirty", BuildDate:"2021-03-15T09:58:13Z", GoVersion:"go1.16.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.2", GitCommit:"faecb196815e248d3ecfb03c680a4507229c2a56", GitTreeState:"clean", BuildDate:"2021-01-13T13:20:00Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}

**What version of Chaos Mesh are you using?**

> Latest

**What did you do?**

Helm install with MySQL driver:

```sh
helm install chaos-mesh helm/chaos-mesh --namespace=chaos-testing --set dashboard.create=true,dashboard.securityMode=false,dashboard.env.DATABASE_DRIVER=mysql,dashboard.env.DATABASE_DATASOURCE="root@tcp(docker.for.mac.localhost:3306)/chaosmesh?parseTime=True"
```

**What did you expect to see?**

After creating an experiment, the database should insert its record successfully.

**What did you see instead?**

Data insertion failed.

After my test, seems this related to https://dev.mysql.com/doc/refman/5.7/en/date-and-time-types.html.

Because the experiment will not set an exact finish time in the beginning. So we will get an error like: `Incorrect datetime value: '0000-00-00'`.

A temporary solution:

```sql
show variables like 'sql_mode';
-- Copy all values except NO_ZERO_IN_DATE.
set global sql_mode=xxx  -- set again
```

Since zero time is not recommended, we need to set the unknown time to 1970-01-01(an example) to avoid this problem.

Atrbear:
Then it is fixed? I meet this problem when using 2.5.1 version

Atrbear:
I am using mysql 8.0.32; 

> show variables like 'sql_mode';
[...]

PR #2510, opened 2021-11-10
Dashboard: Temporary workaround for Issue #1659

craig-seeman:

This is a possible (temporary?) fix until something more official can be placed within #1659.

Note:
+ I am not a very proficient go developer, so this may be a horrible solution, but through some manual testing, seems to be ok.
+ In order for this to work I also had to append `?parseTime=true` on the end of my existing MySQL DSN

> Problem Summary:

> What's Changed:

The solve here was to change some struct definitions containing FinishTime time.time into pointers and update any references (I hope I did not miss any) to allow gorm to place a nil instead of marking the date as 00-00-0000 when initial creation happens (as the experiment is ongoing / not finished.) 

### Checklist

Tests

- [X] Manual test

A few manual tests performed to ensure Experiments and schedules work.

Side effects

None Observed

### Release note 

```text
Allow Mysql 5.7+ to not error out on creation of dates that gorm set as 00-00-0000
```


STRRL:
PTAL
/cc @g1eny0ung @fewdan

STRRL:
Hi @craig-seeman, there are some failed CI checks, would you like to execute `make check` and pushes the changes?

PR #2512, opened 2021-11-11
test: run test for crclients pkg

YangKeao:

The test for container runtime clients package doesn't run (because there is no `Test*` function), and some test logic is wrong.


STRRL (approved):
LGTM

WangXiangUSTC (approved):
LGTM

PR #2522, opened 2021-11-12
Makefile: remove buildx and load argument

YangKeao:

The nightly build failed because `--load` is not supported.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -280,7 +280,7 @@ else
 endif
 
 else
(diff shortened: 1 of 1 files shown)


YangKeao:
@iguoyr Could you help to verify whether this PR will break building on arm macos?

iguoyr (approved):
LGTM, it works well on arm machine :D

fewdan (approved):
LGTM

PR #2509, opened 2021-11-10
*Chaos: add pre-finalizer and record injecting information in the `.Status.Experiment.PodRecords`

YangKeao:
In this PR, the original finalizer was replaced by the `.Status.Experiment.PodRecords`. The `Recover` process will iterate the injected pods according to this field, but not the finalizers.

PR #2534, opened 2021-11-17
client: use updatedClient to fix cache problem

YangKeao:

After #2509 , the cache in client breaks the twophase controller:

1. The controller updates the chaos into Running.
2. The controller reconciles the chaos, gets a cache and think it's waiting
3. Step into the finished state, merge the conflict, remove the finalizer and step out.

In this PR, I backported the `UpdatedClient` in `release-2.0` to make the cache at least updated to the latest successfully `Update`d object.


Hexilee (approved):
LGTM

cwen0 (approved):
LGTM

PR #2554, opened 2021-11-23
auth: add rooturl configuration in gcp authentication

YangKeao:

Try to get the scheme and host from the request is not realistic. If the dashboard is deployed behind a load balancer, it will have no way to guess the scheme. In this PR, I added a config field, so that the users can set the root url manually. The dashboard will append `/api/auth/gcp/callback` to the path.

BTW, it also installs the `ca-certificates` to the `chaos-dashboard` container, which is required to avoid x509 error.


STRRL (approved):
LGTM! It works!

g1eny0ung (approved):
LGTM 🍺

PR #2558, opened 2021-11-24
daemon: avoid tls downgrade attack by setting min version of tls

YangKeao:

The default minimum version of `TLS` is 1.0. This PR will set it to TLS 1.3

It doesn't have any compatibility problem because Chaos Mesh is using TLS 1.3 now.


YangKeao:
I have verified that it avoids serving for the lower version of tls.

```bash
openssl s_client -connect localhost:31767 -tls1_2
```

cwen0 (approved):
LGTM

STRRL (approved):
LGTM

PR #2568, opened 2021-11-27
feat(metrics): add archived objects metrics for chaos-dashboard

TangliziGit:
Signed-off-by: TangliziGit <tanglizimail@foxmail.com>

> Problem Summary: a part of #2397

> What's Changed:
1. add archived objects metrics for `/pkg/dashboard/store`.
2. expose metric host and port for `chaos-dashboard` in `/helm/chaos-mesh/values.yaml`.

### Related changes

- [x] Need to update `Chaos Dashboard`

### Checklist

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

In previous versions, the metric endpoint was exposed by manually adding the environment variable `METRIC_ADDRESS: "0.0.0.0:8080"` to the deployment configuration of 'chaos-dashboard'. But the port can not be exposed in the helm template, because `spec.template.spec.containers.ports.containerPort` in deployment configurations and service configurations should be an integer.

In order to expose metric port from the helm template, `ChaosDashboardConfig.MetricAdress` in `/pkg/config/dashboard/dashboard.go` is modified for `ChaosDashboardConfig.MetricHost` and `ChaosDashboardConfig.MetricPort`.

As a result, if the chaos-dashboard is installed by helm, there is no change required. However, if it is installed manually, you need to change the environment variable `METRIC_ADDRESS: "0.0.0.0:8080"` to `METRIC_HOST: "0.0.0.0"` and `METRIC_PORT: 8080`.

### DIFF
--- a/helm/chaos-mesh/values.yaml
+++ b/helm/chaos-mesh/values.yaml
@@ -45,7 +45,7 @@ enableCtrlServer: true
 
 images:
   # images.registry is the global container registry for the images, you could replace it with your self-hosted container registry.
-  registry: "localhost:5000"
+  registry: "ghcr.io"
   # images.tag is the global image tag (for example, semiVer with prefix v, or latest).
   tag: "latest"
 
--- a/helm/chaos-mesh/values.yaml
+++ b/helm/chaos-mesh/values.yaml
@@ -45,7 +45,7 @@ enableCtrlServer: true
 
 images:
   # images.registry is the global container registry for the images, you could replace it with your self-hosted container registry.
-  registry: "ghcr.io"
+  registry: "localhost:5000"
   # images.tag is the global image tag (for example, semiVer with prefix v, or latest).
   tag: "latest"
 
@@ -276,6 +276,8 @@ dashboard:
   env:
     LISTEN_HOST: "0.0.0.0"
     LISTEN_PORT: 2333
+    METRIC_HOST: "0.0.0.0"
+    METRIC_PORT: 8080
 
     # If you'd like to use a DB other than SQLite (the default), set a driver + DSN here.
     DATABASE_DRIVER: sqlite3
--- a/helm/chaos-mesh/values.yaml
+++ b/helm/chaos-mesh/values.yaml
@@ -277,7 +277,7 @@ dashboard:
     LISTEN_HOST: "0.0.0.0"
     LISTEN_PORT: 2333
     METRIC_HOST: "0.0.0.0"
-    METRIC_PORT: 8080
+    METRIC_PORT: 2334
 
     # If you'd like to use a DB other than SQLite (the default), set a driver + DSN here.
     DATABASE_DRIVER: sqlite3
--- a/pkg/dashboard/store/metrics/metrics.go
+++ b/pkg/dashboard/store/metrics/metrics.go
@@ -20,11 +20,14 @@ import (
 
 	"github.com/prometheus/client_golang/prometheus"
 	"go.uber.org/fx"
+	ctrl "sigs.k8s.io/controller-runtime"
 
 	"github.com/chaos-mesh/chaos-mesh/pkg/dashboard/core"
 )
 
 const chaosDashboardSubsystem = "chaos_dashboard"
+
+var log = ctrl.Log.WithName("metrics-collector")
 
 // Collector implements prometheus.Collector interface
 type Collector struct {
@@ -83,6 +86,7 @@ func (collector *Collector) collectArchivedExperiments() {
 
 	metas, err := collector.experimentStore.ListMeta(context.TODO(), "", "", "", true)
 	if err != nil {
+		log.Error(err, "fail to list all archived chaos experiments")
 		return
 	}
 
@@ -107,6 +111,7 @@ func (collector *Collector) collectArchivedSchedules() {
 
 	metas, err := collector.scheduleStore.ListMeta(context.TODO(), "", "", true)
 	if err != nil {
+		log.Error(err, "fail to list all archived schedules")
 		return
 	}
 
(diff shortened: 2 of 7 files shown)


pkg/dashboard/store/metrics/metrics.go
```diff
@@ -0,0 +1,157 @@
...
+			Subsystem: chaosDashboardSubsystem,
+			Name:      "archived_experiments",
+			Help:      "Total number of archived chaos experiments",
+		}, []string{"namespace", "type"}),
+		archivedSchedules: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Subsystem: chaosDashboardSubsystem,
+			Name:      "archived_schedules",
+			Help:      "Total number of archived chaos schedules",
+
+			// TODO: labels
```
STRRL, line 56: I think there are no other labels that could be attached to this metric. Maybe we could remove this line?

pkg/dashboard/store/metrics/metrics.go
```diff
@@ -0,0 +1,157 @@
...
+		}, []string{"namespace"}),
+		archivedWorkflows: prometheus.NewGaugeVec(prometheus.GaugeOpts{
+			Subsystem: chaosDashboardSubsystem,
+			Name:      "archived_workflows",
+			Help:      "Total number of archived chaos workflows",
+
+			// TODO: labels
```
STRRL, line 63: ditto

STRRL (commented):
rest LGTM!

pkg/dashboard/store/metrics/metrics.go
TangliziGit: Oh, I forgot to delete this line.

TangliziGit: Done.

pkg/dashboard/store/metrics/metrics.go
TangliziGit: Done.

STRRL (approved):
LGTM

pkg/dashboard/store/metrics/metrics.go
```diff
@@ -0,0 +1,153 @@
...
+	collector.archivedSchedules.Collect(ch)
+	collector.archivedWorkflows.Collect(ch)
+}
+
+func (collector *Collector) collectArchivedExperiments() {
+	collector.archivedExperiments.Reset()
+
+	metas, err := collector.experimentStore.ListMeta(context.TODO(), "", "", "", true)
+	if err != nil {
+		return
```
WangXiangUSTC, line 86: Do we need to print error log?

pkg/dashboard/store/metrics/metrics.go
```diff
@@ -0,0 +1,153 @@
...
+func (collector *Collector) Collect(ch chan<- prometheus.Metric) {
+	collector.collectArchivedExperiments()
+	collector.collectArchivedSchedules()
+	collector.collectArchivedWorkflows()
+	collector.archivedExperiments.Collect(ch)
+	collector.archivedSchedules.Collect(ch)
+	collector.archivedWorkflows.Collect(ch)
+}
+
+func (collector *Collector) collectArchivedExperiments() {
```
WangXiangUSTC, line 81: How often does this function run? This function will query all the data in the store. Is it better to collect the metric when the experiment is archived?

STRRL: > How often does this function run?
It depends on how often the Prometheus server pull/scrape/collect data from `chaos-dashboard`. Usually the most frequently collect pattern is every 15 seconds.


> Is it better to collect the metric when the experiment is archived?

I prefer to count all the records in the DB at each time. If we calculate/count the number of archived experiments when we archive them, there needs more logic to resolve about things like `chaos-dashboard` process restart and the user manually updates/ deletes the records in the database.

STRRL: I have considered the resource usage for reading all the records in the table from the database, I think it is just OK, for reading thousands of records from sqlite and MySQL every 15 seconds. 

But using `count` instead of `list-all` in store-layer/SQL is necessary, I think we should make this improvement in the next PRs.

pkg/dashboard/store/metrics/metrics.go
TangliziGit: Yes, we need it. I will add it later.

pkg/dashboard/store/metrics/metrics.go
WangXiangUSTC: OK, We can optimize it at the SQL level first

WangXiangUSTC (approved):
LGTM

pkg/dashboard/store/metrics/metrics.go
WangXiangUSTC: need to create a issue for this

helm/chaos-mesh/values.yaml
```diff
@@ -276,6 +276,8 @@ dashboard:
   env:
     LISTEN_HOST: "0.0.0.0"
     LISTEN_PORT: 2333
+    METRIC_HOST: "0.0.0.0"
+    METRIC_PORT: 8080
```
g1eny0ung, line 280: Port `8080` is usually used for the default web service. I don't really agree with exposing metrics-related stuff to this.

g1eny0ung: @STRRL @WangXiangUSTC Any suggestions? For example, `2334`.

STRRL: yes,using 2334 makes it looks better, because even the origin listen port did not use 8080.

TangliziGit: Done.

PR #2570, opened 2021-11-28
refactor pkg/time

STRRL:

No. It does not relate to any problem. 

This PR aims to make it easier to extend TimeChaos Injection.

This PR introducer new interface called `FakeClockInjector` to present a component that could modify the time of the target process. 

```go
type FakeClockInjector interface {
	Inject(pid int) error
	Recover(pid int) error
}
```

Currently, one implementation of `FakeClockInjector` has its corresponding "fake image" as the data which would be poked into the target process. 

For each different `FakeImage`, I suggest building different `FakeClockInjector` based on it. And use the composite pattern if we need to inject more than one fake image into the target process.

We have only one implementation `TimeSkew` for `FakeClockInjector` now.

And it uses full filename(not trim `.o`) as the name of FakeImage, that's a change.

### Related changes

- [x] Need to **cheery-pick to release branches**
	- [x] release-2.1 

### Checklist

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DIFF
--- a/pkg/time/asset.go
+++ b/pkg/time/asset.go
@@ -36,32 +36,6 @@ type FakeImage struct {
 	offset map[string]int
 }
 
-// fakeImages would be initialized in init.
-// Key presents the filename(compiled .o file, for example fake_clock_gettime.o), Value presents the corresponded FakeImage struct.
-// Deprecated: use LoadFakeImageFromEmbedFs
-var fakeImages = map[string]FakeImage{}
-
-func init() {
-	// in this function, we will load fake image from `fakeclock/*.o`
-	entries, err := fakeclock.ReadDir("fakeclock")
-	if err != nil {
-		log.Error(err, "readdir from embedded fs")
-		os.Exit(1)
-	}
-
-	for _, entry := range entries {
-		if entry.Name() == ".embed.o" {
-			// skip the .embed.o file, as it's used to remove the error of `go fmt`
-			continue
-		}
-		fakeImage, err := LoadFakeImageFromEmbedFs(entry.Name())
-		if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/time/asset.go
```diff
@@ -0,0 +1,140 @@
...
+			rela_section, err := r.Data()
+			if err != nil {
+				log.Error(err, "read rela section")
+				os.Exit(1)
+			}
+			rela_section_reader := bytes.NewReader(rela_section)
+
+			var rela elf.Rela64
+			for rela_section_reader.Len() > 0 {
+				binary.Read(rela_section_reader, elfFile.ByteOrder, &rela)
```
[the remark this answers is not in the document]
STRRL: ignore

STRRL:
PTAL

/cc @YangKeao @Andrewmatilde @ReigenDing

pkg/time/asset.go
```diff
@@ -0,0 +1,116 @@
...
+	"embed"
+	"encoding/binary"
+	"os"
+)
+
+//go:embed fakeclock/*.o
+var fakeclock embed.FS
+
+// FakeImage introduce the replacement of VDSO ELF entry and customizable variables.
+// FakeImage could be constructed with
```
iguoyr, line 30: This comment seems not completed :D

STRRL: 🤣

Updated in a1b3a51! PTAL!

YangKeao (approved):
LGTM

iguoyr (approved):
LGTM

PR #2610, opened 2021-12-07
httpchaos: chaos-tproxy will be recreate if instance not found

### LINKED ISSUE #2536: chaos-tproxy process not found after chaos-daemon restarting
## Bug Report

v2.0.x

**What did you do?**

Restart chaos-daemon.

HTTPChaos works well.

**What did you see instead?**

Events of PodHTTPChaos: 

```
 Type     Reason           Age   From          Message
  ----     ------           ----  ----          -------
 Warning  Failed           0s    records       Failed to recover chaos: rpc error: code = Unknown desc = fail to get stdio of process
```

**The Cause and Solutions**

As the chaos-tproxy supports configuration hot-reload, we keep the tproxy processes alive once it's created, and store the pid in the status of PodHTTPChaos. However, the controller of PodHTTPChaos doesn't know the chaos-tproxy instance exits and the network is reset after chaos-daemon restarts.

The configuration hot-reload may be unnecessary currently, but we need it after the [chaos-tproxy#19](https://github.com/chaos-mesh/rs-tproxy/pull/19) as loading plugins may cost tens of seconds.

The simplest solution is ignoring the `process not found` errors, but if the chaos-tproxy instance exits accidentally, there will be no events emitted.

A more reasonable solution is creating another pod to run a chaos-tproxy instance of each pod, then the k8s will manage this process. However, creating pods temporarily may cost a lot of time.

A more complete solution is creating another daemon set to run a chaos-tproxy instance for pods on each node. The chaos-tproxy needs a big refactoring for this solution.

Hexilee:

Close #2536

### Related changes

  - [x] release-2.1
  - [x] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


Andrewmatilde (approved):
LGTM

pkg/chaosdaemon/httpchaos_server.go
```diff
@@ -64,15 +64,15 @@ func (s *DaemonServer) ApplyHttpChaos(ctx context.Context, in *pb.ApplyHttpChaos
 	log := log.WithValues("Request", in)
 	log.Info("applying http chaos")
 
-	if in.Instance == 0 {
+	stdio := s.backgroundProcessManager.Stdio(int(in.Instance), in.StartTime)
+	for stdio == nil {
```
cwen0, line 68: How about adding a times limit? If the instances can never be found, here will block the whole program

STRRL: +1, it would be stuck in here if `chaos-tproxy` failed to start by invalid arguments or other potential bugs.

And the deeper issue is that we do not specify the lifecycle/status of a process managed by `bpm`, we always assume that processes controlled by bpm SHOULD work well, but that's out of the scope of this PR.

PTAL @Hexilee

cwen0: @Hexilee PTAL

Hexilee: > How about adding a times limit? If the instances can never be found, here will block the whole program

I think the whole program shouldn't be blocked as this is just a gRPC call, only a goroutine will be blocked. Moreover, this call will timeout if the instances can never be found.

> +1, it would be stuck in here if `chaos-tproxy` failed to start by invalid arguments or other potential bugs.
> 
> And the deeper issue is that we do not specify the lifecycle/status of a process managed by `bpm`, we always assume that processes controlled by bpm SHOULD work well, but that's out of the scope of this PR.
> 
> PTAL @Hexilee

If chaos-tproxy failed to start by invalid arguments or other potential bugs, the experiment SHOULD fail by gRPC timeout.

Hexilee: The first not-found may be caused by restarting of chaos-daemon, so we try to recreate one; the second not-found is unexpected, so we should retry once only.

PTAL changes @STRRL @cwen0

cwen0 (approved):
LGTM

PR #2613, opened 2021-12-08
chore: change the module path to v2

STRRL:

fix the module path for Chaos Mesh, let it could be imported as a dependency, and show correctly on pkg.go.dev

change the module path in go.mod

### Related changes

  - [x] release-2.1
  - [x] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [x] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module github.com/chaos-mesh/chaos-mesh
+module github.com/chaos-mesh/chaos-mesh/v2
 
 require (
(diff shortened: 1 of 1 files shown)


STRRL:
changing module path to v2 is much complex than I thought, I will raise an issue before we actually make that.

Issue #2014, opened 2021-06-22
We need use upper case with some name of chaos


STRRL:
## Question

As https://github.com/golang/go/wiki/CodeReviewComments#initialisms says, we should keep the consistent case with words: AWS, GCP, etc. And we have rename `IoChaos` to `IOChaos`

Related chaos:

- AwsChaos -> AWSChaos
- GcpChaos -> GCPChaos

STRRL:
PTAL @fewdan

fewdan:
I agree with you that we should keep the naming style consistent.

PR #2623, opened 2021-12-09
Replace stress-ng with memStress to implement memory stressChaos

### LINKED ISSUE #2383: Create a program to replace stress-ng to realize mem stress
## Feature Request

Because stress-ng realize memory stress by applying read and write pressure instead of occupying memory space, a program which purely occupies memory space is need to generate a memory StressChaos.

Maybe it's a program that uses mmap's system call to allocate memory space.

FingerLeader:

Close #2383

This PR replaces stress-ng with memStress to implement memory stressChaos, and stressChaos can simulate linear memory growth model using memStress.
You can get the reason at [rfc](https://github.com/chaos-mesh/rfcs/pull/25).
And the implementation of [memStress](https://github.com/chaos-mesh/memStress).

### Related changes

- [x] Need to update `chaos-mesh/website` https://github.com/chaos-mesh/website/issues/232
  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code


STRRL:
@FingerLeader  

Please add `go.mod` and some build script(like Makefile) for https://github.com/chaos-mesh/memStress :)

FingerLeader:
> Please add `go.mod` and some build script(like Makefile) for https://github.com/chaos-mesh/memStress :)

Ohh, @YangKeao reminded me of this, but I still forget :scream:

api/v1alpha1/stresschaos_types.go
```diff
@@ -78,12 +77,18 @@ type StressChaosStatus struct {
...
+	// MemoryUID is the memStress identifier
 	// +optional
-	StartTime *metav1.Time `json:"startTime"`
+	MemoryUID string `json:"memoryUid"`
+	// CpuStartTime specifies when the stress-ng starts
+	// +optional
+	CpuStartTime *metav1.Time `json:"cpuStartTime"`
+	// MemoryStartTime specifies when the memStress starts
+	// +optional
+	MemoryStartTime *metav1.Time `json:"memoryStartTime"`
```
STRRL, line 91: Adding fields LGTM, but deleting/renaming not, that would bring break changes in our API. 

We could only update this field after multi-version support.
FYI @YangKeao @iguoyr 

Please consider still using `UID` and `StartTime` for CPU stress, and update the comments for them.

STRRL: We make breaking changes on JVMChaos in the last 2.1 release. Actually it's dirty as well. But there are not so widely used with JVMChaos, it might be acceptable. :cold_sweat:

Andrewmatilde:
image
I don't know the reason but I think it is amazing.

FingerLeader:
> image I don't know the reason but I think it is amazing.

The memStress process will exit soon because OOM, but there is only a INFO log written in Chaos Daemon, while Controllor Manager doesn't know this expriment has finished and still marks the status as Running.
I think Controllor Manager should know the expriment has finished and remind users to set memory size lower.

Andrewmatilde:
> > image I don't know the reason but I think it is amazing.
> 
> The memStress process will exit soon because OOM, but there is only a INFO log written in Chaos Daemon, while Controllor Manager doesn't know this expriment has finished and still marks the status as Running. I think Controllor Manager should know the expriment has finished and remind users to set memory size lower.

Thanks for your reply , I agree with you.

api/v1alpha1/stresschaos_types.go
STRRL: PTAL @FingerLeader

FingerLeader:
I also tried to use stress-ng to allocate 600G memory, when the process be killed for OOM, stress-ng will create a new process trying to allocate memory.
These two tools perform differently, and I think the process should be killed and let the controllor manager know the experime failed, so that users can create a new experiment with lower memory size.

api/v1alpha1/stresschaos_types.go
FingerLeader: It's necessary for canceling stressors, becasue `stress-ng` and `memStress` have different process UIDs. If we don't want to change the structure of `StressInstance`, we may need something like `stress-ng`'s "server" to manage them.

YangKeao:
> I also tried to use stress-ng to allocate 600G memory, when the process be killed for OOM, stress-ng will create a new process trying to allocate memory. These two tools perform differently, and I think the process should be killed and let the controllor manager know the experime failed, so that users can create a new experiment with lower memory size.

But there is noway to push state from chaos-daemon to chaos-controller-manager now, and the chaos-daemon is the only component which knows this process is died.

api/v1alpha1/stresschaos_types.go
YangKeao: @STRRL @FingerLeader I understand. The problem is that one stresschaos can create both cpu stressor and memory stressor at the same time, so we need two more fields to store the `PID` and `startedTime` for `memStress`.

One dirty solution would be adding a single parent process for these two processes:

```
bash -c "stress-ng & memStress & wait"
```

(I'm not sure whether it would work correctly, especially considering the `pause`, cgroup change and termination problem 😢 .

YangKeao: But I agree with @STRRL , we shouldn't break the compatiblity. If the user is using the `StressChaos` and updating the Chaos Mesh, all existing `StressChaos` will break.

Andrewmatilde: Sorry @YangKeao , whether there are any problem that we parse API-struct like `StressInstance` to a struct like `StressInstanceInner` and use `StressInstanceInner` in apply & recover ?

Andrewmatilde: @FingerLeader Since type of `UID` is string ,we can store everything inside it. We can try marshual UID from,
```
type UID struct {
	CpuUID string `json:"cpuUid"`
	MemoryUID string `json:"memoryUid"`
}
```

YangKeao: > 

No. It's not about the type. It's about whether we can handle the old resource.

STRRL: I just prefer to not change the field name of `UID` and `StartTime`, we could still append fields like `MemoryUID` and `MemroyStartTime`🤣

```go
type StressInstance struct {
	// UID is the stress-ng identifier
	// +optional
	UID string `json:"Uid"`
	// MemoryUID is the memStress identifier
	// +optional
	MemoryUID string `json:"memoryUid"`
	// StartTime specifies when the stress-ng starts
	// +optional
	StartTime *metav1.Time `json:"startTime"`
	// MemoryStartTime specifies when the memStress starts
	// +optional
	MemoryStartTime *metav1.Time `json:"memoryStartTime"`
}
```

STRRL: > @FingerLeader Since type of `UID` is string ,we can store everything inside it. We can try marshual UID from,
> 
> ```
> type UID struct {
> 	CpuUID string `json:"cpuUid"`
> 	MemoryUID string `json:"memoryUid"`
> }
> ```

Please DO NOT DO THAT anywhere, it's dangerous and would lose much information(the purpose of this field and developer's intuition), users and other developers would never know that could be a JSON or YAML or other serialized data. 😭 

A `string` type field could be often abused to storage any serialized data. It breaks a similar rule within database design: "multivalued column". Please treat `string` as other "primitive" type like `int` when you design some structure.

Andrewmatilde: > > @FingerLeader Since type of `UID` is string ,we can store everything inside it. We can try marshual UID from,
> > ```
> > type UID struct {
> > 	CpuUID string `json:"cpuUid"`
> > 	MemoryUID string `json:"memoryUid"`
> > }
> > ```
> 
> Please DO NOT DO THAT anywhere, it's dangerous and would lose much information(the purpose of this field and developer's intuition), users and other developers would never know that could be a JSON or YAML or other serialized data. 😭
> 
> A `string` type field could be often abused to storage any serialized data. It breaks a similar rule within database design: "multivalued column". Please treat `string` as other "primitive" type like `int` when you design some structure.

I think it is also dangerous to treat `string` as `int` if users do not have any preset on it. But if it seems dirty , we can use type definition:
```
type UIDString string

func (s *UIDString) ToUID() UID {
	pass
}

func (s *UIDString) FromUID(uid UID){
	pass
}

type UID struct {
	CpuUID string `json:"cpuUid"`
	MemoryUID string `json:"memoryUid"`
}

type StressInstance struct {
	// UID is the stress-ng identifier
	// +optional
	UID UIDString `json:"Uid"`
}
```

STRRL: I do not get your point @Andrewmatilde 🙈, we could not change the type of the field because of the compatibility.

Andrewmatilde: If we are using type definition ? or type alias ? , the yaml config definition of type will not be changed. 
Example : 
https://github.com/chaos-mesh/chaos-mesh/blob/master/api/v1alpha1/dnschaos_type.go#L66
```
type DNSChaosSpec struct {
	// Action defines the specific DNS chaos action.
	// Supported action: error, random
	// Default action: error
	// +kubebuilder:validation:Enum=error;random
	Action DNSChaosAction `json:"action"`
}
```
https://github.com/chaos-mesh/chaos-mesh/blob/master/helm/chaos-mesh/crds/chaos-mesh.org_dnschaos.yaml#L46
```
          spec:
            description: Spec defines the behavior of a pod chaos experiment
            properties:
              action:
                description: 'Action defines the specific DNS chaos action. Supported
                  action: error, random Default action: error'
                enum:
                - error
                - random
                type: string
``` 

& Because the client of user will send the json info to k8s api server first, the change will not break the compatibility.

FingerLeader: > we could still append fields like

@STRRL @Andrewmatilde I think just add fields is acceptable. I don't want to make the definition of StressInstance be as complicated as @Andrewmatilde said, and this solution eliminates our concerns about compatibility.

Andrewmatilde: @FingerLeader , I agree with you . The simple method may be better.

STRRL (commented):
basically LGTM! PTAL @FingerLeader

pkg/chaosdaemon/stress_server_linux.go
```diff
@@ -33,16 +33,67 @@ import (
...
 		return nil, err
 	}
-	control, err := cgroups.Load(daemonCgroups.V1, daemonCgroups.PidPath(int(pid)))
+
+	// memoryStressor
+	memoryInstance, memoryStartTime, err := s.ExecMemoryStressors(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+	log.Info("message", "cpu", cpuInstance, "mem", memoryInstance)
```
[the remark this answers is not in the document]
FingerLeader: I deleted these two logs that were suggested to be modified. They seem to be debugging informations.

api/v1alpha1/stresschaos_types.go
```diff
@@ -99,43 +104,35 @@ type Stressors struct {
 }
 
 // Normalize the stressors to comply with stress-ng
-func (in *Stressors) Normalize() (string, error) {
-	stressors := ""
+func (in *Stressors) Normalize() (string, string, error) {
+	CPUStressors := ""
+	MemoryStressors := ""
```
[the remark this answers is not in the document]
FingerLeader: Yes, they might misdirect devlopers. I have changed first letters to lowercase.

STRRL (approved):
LGTM!

images/chaos-daemon/Dockerfile
```diff
@@ -42,5 +42,7 @@ RUN case "$TARGET_PLATFORM" in \
     mv /usr/local/bin/libnsenter.so /usr/local/lib/libnsenter.so; \
     curl -L https://github.com/chaos-mesh/rs-tproxy/releases/download/v0.4.5/tproxy-$NSEXEC_ARCH-unknown-linux-gnu.tar.gz | tar xz -C /usr/local/bin
 
+RUN curl -L https://github.com/chaos-mesh/memStress/releases/download/v0.2/memStress_v0.2_linux-amd64.tar.gz | tar xz -C /usr/local/bin
```
cwen0, line 45: We should use the different download link according to different arch.

FingerLeader: GET. I'll add some releases for different arch.

pkg/chaosdaemon/stress_server_linux.go
```diff
@@ -33,16 +33,67 @@ import (
...
+	err = s.backgroundProcessManager.KillBackgroundProcess(ctx, MemoryPid, req.MemoryStartTime)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Info("killing stressor successfully")
+	return &empty.Empty{}, nil
+}
+
+func (s *DaemonServer) ExecCpuStressors(ctx context.Context,
```
cwen0, line 85: s/ExecCpuStressors/ExecCPUStressors/g

FingerLeader: >When using acronyms, use Pascal case or camel case for acronyms more than two characters long. For example, use HtmlButton or htmlButton. However, you should capitalize acronyms that consist of only two characters, such as System.IO instead of System.Io.(https://stackoverflow.com/questions/15526107/acronyms-in-camelcase#:~:text=Some%20guidelines%20Microsoft%20has%20written%20about%20camelCase%20are%3A)


I looked up the acronym using guideline of camelCase and only the first character should be capitalized in the example of "Cpu" in `ExecCpuStressors`.

STRRL: Aha, go has its own naming convention, https://github.com/golang/go/wiki/CodeReviewComments#initialisms.

And we have talked about this before, https://github.com/chaos-mesh/chaos-mesh/issues/2014

STRRL: Just a mention, a bad example of MS's naming convention is `XMLHttpRequest`. 🌚

FingerLeader: GOT IT, I have changed `ExecCpuStressors` to `ExecCPUStressors`. 
But I find a lot of generated code still using identifier like CpuXXXXX. 
>Code generated by the protocol buffer compiler is exempt from this rule. 

It is strange that generated code is not of the same standard as the human written one, although it may be diffcult for generator to identify acronyms.🌚

cwen0 (approved):
LGTM

PR #2668, opened 2021-12-17
chaos-daemon support graceful shutdown

### LINKED ISSUE #2650: subprocesses of chaos-daemon cannot gracefully stop
## Bug Report

v2.x.x

**What did you do?**
delete or upgrade chaos-daemon.

all injected faults are recovered after chaos-daemon stops.

**What did you see instead?**
some faults (like http chaos) are not recovered.

**The Reason**
I found subprocesses of chaos-daemon are killed forcibly by the `runc`:

image
ps

`runc` sends SIGTERM to chaos-daemon, and the subprocesses (`nsexec` and `tproxy`) become orphans. Then, `runc` sends SIGKILL to `nsexec` and `tproxy` and they exit without graceful shutdown.

**Solutions**
We can handle SIGTERM in chaos-daemon, kill all subprocesses managed by the `bpm`, and wait for their shutdowns.

Hexilee:

 Close #2650

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [x] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


pkg/chaosdaemon/server.go
```diff
@@ -147,7 +141,7 @@ func newGRPCServer(containerRuntime string, reg prometheus.Registerer, tlsConf t
 	s := grpc.NewServer(grpcOpts...)
```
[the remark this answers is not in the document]
Hexilee: ignore

pkg/chaosdaemon/server.go
```diff
@@ -159,45 +153,75 @@ type RegisterGatherer interface {
...
 		log.Error(err, "failed to create grpc server")
 		return err
 	}
 
-	g.Go(func() error {
+	go func() {
 		log.Info("Starting http endpoint", "address", httpBindAddr)
 		if err := httpServer.ListenAndServe(); err != nil {
 			log.Error(err, "failed to start http endpoint")
 			httpServer.Shutdown(context.Background())
```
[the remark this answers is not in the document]
Hexilee: ignore

Hexilee:
All tests are passed, PTAL! @YangKeao @STRRL @Andrewmatilde @cwen0

Andrewmatilde:
In my opinion , graceful shitdown is just a death sig sender & waiter. So I am surprised about we have called delete() in deathsig map.  
I think their may have some problem in the design. 
I suggest :

1. For bpm , we don't actually need any channel to send any message to shutdown all the process. Bpm start every process , as a result bpm should manage them. I suggest bpm save every commands & send SIGTERM to them all by itself. You can add a waitgroup to count the death of all the process managed by bpm.
2. For gracefull shutdown, we need a simple moudle regist a death channel & send shutdown message to who regist the message.

Hexilee:
> In my opinion , graceful shitdown is just a death sig sender & waiter. So I am surprised about we have called delete() in deathsig map. I think their may have some problem in the design. I suggest :
> 
>     1. For bpm , we don't actually need any channel to send any message to shutdown all the process. Bpm start every process , as a result bpm should manage them. I suggest bpm save every commands & send SIGTERM to them all by itself. You can add a waitgroup to count the death of all the process managed by bpm.
> 
>     2. For gracefull shutdown, we need a simple moudle regist a death channel & send shutdown message to who regist the message.

Actually, I'm going to refactor bpm, we can have another discussion, but it's irrelevant with this PR.

Andrewmatilde:
I think we should have provided a common graceful shutdown tools , since the title is `chaos-daemon support graceful shutdown`. But now there are only a solution  for BPM .

Hexilee:
> I think we should have provided a common graceful shutdown tools , since the title is `chaos-daemon support graceful shutdown`. But now there are only a solution for BPM .

We can just wait for closing other resources after chaos-daemon receiving SIGTERM, but I cannot figure out common solution to close any resource and wait for their exiting.

pkg/chaosdaemon/server.go
```diff
@@ -159,45 +154,70 @@ type RegisterGatherer interface {
...
+	return &Server{
+		errors:       make(chan error),
+		daemonServer: daemonServer,
+		conf:         conf,
+		reg:          reg,
+	}, nil
+}
+
+// Start starts chaos-daemon.
+func (s *Server) Start() error {
```
STRRL, line 182: How about building `Start()` like `http.Server`'s `ListenAndServe()`? Do not use `Errors() <-chan error`.

pkg/chaosdaemon/server.go
```diff
@@ -159,45 +154,70 @@ type RegisterGatherer interface {
...
+}
+
+// Errors return the error channel
+func (s *Server) Errors() <-chan error {
+	return s.errors
+}
 
-	return g.Wait()
+func (s *Server) Shutdown() {
+	s.daemonServer.backgroundProcessManager.Shutdown()
```
STRRL, line 222: we should also "Close" other resources like `grpc` and `http` server

STRRL (commented):
Graceful shutdown usually relates to release/close opened resources. And it makes sense to` Shutdown()` return an error. Bu

pkg/chaosdaemon/server.go
Hexilee: fine

pkg/chaosdaemon/server.go
Hexilee: How about errgroup?

Hexilee: Fixed

pkg/chaosdaemon/server.go
Hexilee: Fixed

Andrewmatilde (approved):
LGTM

YangKeao (approved):
LGTM

PR #2696, opened 2021-12-22
ci: run different jobs depending on the modified files

### LINKED ISSUE #2233: Proposal: run different CI depending on the modified files
## Feature Request

Yes, this proposal relates to the long-standing question of why our CI has to be run every time, even if just one `.md` file is modified.

Run tests based on the type of change files.

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-paths

g1eny0ung:

Close #2233.

In addition to what the title says, this PR adds UI tests in GitHub Actions and also removes some tests because we decided to use an `OpenAPI` converter to generate HTTP APIs.

### Related changes

  - [x] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


g1eny0ung:
/cc @chaos-mesh/maintainers @chaos-mesh/committers

Here are the more detailed updates:

- The following workflows remove the `on push` event in the `master` branch (since we have already checked it at `pull request`):
  - [x] `checklink.yaml`
  - [x] `codecov_unittest.yaml`
  - [x] `integration_test.yml`
  - [x] `license_checker.yml`
  - [x] `script_test.yml`

- For a long time, we have always run several **required checks** before merging a PR into the master, and if we want to run different jobs depending on modified files, it's needed to skip unmatched jobs. But GitHub Action can't do this at the job level. So in `ci.yml`, I use [dorny/paths-filter@v2](https://github.com/dorny/paths-filter) to do this:
  - Example: https://github.com/dorny/paths-filter#conditional-execution.
  - **Note:** we need to update the settings of `Require status checks to pass before merging` to make modified workflows finally work.

STRRL:
I am afraid that we could not remove the check with codecov_unittest.yaml, because we use squash commit when mething PRs, codecov would not find the correct commit as the base when diff the coverage with PR.

So we should upload the latest coverage report on `push`

g1eny0ung:
> I am afraid that we could not remove the check with codecov_unittest.yaml, because we use squash commit when mething PRs, codecov would not find the correct commit as the base when diff the coverage with PR.
> 
> So we should upload the latest coverage report on `push`

Uh-oh, I forgot to think about it...I will re-add the push event in coverage workflow.

g1eny0ung:
@STRRL PTAL again.

.github/workflows/ci.yml
```diff
@@ -7,24 +14,41 @@ on:
...
+      go: ${{ steps.filter.outputs.go }}
+      ui: ${{ steps.filter.outputs.ui }}
+    steps:
+      # For pull requests it's not necessary to checkout the code
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            go:
+              - makefile
```
STRRL, line 31:
```suggestion
              - Makefile
```

I found that `paths-filter` uses `picomatch` for matching files, and `picomatch` follows case-sensitive by default.

.github/workflows/codecov_unittest.yaml
```diff
@@ -1,9 +1,17 @@
 name: Unit Test And Code Coverage
 
-# this workflow would work on all the branches and all the prs
+# this workflow would work on all the prs
 on:
-  push: {}
-  pull_request: {}
+  pull_request:
+    paths:
+      - makefile
```
STRRL, line 7: ditto

.github/workflows/script_test.yml
```diff
@@ -1,15 +1,13 @@
...
+name: Test install.sh
 on:
-  push:
-    branches:
-      - master
   pull_request:
     branches:
       - master
+    paths:
+      - 'install.sh'
```
STRRL, line 7: I am not sure do we need to test the `install.sh` not only on `install.sh` changes, but also like changes in Dockerfile and CRD.

PTAL @YangKeao @Hexilee @iguoyr

STRRL (approved):
rest LGTM

.github/workflows/codecov_unittest.yaml
```diff
@@ -1,9 +1,17 @@
...
+      - go.*
+      - "**.go"
+  push:
+    paths:
+      - makefile
```
STRRL, line 12: ditto, use `Makefile`

STRRL:
sorry, I just want to click "Comment", not "Approve". Please take a look on the comments @g1eny0ung

g1eny0ung:
> sorry, I just want to click "Comment", not "Approve". Please take a look on the comments @g1eny0ung

Got it! I'm have already fixed them.

.github/workflows/script_test.yml
iguoyr: Integration test will run `hack/local-up-chaos-mesh.sh`, and it calls `install.sh`, will it solve this problem?

STRRL (approved):

iguoyr (approved):
LGTM

Issue #2641, opened 2021-12-14
Some specified chaos experiment returns "not found" on Chaos Dashboard


cwen0:
## Bug Report

**What version of Kubernetes are you using?**
 

v1.22

**What version of Chaos Mesh are you using?**

master-g63345a39946779

**What did you do?**
 

A chaos experiments always keep the `Injecting` status and return a `not found` error when clicking it.  
![image](https://user-images.githubusercontent.com/22956341/145957307-3bc9416d-999f-4367-95b2-d0aa84d28f33.png)

Then click the `Archive` button, return a `not found` error too. 

![image](https://user-images.githubusercontent.com/22956341/145957554-476b5c9e-4933-495a-b028-8ebde3d45f12.png)
![image](https://user-images.githubusercontent.com/22956341/145957589-f53cbc7e-837c-41f8-a159-b53e99c03b08.png)

**What did you expect to see?**

**What did you see instead?**

**Output of chaosctl**

cwen0:
Dashboard log: 

```
[GIN] 2021/12/16 - 09:53:11 | 200 |   87.597026ms |      10.244.0.0 | GET      "/api/experiments"
2021-12-16T09:53:16.141Z        ERROR   apiserver       error.api.resource_not_found
github.com/chaos-mesh/chaos-mesh/pkg/dashboard/apiserver/utils.SetAPIError
        /mnt/pkg/dashboard/apiserver/utils/error.go:70
github.com/chaos-mesh/chaos-mesh/pkg/dashboard/apiserver/experiment.(*Service).delete
        /mnt/pkg/dashboard/apiserver/experiment/experiment.go:332
github.com/gin-gonic/gin.(*Context).Next
        /tmp/go/pkg/mod/github.com/gin-gonic/gin@v1.7.2/context.go:165
github.com/gin-gonic/gin.CustomRecoveryWithWriter.func1
        /tmp/go/pkg/mod/github.com/gin-gonic/gin@v1.7.2/recovery.go:99
github.com/gin-gonic/gin.(*Context).Next
        /tmp/go/pkg/mod/github.com/gin-gonic/gin@v1.7.2/context.go:165
github.com/gin-gonic/gin.LoggerWithConfig.func1
        /tmp/go/pkg/mod/github.com/gin-gonic/gin@v1.7.2/logger.go:241
github.com/gin-gonic/gin.(*Context).Next
[...]

PR #2698, opened 2021-12-23
fix(chaos-dashboard): ignore the first event on `Reconcile` that triggered by deleting action

cwen0:

 

Fixed: #2641 

Deleting chaos experiment action will trigger `Reconcile` function twice.  If we handle the first event, chaos experiments that were blocked by deleting action will ignore when the chaos dashboard was restarted, refer to https://github.com/chaos-mesh/chaos-mesh/issues/2641#issuecomment-1000002369.  To fix it, we can ignore the first trigger event. 

```
[GIN] 2021/12/23 - 12:02:41 | 200 |   12.155996ms | 192.168.197.179 | DELETE   "/api/experiments/713730b9-3183-4826-8ad2-6e5bd1abf0e9"
2021-12-23T12:02:41.235+0800    INFO    collector.HTTPChaos     request {"request:": "default/http-ssss"}  // obj.IsDeleted = true 
2021-12-23T12:02:41.255+0800    INFO    collector.HTTPChaos     request {"request:": "default/http-ssss"}  // obj is not found
```

### Related changes

  - [x] release-2.1
  - [x] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


pkg/dashboard/collector/collector.go
```diff
@@ -95,19 +95,6 @@ func (r *ChaosCollector) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		manageFlag = true
```
g1eny0ung, line 95: The related statements should also be deleted.

g1eny0ung (approved):
Rest LGTM.

pkg/dashboard/collector/collector.go
cwen0: Fixed

hack/lib.sh
```diff
@@ -37,7 +37,7 @@ KUBEBUILDER_PATH=$OUTPUT_BIN/kubebuilder
 KUBEBUILDER_BIN=$KUBEBUILDER_PATH/bin/kubebuilder
 KUBEBUILDER_VERSION=2.3.2
 KUSTOMIZE_BIN=$OUTPUT_BIN/kustomize
-KUSTOMIZE_VERSION=3.5.4
```
STRRL, line 40: It seems this line of change is not related to this issue, is it required?

STRRL (commented):
It seems that after we remove these codes, there is nowhere to update the column `archived` in the database table. And it still uses `archived = true` as the condition when list data: https://github.com/chaos-mesh/chaos-mesh/blob/290890941b6d7667937007a7f889053f46a2e9b4/pkg/dashboard/apiserver/archive/archive.go#L113

hack/lib.sh
cwen0: This is not required by this issue. When I run the e2e test on my local environment, require `KUSTOMIZE` more hight version

cwen0:
> It seems that after we remove these codes, there is nowhere to update the column `archived` in the database table. And it still uses `archived = true` as the condition when list data:
> 
> https://github.com/chaos-mesh/chaos-mesh/blob/290890941b6d7667937007a7f889053f46a2e9b4/pkg/dashboard/apiserver/archive/archive.go#L113

The column `archived` will be updated at https://github.com/chaos-mesh/chaos-mesh/blob/290890941b6d7667937007a7f889053f46a2e9b4/pkg/dashboard/collector/collector.go#L74. When the experiments were deleted successfully, this logic would be executed.

STRRL:
> > It seems that after we remove these codes, there is nowhere to update the column `archived` in the database table. And it still uses `archived = true` as the condition when list data:
> > https://github.com/chaos-mesh/chaos-mesh/blob/290890941b6d7667937007a7f889053f46a2e9b4/pkg/dashboard/apiserver/archive/archive.go#L113
> 
> The column `archived` will be updated at
> 
> https://github.com/chaos-mesh/chaos-mesh/blob/290890941b6d7667937007a7f889053f46a2e9b4/pkg/dashboard/collector/collector.go#L66-L83
> 
> . When the experiments were deleted successfully, this logic would be executed.


I noticed that there might be a confusing logic: 

```go
	obj, ok := r.apiType.DeepCopyObject().(v1alpha1.InnerObject)
	if !ok {
		r.Log.Error(nil, "it's not a stateful object")
		return ctrl.Result{}, nil
	}

	err := r.Get(ctx, req.NamespacedName, obj)
	if apierrors.IsNotFound(err) {
		if chaosMeta, ok = obj.(metav1.Object); !ok {
			r.Log.Error(nil, "failed to get chaos meta information")
		}
		if chaosMeta.GetLabels()[v1alpha1.LabelManagedBy] != "" {
			manageFlag = true
		}
		if !manageFlag {
			if err = r.archiveExperiment(req.Namespace, req.Name); err != nil {
				r.Log.Error(err, "failed to archive experiment")
			}
		} else {
			if err = r.event.DeleteByUID(ctx, string(chaosMeta.GetUID())); err != nil {
				r.Log.Error(err, "failed to delete experiment related events")
			}
		}
		return ctrl.Result{}, nil
	}

```

- `chaosMeta` is built from `obj` with type assertion, and `obj` is spawned from `r.apiType.DeepCopyObject().(v1alpha1.InnerObject)`
- when `apierrors.IsNotFound(err)` is `ture`, `obj` and `chaosMeta` would keep the struct initiaial value, so `chaosMeta.GetLabels()[v1alpha1.LabelManagedBy] != ""` would always be `false`; the if condition on here and following `manageFlag` is meanless.


As `manageFlag` would always be `false`, these codes like `r.event.DeleteByUID(ctx, string(chaosMeta.GetUID()))` would never be reached, please consider to remove them.

In other words, `managedFlag` is meanless within `apierrors.IsNotFound(err)`, but it was useful within the deleted code under `if obj.IsDeleted()`. Do we lose some features or introduce undefined behaviours after removing these codes?  FYI @g1eny0ung

cwen0:
> > > It seems that after we remove these codes, there is nowhere to update the column `archived` in the database table. And it still uses `archived = true` as the condition when list data:
> > > https://github.com/chaos-mesh/chaos-mesh/blob/290890941b6d7667937007a7f889053f46a2e9b4/pkg/dashboard/apiserver/archive/archive.go#L113
> > 
> > 
> > The column `archived` will be updated at
> > https://github.com/chaos-mesh/chaos-mesh/blob/290890941b6d7667937007a7f889053f46a2e9b4/pkg/dashboard/collector/collector.go#L66-L83
> > 
> > . When the experiments were deleted successfully, this logic would be executed.
> 
> I noticed that there might be a confusing logic:
> 
> ```go
> 	obj, ok := r.apiType.DeepCopyObject().(v1alpha1.InnerObject)
> 	if !ok {
> 		r.Log.Error(nil, "it's not a stateful object")
> 		return ctrl.Result{}, nil
> 	}
> 
> 	err := r.Get(ctx, req.NamespacedName, obj)
> 	if apierrors.IsNotFound(err) {
> 		if chaosMeta, ok = obj.(metav1.Object); !ok {
> 			r.Log.Error(nil, "failed to get chaos meta information")
> 		}
> 		if chaosMeta.GetLabels()[v1alpha1.LabelManagedBy] != "" {
> 			manageFlag = true
> 		}
> 		if !manageFlag {
> 			if err = r.archiveExperiment(req.Namespace, req.Name); err != nil {
> 				r.Log.Error(err, "failed to archive experiment")
> 			}
> 		} else {
> 			if err = r.event.DeleteByUID(ctx, string(chaosMeta.GetUID())); err != nil {
> 				r.Log.Error(err, "failed to delete experiment related events")
> 			}
> 		}
> 		return ctrl.Result{}, nil
> 	}
> ```
> 
> * `chaosMeta` is built from `obj` with type assertion, and `obj` is spawned from `r.apiType.DeepCopyObject().(v1alpha1.InnerObject)`
> * when `apierrors.IsNotFound(err)` is `ture`, `obj` and `chaosMeta` would keep the struct initiaial value, so `chaosMeta.GetLabels()[v1alpha1.LabelManagedBy] != ""` would always be `false`; the if condition on here and following `manageFlag` is meanless.
> 
> As `manageFlag` would always be `false`, these codes like `r.event.DeleteByUID(ctx, string(chaosMeta.GetUID()))` would never be reached, please consider to remove them.
> 
> In other words, `managedFlag` is meanless within `apierrors.IsNotFound(err)`, but it was useful within the deleted code under `if obj.IsDeleted()`. Do we lose some features or introduce undefined behaviours after removing these codes? FYI @g1eny0ung

If the `managedFlag` is true, archive action just deletes events and not store them in DB. Maybe we need to update this logic here.

g1eny0ung:
https://github.com/chaos-mesh/chaos-mesh/blob/8cdbeacb31872eccb17dcef5419799243116d13a/pkg/dashboard/collector/collector.go#L66-L83

After finishing read @STRRL's comment, I think this piece code can be removed safely because we still delete the related events in:

https://github.com/chaos-mesh/chaos-mesh/blob/8cdbeacb31872eccb17dcef5419799243116d13a/pkg/dashboard/apiserver/archive/archive.go#L215

@cwen0 What do you think?

cwen0:
https://github.com/chaos-mesh/chaos-mesh/blob/8cdbeacb31872eccb17dcef5419799243116d13a/pkg/dashboard/collector/collector.go#L66-L83 

@STRRL @g1eny0ung The logical purpose here is to avoid saving the experimental objects created by schedule and workflow. We can't remove it directly!

cwen0:
> https://github.com/chaos-mesh/chaos-mesh/blob/8cdbeacb31872eccb17dcef5419799243116d13a/pkg/dashboard/collector/collector.go#L66-L83
> 
> @STRRL @g1eny0ung The logical purpose here is to avoid saving the experimental objects created by schedule and workflow. I don't suggest removing it directly.

cwen0:
@STRRL @g1eny0ung PTAL again! Thanks! 
In this pr, I search managed experiments from the database and delete them instead of using the `manageFlag` flag. Because managed experiments and related events both should be removed from the database when the schedule or workflow is deleted.  Using `manageFlag` will keep the experiment object in the archive list and only delete the related events, this is not the effect we want.

g1eny0ung (approved):
LGTM. Good job. 👍🏽

STRRL (approved):
LGTM!

PR #2708, opened 2021-12-23
feat(chaos dashboard): add deleting status

cwen0:

If the chaos object is deleted and blocked, we should show it on chaos dashboard. In this pr, I add a new status `deleting` to describe this status. 

### Related changes

- [x] Need to update `Dashboard UI`
  - [x] release-2.1
  - [x] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

image

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


g1eny0ung:
The name `Deleted` is more accurate?

cwen0:
> 

I don't think so. Because the real status is `deleting` and the chaos experiment has not been recovered.  `Deleted` maybe make users confused.

cwen0:
> The name `Deleted` is more accurate?

I don't think so. Because the real status is `deleting` and the chaos experiment has not been recovered.  `Deleted` maybe make users confused.

g1eny0ung:
> > The name `Deleted` is more accurate?
> 
> I don't think so. Because the real status is `deleting` and the chaos experiment has not been recovered. `Deleted` maybe make users confused.

According to my understanding, when users click on the `Deleting` experiment, API will return experiment not found, it's also weird. 😢

cwen0:
> > > The name `Deleted` is more accurate?
> > 
> > 
> > I don't think so. Because the real status is `deleting` and the chaos experiment has not been recovered. `Deleted` maybe make users confused.
> 
> But according to my understanding, when users click on the `Deleting` experiment, API will return experiment not found, it's also weird. 😢

This issue was fixed in #2698

g1eny0ung (approved):
Know about it. This LGTM. 👍🏽

STRRL (approved):
LGTM

PR #2731, opened 2021-12-29
fix: specify branch name as env image tag

STRRL:

GitHub Action failed to build images for release: https://github.com/chaos-mesh/chaos-mesh/actions/runs/1633015443

Use the dev-env and build-env with the branch `release-2.1`

### Related changes

  - [x] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [x] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/.github/workflows/upload_image.yml
+++ b/.github/workflows/upload_image.yml
@@ -16,9 +16,12 @@ jobs:
     outputs:
       image_tag: ${{ steps.image_tag.outputs.image_tag }}
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v2
+        with:
+          # it is required for fetching branches which contains certain tag
+          fetch-depth: 0
 
-      - name: Extract Image Tag
+      - name: Extract Image Tags
         shell: bash
         run: |
           IMAGE_TAG=${GITHUB_REF##*/}
@@ -26,7 +29,18 @@ jobs:
             IMAGE_TAG=latest;
           fi
 
+          ENV_IMAGE_TAG=${IMAGE_TAG}
+
+          if [ "${GITHUB_REF_TYPE}" = "tag" ] ; then
+              BRANCHES=$(git branch -r --contains ${{ github.ref }})
+              ENV_IMAGE_TAG_RAW=$(git branch -r --contains ${{ github.ref }} | head -n 1)
+              ENV_IMAGE_TAG=${ENV_IMAGE_TAG_RAW/origin\/}
+              echo -e "available branches for this tag:\n${BRANCHES}\n"
+              echo "use dev-env and build-env for branch: ${ENV_IMAGE_TAG}"
+          fi
+
           echo "::set-output name=image_tag::$(echo $IMAGE_TAG)"
+          echo "::set-output name=env_image_tag::$(echo $ENV_IMAGE_TAG)"
         id: image_tag
 
       - name: Log in to GitHub Docker Registry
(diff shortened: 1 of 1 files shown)


STRRL:
PTAL

/cc @chaos-mesh/committers @chaos-mesh/maintainers

YangKeao (approved):
If the `build-env` and `dev-env` is modified, and this action is run for an older release, it will fail.

But it's not a big problem, I prefer to merge this PR before considering that situation 😃 .

.github/workflows/upload_image.yml
```diff
@@ -16,17 +16,31 @@ jobs:
...
+      - uses: actions/checkout@v2
+        with:
+          # it is required for fetching branches which contains certain tag
+          fetch-depth: 0
 
-      - name: Extract Image Tag
+      - name: Extract Image Tags
         shell: bash
         run: |
           IMAGE_TAG=${GITHUB_REF##*/}
```
g1eny0ung, line 27: This can be replaced by `${{ github.ref_name }}`.

STRRL:
> If the build-env and dev-env is modified, and this action is run for an older release, it will fail.

I think we would only build once for a certain release, so here is no situation for using modified build-env and dev-env for building older releases.

Or should we maintain `build-env` or `dev-env` for the branch, or the tagged commit? Which one would you choose? 🤔

.github/workflows/upload_image.yml
STRRL: Updated! PTAL

STRRL:
> Or should we maintain `build-env` or `dev-env` for the tagged commit instead of branch? Which one would you choose?

Or none of then, set `IMAGE_DEV_ENV_BUILD=1` and `IMAGE_BUILD_ENV_BUILD=1` in `uplad_image.yml`?

g1eny0ung:
> I think we would only build once for a certain release, so here is no situation for using modified build-env and dev-env for building older releases.

@STRRL I prefer this.

> > Or should we maintain `build-env` or `dev-env` for the tagged commit instead of branch? Which one would you choose?
> 
> Or none of then, set `IMAGE_DEV_ENV_BUILD=1` and `IMAGE_BUILD_ENV_BUILD=1` in `uplad_image.yml`?

I thought of this too, do we need to build these two mirrors at release?

g1eny0ung:
> I think we would only build once for a certain release, so here is no situation for using modified build-env and dev-env for building older releases.

@STRRL I think this is more intuitive.

> > Or should we maintain `build-env` or `dev-env` for the tagged commit instead of branch? Which one would you choose?
> 
> Or none of then, set `IMAGE_DEV_ENV_BUILD=1` and `IMAGE_BUILD_ENV_BUILD=1` in `uplad_image.yml`?

I thought of this too, do we need to build these two images at release?

STRRL:
> I thought of this too, do we need to build these two images at release?

Currently, we maintain `dev-env` and `build-env` for each branch: `master` and `release-2.1`. When releasing `2.1.x`,  it should use the image `dev-env:release-2.1` and `build-env:release-2.1`, it does not build these two images at release.

YangKeao:
> > If the build-env and dev-env is modified, and this action is run for an older release, it will fail.
> 
> I think we would only build once for a certain release, so here is no situation for using modified build-env and dev-env for building older releases.

Yes. As we will never "release again" a tag, it's fine to assume that the latest `build-env:v2.1` is the one built for release 2.1.2 (though this assumption cannot be true for a long time.)

I prefer to build a release image with `IMAGE_DEV_ENV_BUILD=1` and `IMAGE_BUILD_ENV_BUILD=1`, which sounds cleaner (though, there is no actual benefit 😃 )

STRRL:
> I prefer to build a release image with `IMAGE_DEV_ENV_BUILD=1` and `IMAGE_BUILD_ENV_BUILD=1`, which sounds cleaner (though, there is no actual benefit  )

+1

I am going to close this PR and create another one for applying these changes.

PR #2789, opened 2022-01-13
bump ginkgo to v2.0.0

Hexilee:

In #2668, I would like to add some e2e tests that run with a serial decorator, so we need ginkgo v2.

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [x] Unit test
- [x] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/e2e-test/e2e/chaos/stresschaos/memory.go
+++ b/e2e-test/e2e/chaos/stresschaos/memory.go
@@ -21,7 +21,7 @@ import (
 	"net/http"
 	"time"
 
-	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/v2"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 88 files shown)


e2e-test/e2e/chaos/stresschaos/memory.go
```diff
@@ -21,7 +21,7 @@ import (
 	"net/http"
 	"time"
 
-	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/v2"
```
YangKeao, line 24: Why don't we globally import this package?

Hexilee: My linter suggested it to me.

Andrewmatilde (approved):
LGTM

Hexilee:
It seems the ginkgo v2 is too hard to be upgraded, we will do it in later days.

PR #2792, opened 2022-01-14
ci: fix cache path calculation

YangKeao:

Recent e2e test image build doesn't hit the cache, because of a bug in calculating the cache directory 😢 This PR solves two problems:

1. The fake_image targets should depend on build-env, but not dev-env.
2. The cache directory should be joined with an image name directory.


STRRL (approved):
LGTM!

STRRL:
Should this PR be cheery-picked to release-2.1?

YangKeao:
> Should this PR be cheery-picked to release-2.1?

No. `build-script` is only merged on master (which is expected, I think)

cwen0 (approved):
LGTM

PR #2839, opened 2022-01-27
fix podhttpchaos: check observedGeneration to avoid duplicated applying

Hexilee:

Now, the controller of PodHttpChaos applies rules in each time reconciling. We should check the observedGeneration to avoid duplicated applying.

Add checking block:

```go
	if obj.ObjectMeta.Generation <= obj.Status.ObservedGeneration && obj.Status.FailedMessage == "" {
		r.Log.Info("the target pod has been up to date", "pod", obj.Namespace+"/"+obj.Name)
		return ctrl.Result{}, nil
	}
```

### Related changes

  - [x] release-2.1
  - [x] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


controllers/podhttpchaos/controller.go
```diff
@@ -58,6 +58,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, nil
 	}
 
+	if obj.ObjectMeta.Generation <= obj.Status.ObservedGeneration && obj.Status.FailedMessage == "" {
```
YangKeao, line 61: Maybe we could write `PodXXXChaosDelegate` (with a better name), as the controllers of `podiochaos`, `podhttpchaos`, and `podnetworkchaos` are nearly the same in the future.

This PR is fine to merge.

YangKeao (approved):
LGTM

fewdan (approved):
LGTM

controllers/podhttpchaos/controller.go
Hexilee: Looks good

PR #2841, opened 2022-01-28
combine groups together before group imports

YangKeao:

Chaos Mesh has its own dev-env now, so we can use our own goimports 😄 . It can combine and sort all groups together.

The single commit is here: https://github.com/YangKeao/tools/commit/2075ebeb37071f8ac14648e2b72c6498b76ec068

It adds a flag to the goimports, to sort after combining all manually split groups together.


Andrewmatilde (approved):
LGTM

STRRL (approved):
LGTM!

PR #2849, opened 2022-02-07
fix: trying to fix real_gettimeofday on arm64

### LINKED ISSUE #2844: timechaos for gettimeofday has not been ported on arm64
## Bug Report

N/A

master, 27e45d2cdef35a6f03482f2ed76a5e763fe8918a

**What did you do?**

N/A

N/A

**What did you see instead?**

Scheduled daily CI failed: https://github.com/chaos-mesh/chaos-mesh/actions/workflows/upload_image.yml

It failed when building chaos-daemon on arm64.

I found that gettimeofday did not be ported on aarch64:

https://github.com/chaos-mesh/chaos-mesh/blob/27e45d2cdef35a6f03482f2ed76a5e763fe8918a/pkg/time/fakeclock/fake_gettimeofday.c#L39-L52

When we review https://github.com/chaos-mesh/chaos-mesh/pull/2742, reviews only test it on amd64.

I think it's necessary to extend our CI onto arm64.

**Output of chaosctl**

STRRL:

close #2844

- add parameter `tz` in `real_gettimeofday`

### Related changes

- [x] Need to update `Dashboard UI`
  - [x] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

It could be compiled on my machine.

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/pkg/time/fakeclock/fake_gettimeofday.c
+++ b/pkg/time/fakeclock/fake_gettimeofday.c
@@ -24,23 +24,23 @@ extern int64_t TV_SEC_DELTA;
 extern int64_t TV_NSEC_DELTA;
 
 #if defined(__amd64__)
-inline int real_gettimeofday(struct timeval *tv, struct timezone *__tz)
+inline int real_gettimeofday(struct timeval *tv, struct timezone *tz)
 {
     int ret;
     asm volatile(
         "syscall"
         : "=a"(ret)
-        : "0"(__NR_gettimeofday), "D"(tv), "S"(__tz)
+        : "0"(__NR_gettimeofday), "D"(tv), "S"(tz)
         : "memory");
 
     return ret;
 }
 
 #elif defined(__aarch64__)
-inline int real_gettimeofday(struct timeval *tv, struct timezone *__tz)
+inline int real_gettimeofday(struct timeval *tv, struct timezone *tz)
 {
     register struct timeval *x0 __asm__("x0") = tv;
-    register struct timezone *x1 __asm__("x1") = __tz;
+    register struct timezone *x1 __asm__("x1") = tz;
     register uint64_t w8 __asm__("w8") = __NR_gettimeofday; /* syscall number */
     __asm__ __volatile__(
         "svc 0;"
@@ -52,9 +52,9 @@ inline int real_gettimeofday(struct timeval *tv, struct timezone *__tz)
 }
 #endif
 
-int gettimeofday(struct timeval *tv, struct timezone *__tz)
+int gettimeofday(struct timeval *tv, struct timezone *tz)
 {
-    int ret = real_gettimeofday(tv, __tz);
+    int ret = real_gettimeofday(tv, tz);
 
     int64_t sec_delta = TV_SEC_DELTA;
     int64_t nsec_delta = TV_NSEC_DELTA;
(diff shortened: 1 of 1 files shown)


STRRL:
PTAL

/cc @ReigenDing @YangKeao @iguoyr

pkg/time/fakeclock/fake_gettimeofday.c
```diff
@@ -24,36 +24,37 @@ extern int64_t TV_SEC_DELTA;
...
         : "memory");
 
     return ret;
 }
 
 #elif defined(__aarch64__)
-inline int real_gettimeofday(struct timeval *tv)
+inline int real_gettimeofday(struct timeval *tv, struct timezone *tz)
 {
     register struct timeval *x0 __asm__("x0") = tv;
```
iguoyr, line 42:
```
./pkg/time/fakeclock/fake_gettimeofday.c:50:12: warning: returning 'struct timeval *' from a function with return type 'int' makes integer from pointer without a cast [-Wint-conversion]
     return x0;
            ^~
```
I think the type of `x0` should be fixed too :)

iguoyr (commented):
Rest LGTM

pkg/time/fakeclock/fake_gettimeofday.c
YangKeao: The problem is that the `x0` is both the first argument and the return value. Maybe it should be converted to `int` after calling the system call 🤔 .

STRRL: It would give another warning `warning: cast from pointer to integer of different size [-Wpointer-to-int-cast]` when I trying to convert it to `int`. What should I do? How about introducing another new variable `x2` (as using the same register) for receiving return value. I am NOT sure about using `x2` in syscall would break something convention.

STRRL: PTAL @YangKeao

YangKeao (approved):
LGTM

YangKeao:
> > LGTM
> 
> What should I do with compiler warning?

As I have tried on my raspberry pi, it can be solved by adding a new `register int w0 __asm__("w0")`. The result assembly will be really simple:

```
mov x8, #0xa9
svc #0x0
ret
```

STRRL:
> > > LGTM
> > 
> > 
> > What should I do with compiler warning?
> 
> @STRRL As I have tried on my raspberry pi, it can be solved by adding a new `register int w0 __asm__("w0")`. The result assembly will be really simple:
> 
> ```
> mov x8, #0xa9
> svc #0x0
> ret
> ```

I do not get the point, it seems it contains many conventions about syscall and assembly that I do not know. 🥲

Could you help me with doing that( by directly pushing codes to this PR)? :P

YangKeao:
@STRRL  Done.

I'm not sure whether it will actually work, because the `ptrace` related package doesn't provide an arm implementation, so I cannot verify this implementation through watchmaker.

iguoyr:
> @STRRL Done.
> 
> I'm not sure whether it will actually work, because the `ptrace` related package doesn't provide an arm implementation, so I cannot verify this implementation through watchmaker.

Right! Actually we can't execute timechaos experiments on arm64 right now, so in order to fix the CI ASAP, I thought I'd approve it first and fix this feature later.

iguoyr (approved):
LGTM

STRRL:
> [`4d3bfed`](/chaos-mesh/chaos-mesh/commit/4d3bfeda0f2c68c0416cfbf06c132191f751af6a)

It makes no sense, I could manually cherry-pick it with no conflict.

PR #2893, opened 2022-02-11
temporarily use global logger in pkg/chaosdaemon

STRRL:
### What problem does this PR solve?

It's a part of https://github.com/chaos-mesh/chaos-mesh/issues/2872

### What's changed and how it works?

- remove `log = ctrl.Log.WithName("chaos-daemon")`
- replace with `log.L().withname("<logger-name>")`
 

### Related changes

- [ ] Need to update `chaos-mesh/website`
- [ ] Need to update `Dashboard UI`
- Need to **cheery-pick to release branches**
  - [x] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code
- [x] Manual test (add steps below)

build succeed and played with several experiments in `/example`

Side effects

- [ ] Breaking backward compatibility

### Release note 

```text
Please add a release note.

You can safely ignore this section if you don't think this PR needs a release note.
```

### DCO

If you find the DCO check fails, please run commands like below (Depends on the actual situations. For example, if the failed commit isn't the most recent) to fix it:

```shell
git commit --amend --signoff
git push --force
```

PR #2902, opened 2022-02-16
log: add customized logger for daemon and bpm

YangKeao:

The logger inside chaosdaemon and bpm is modified to print the `namespacedName` of the corresponding chaos resources.

1. A client interceptor was added to the daemon grpc client to automatically install the `namespacedName` into the request.
2. Modify every logger in chaos daemon and bpm to print the `namespacedName` in the context.

I and @STRRL have discussed the tracing under https://github.com/chaos-mesh/rfcs/pull/28#issuecomment-1000073106 , but the situation is more complicated than we thought. The context (and its `namespacedName`) will be passed across many components (like bpm) and structs (like `tcClient`, `processBuilder`). Building a factory for them isn't good (or every function with log will be equipped with a factory in the future, which is definitely unacceptable).

Then it turns back to the discussion: one two choices left: passing log everywhere, or constructing log from the context.

In this PR, I tried to construct log from the context, which seems fine. If you have better idea, feel free to discuss under this PR.

Conflict with #2893


STRRL:
I just wonder if we put NamespacedName into context, and carried by tracing system(like opentelemetry). 

Then we build another instrumentation on logr like https://github.com/uptrace/opentelemetry-go-extra/tree/main/otelzap, it would also resolve problems like "logging trace id or span id". 🤔

And the codes for logging would like 

```
type Service struct {
   ...
  logger instrumentedlogr.Logger
}

func (s *Service) SomeMethod(ctx context, other, parameters string)  {
    ...
    s.logger.WithContext(ctx).Info("key1", other, "key2", string);
    ...
}
```

What do you think about it?

YangKeao:
@STRRL 

> Then we build another instrumentation on logr like https://github.com/uptrace/opentelemetry-go-extra/tree/main/otelzap, it would also resolve problems like "logging trace id or span id". 🤔 (Resolving our issue as resolving logging "context-with-values from opentelemetry" )

It sounds good. But it seems that the main function of this package is to insert log events into the span, but not printing the information in the span to the log, so only the `trace_id` will be printed (I haven't tried, but the briefly the codes told me so: https://github.com/uptrace/opentelemetry-go-extra/blob/main/otelzap/otelzap.go#L170) . I wonder whether the `trace_id` is what we needed. 

Only with `trace_id`, we should look into the `controller-manager` and then filter the log in chaos-daemon with `trace_id`., but with `namespacedName`, we can filter the corresponding log (in chaos-daemon) directly. I cannot say which one is better. The former one is definitely more accurate (as it will contain only one request), but the latter one is more convenient (in the current situation), while the `trace_id` is also nice to have.

If both of the `trace_id` and `namespacedName` are needed, we can integrate with the `otelzap` in the future (another PR).

STRRL:
> But it seems that the main function of this package is to insert log events into the span, but not printing the information in the span to the log, so only the trace_id will be printed 

Yes, that's expected.

> Only with trace_id, we should look into the controller-manager and then filter the log in chaos-daemon with trace_id.

With the help of distribution systems like Zipkin/Jaeger, it would be much easier to profile the whole life of span. And the experience of profiling/debugging application with tracing is much better than logging.

I agree with "namespacedName" is much more convenient. 👍

I still have another question, why do you choose to use a grpc interceptor instead of modifying the protocol, passing the `NamespacedName` as a parameter directly?

YangKeao:
@STRRL 

> I still have another question, why do you choose to use a grpc interceptor instead of modifying the protocol, passing the NamespacedName as a parameter directly?

Simply because I don't want to modify every message and grpc call manually 😢. And the `namespacedName` doesn't affect the running logic of the rpc call.

As I know, the normal tracing solution also passes the span context through the interceptor, which is similar to this scenario.

If we establish a strong tracing infrastructure in the future, and the `namespacedName` is not needed anymore, it's also convenient to remove those interceptors.

YangKeao:
> > But it seems that the main function of this package is to insert log events into the span, but not print the information in the span to the log, so only the trace_id will be printed

> Yes, that's expected.

@STRRL  I wonder whether we can modify this library to also print the attributes? Or is it suggested? If it is, I think it would be better than passing the `namespacedName` manually, though it needs more work.

Hexilee:
>  I wonder whether we can modify this library to also print the attributes? Or is it suggested? If it is, I think it would be better than passing the namespacedName manually, though it needs more work.

@YangKeao  Could we customize the trace_id? For example, use `<Namespace>/<Name>-<uuid>`.

STRRL:
> > > But it seems that the main function of this package is to insert log events into the span, but not print the information in the span to the log, so only the trace_id will be printed
> 
> > Yes, that's expected.
> 
> @STRRL I wonder whether we can modify this library to also print the attributes? Or is it suggested? If it is, I think it would be better than passing the `namespacedName` manually, though it needs more work.

I am not sure about it's the best practice to log all the attributes in the span. 🤣

STRRL:
> > I wonder whether we can modify this library to also print the attributes? Or is it suggested? If it is, I think it would be better than passing the namespacedName manually, though it needs more work.
> 
> @YangKeao Could we customize the trace_id? For example, use `<Namespace>/<Name>-<uuid>`.

I think carrying data into trace_id is not a good idea.

pkg/chaosdaemon/iochaos_server.go
```diff
@@ -130,7 +132,7 @@ func (s *DaemonServer) ApplyIOChaos(ctx context.Context, in *pb.ApplyIOChaosRequ
 	}, nil
 }
 
-func (s *DaemonServer) killIOChaos(ctx context.Context, pid int64, startTime int64) error {
+func (s *DaemonServer) killIOChaos(ctx context.Context, log logr.Logger, pid int64, startTime int64) error {
```
STRRL, line 135: That's a little weird, we could access two loggers inside method `killIOChaos`.

I prefer to also call `getLoggerFromContext` inside of `killIOChaos`

controllers/utils/chaosdaemon/chaosdaemon.go
```diff
@@ -71,7 +71,7 @@ func (b *ChaosDaemonClientBuilder) FindDaemonIP(ctx context.Context, pod *v1.Pod
 	return daemonIP, nil
 }
 
-func (b *ChaosDaemonClientBuilder) Build(ctx context.Context, pod *v1.Pod) (chaosdaemonclient.ChaosDaemonClientInterface, error) {
```
STRRL, line 74: How about appending comments for the parameter `id`?

STRRL (commented):
rest LGTM!

pkg/bpm/build_linux.go
```diff
@@ -25,7 +25,11 @@ import (
 )
 
 // Build builds the process
-func (b *ProcessBuilder) Build() *ManagedProcess {
+// the ctx argument passes the context information to this function
+// e.g. the corresponding resource name
+func (b *ProcessBuilder) Build(ctx context.Context) *ManagedProcess {
```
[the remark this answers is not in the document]
YangKeao: They have different meanings (and that's why `context.Context` sucks 😢). The `context.Context` in the `ProcessBuilder` is used to create a command with the context, if the context is closed, the process will be killed.

For `toda` and `chaos-tproxy`, their lifetime is much longer than the request, so it's not expected to set a context for them. The `context.Context` in this parameter is only used to log.

pkg/chaosdaemon/iochaos_server.go
YangKeao: done

pkg/chaosdaemon/util.go
```diff
@@ -112,6 +115,10 @@ func GetChildProcesses(ppid uint32) ([]uint32, error) {
 	}
 }
 
+func (s *DaemonServer) getLoggerFromContext(ctx context.Context) logr.Logger {
```
Andrewmatilde, line 118: Is there a direct way to let developers know what is inside the ctx?
I have tried to find whether a ctx from time_server.SetTimeOffset have the namespase , but found a lot of interface with Client.Get(ctx,...), I think it is too hard to know it.

YangKeao: The developer has to know where the context comes from. If the context is created by grpc-server (for every request), it's equipped with the `namespacedName` (automatically, with the help of grpc server-side interceptor).

Andrewmatilde: I suggest to add a comment about where it comes from add what will happen.

pkg/bpm/build_linux.go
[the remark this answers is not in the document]
YangKeao: Done.

pkg/chaosdaemon/util.go
STRRL: > Is there a direct way to let developers know what is inside the ctx?

We(the provider of module/lbrary `DaemonServer`) could not assume that more things are carried with `context.Context`, because it's out of the scope of `context.Context`. 

The current `log.EnrichLoggerWithContext` works in a "best-effort pattern", it would append more information when context brings more information and would not be panic if it doesn't, which looks great.

pkg/bpm/build_linux.go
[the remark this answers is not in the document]
STRRL: > They have different meanings (and that's why context.Context sucks 😢).

I think what actually and finally needs is a tracing `span`, not a `context.Context`. A span would propagate with `context.Context`, and could "manually propagate" with another `context.TODO()`. But context could not only "inherts" `Value(key interface{}) interface{}` without `Done() <-chan struct{}` from it's parents.

STRRL (approved):
LGTM!

Looking forward to one day that we could use tracing facilities.

pkg/chaosdaemon/util.go
Andrewmatilde: > The current log.EnrichLoggerWithContext works in a "best-effort pattern", it would append more information when context brings more information and would not be panic if it doesn't, which looks great.

But in most of package , I wish they log with chaos name forever. Like in http-chaos I want to ensure that logger wiil log with "HTTP-Chaos"

pkg/chaosdaemon/util/util.go
```diff
@@ -78,7 +79,7 @@ func GetChildProcesses(ppid uint32, logger logr.Logger) ([]uint32, error) {
 
 				reader, err := os.Open(statusPath)
 				if err != nil {
-					logger.Error(err, "read status file error", "path", statusPath)
+					log.L().Error(err, "read status file error", "path", statusPath)
```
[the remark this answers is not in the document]
YangKeao: ignore

pkg/chaosdaemon/util.go
YangKeao: @Andrewmatilde There is no way, at least in golang, to ensure the content of context in the compile-time. You can never avoid a bad guy to pass `context.TODO()` or `context.Background()` to these functions. What we could do is to find the problem as early as possible (during the review, or when we find the log is incorrect) and fix them.

Makefile
```diff
@@ -308,7 +308,7 @@ generate-deepcopy: images/dev-env/.dockerbuilt chaos-build
 
 generate: generate-ctrl swagger_spec generate-deepcopy chaos-build
 
-check: generate yaml vet boilerplate lint tidy install.sh fmt
+check: generate yaml vet lint tidy install.sh fmt
```
iguoyr, line 311: Why remove the prerequisite `boilerplate` from the target `check` here?

YangKeao: Oops. It's a misoperation. Added back.

pkg/chaosdaemon/util.go
Hexilee: Why not a concrete context type? For example, we defined a concrete context `Ctx`, which implements interface 
`context.Context`. For the functions that need special content, we can receive `*Ctx` in parameter; and for the functions that only need common context, we just pass `*Ctx` as `context.Context`

YangKeao: @Hexilee Using a concrete context `Ctx` means we couldn't use `context.With*` group of functions, which will return a type different from the beginning `Ctx` (see [context.WithValue](https://cs.opensource.google/go/go/+/refs/tags/go1.17.7:src/context/context.go;l=523)), and I don't want to put this limitation on the Chaos Mesh codes.

It has nearly the same problem with `error`. We need a `context.As(ctx, &Ctx)` to extract a `Ctx` from a context with an arbitrary type (but it doesn't solve any problem, it could also fail if it doesn't have a `Ctx` inside).

YangKeao: This suggestion is a little interesting to think further. If the golang can define such interface:

```go
type Context interface {
    func WithValue(self, key, val interface{}) Self;
}
```

Then it would be fine. But it cannot! (see the `Self` type, which is [not implemented](https://github.com/golang/go/issues/28254))

Hexilee: I think following function also meets our requirements:

```go
func WithValue[C Context](ctx C, key, value interface{}) C {

}
```

Maybe we can consider it in go-1.18 or later versions.

Hexilee: > I think following function also meets our requirements:
> 
> ```go
> func WithValue[C Context](ctx C, key, value interface{}) C {}
> ```
> 
> Maybe we can consider it in go-1.18 or later versions.

My fault, it cannot be implemented in the stdlib. How about the following code 😂.

```
type Context interface {
    func WithValue(ret *Context, key, val interface{});
}
```

YangKeao:
```go
type Context interface {
    WithValue(ret *Context, key, val interface{});
}
```

Then the beautiful pure function becomes an ugly dirty function that mutates the object itself 😿 .

Hexilee: > > ```go
> > type Context interface {
> >    WithValue(ret *Context, key, val interface{});
> > }
> > ```
> 
> Then the beautiful pure function becomes an ugly dirty function that mutates the object itself 😿 .

There may be some misunderstanding. Here are my examples:

```go
func doSomething(ctx *Ctx) {
    var newCtx Ctx
    ctx.WithValue(&newCtx, "name", "foo")
    go someFunc(&newCtx)
}
```

I think it's an acceptable API in the C programming language 😂.

STRRL: > But in most of package , I wish they log with chaos name forever. Like in http-chaos I want to ensure that logger wiil log with "HTTP-Chaos"

IMO, if you want **always** log the "chaos name", it means the "chaos name" is the required parameter. Please change the function signature, and append the new parameter `chaosName string` explicity.

Andrewmatilde: > > But in most of package , I wish they log with chaos name forever. Like in http-chaos I want to ensure that logger wiil log with "HTTP-Chaos"
> 
> IMO, if you want to **always** log the "chaos name", it means the "chaos name" is the required parameter. Please change the function signature, and append the new parameter `chaosName string` explicity.

I agree with you.

Andrewmatilde (approved):
LGTM

Issue #2751, opened 2022-01-06
JVMChaos: byteman inject fail when java pid is not 1


imlonghao:
## Bug Report

**What version of Kubernetes are you using?**

```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.10", GitCommit:"8152330a2b6ca3621196e62966ef761b8f5a61bb", GitTreeState:"clean", BuildDate:"2021-08-11T18:06:15Z", GoVersion:"go1.15.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.3", GitCommit:"01849e73f3c86211f05533c2e807736e776fcf29", GitTreeState:"clean", BuildDate:"2021-02-17T12:35:49Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
```

**What version of Chaos Mesh are you using?**

```
Controller manager Version: version.Info{GitVersion:"v2.1.2", GitCommit:"24a1f565ed000ba8738a4ed568b4e9a0925fe2fb", BuildDate:"2021-12-29T11:43:40Z", GoVersion:"go1.16.7", Compiler:"gc", Platform:"linux/amd64"}
```

**What did you do?**

Create a Java deployment, https://github.com/WangXiangUSTC/byteman-example/tree/main/example.helloworld

But change the Dockerfile to the following one

```
FROM openjdk:16-alpine3.13
COPY Main.jar Main.jar
COPY Manifest.txt Manifest.txt
ENV TINI_VERSION v0.19.0
ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
RUN chmod +x /tini
ENTRYPOINT ["/tini", "--"]
CMD ["java", "-jar", "Main.jar"]
```

Then create a JVMChaos experiment

**What did you expect to see?**

Byteman inject success and JVMChaos experiment run as expected

**What did you see instead?**

Byteman inject failed and JVMChaos stuck in injecting phase.

`kubectl describe JVMChaos xxx`

```
Events:
  Type     Reason           Age   From          Message
  ----     ------           ----  ----          -------
  Normal   FinalizerInited  90s   finalizers    Finalizer has been inited
  Normal   Updated          90s   finalizers    Successfully update finalizer of resource
  Normal   Paused           90s   desiredphase  Experiment has been paused
[...]

PR #2904, opened 2022-02-16
fix(JVMChaos): find the correct pid by CommName

imlonghao:
Signed-off-by: imlonghao <git@imlonghao.com>

https://github.com/chaos-mesh/chaos-mesh/issues/2751

Read all the CommName from container, if the pid's CommName
is `java\n` then it is the correct pid to be injected.

### Related changes

  - [x] release-2.1
  - [x] release-2.0

It is a bug fix so I think it need to be backported.

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

1. Build the chaos-daemon
2. Deploy a java Pod which java pid is not 1
3. Create a JVMChaos and see whether the inject is work

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/pkg/chaosdaemon/jvm_server.go
+++ b/pkg/chaosdaemon/jvm_server.go
@@ -43,6 +43,24 @@ func (s *DaemonServer) InstallJVMRules(ctx context.Context,
 		return nil, err
 	}
 
+	containerPids := []uint32{pid}
+	childPids, err := GetChildProcesses(pid)
+	if err != nil {
+		log.Error(err, "GetChildProcesses")
+	}
+	containerPids = append(containerPids, childPids...)
+	for _, containerPid := range containerPids {
+		name, err := ReadCommName(int(containerPid))
+		if err != nil {
+			log.Error(err, "ReadCommName")
+			continue
+		}
+		if name == "java\n" {
+			pid = containerPid
+			break
+		}
+	}
+
 	bytemanHome := os.Getenv("BYTEMAN_HOME")
 	if len(bytemanHome) == 0 {
 		return nil, errors.New("environment variable BYTEMAN_HOME not set")


pkg/chaosdaemon/jvm_server.go
```diff
@@ -43,6 +43,23 @@ func (s *DaemonServer) InstallJVMRules(ctx context.Context,
...
+	containerPids := []uint32{pid}
+	childPids, err := GetChildProcesses(pid)
+	if err != nil {
+		log.Error(err, "GetChildProcesses")
+	}
+	containerPids = append(containerPids, childPids...)
+	for _, containerPid := range containerPids {
+		name, err := ReadCommName(int(containerPid))
+		if err != nil {
+			continue
```
WangXiangUSTC, line 55: how about adding a log here?

WangXiangUSTC (commented):
rest LGTM, thanks!

pkg/chaosdaemon/jvm_server.go
```diff
@@ -43,6 +43,24 @@ func (s *DaemonServer) InstallJVMRules(ctx context.Context,
...
+		log.Error(err, "GetChildProcesses")
+	}
+	containerPids = append(containerPids, childPids...)
+	for _, containerPid := range containerPids {
+		name, err := ReadCommName(int(containerPid))
+		if err != nil {
+			log.Error(err, "ReadCommName")
+			continue
+		}
+		if name == "java\n" {
```
WangXiangUSTC, line 58: Is it possible that there are more than one java process in the container?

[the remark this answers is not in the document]
WangXiangUSTC: For this pr, I think injecting into the first java process is OK. Maybe we will add a new config like `process-name` later, and find a java process containing the `process-name`.

Issue #2891, opened 2022-02-11
Refine logging in `pkg/webhook/inject`, refactor these function as methods of a Struct, them move package level variable `log` as a field


STRRL:

afzal442:
@STRRL, I'm interested, could you make little more clear here which `logging` needs to be refactored?
Do you mean like this as below?
e.g.
https://github.com/chaos-mesh/chaos-mesh/blob/a2854f0e1a74130f8cb304b9036fea0d5eb91c4c/pkg/webhook/inject/command.go#L23
```go
type mergeCmd struct {
inject []string
origin []string 
args []string
}
func (mg mergeCmd)MergeCommands() []string {

```

STRRL:
Hi @afzal442! Thanks for your attention on Chaos Mesh! 

Well, the refactor pattern which you used is excelly correct, but this issue only want to refactor about functions which have used `log`, for example, `Inject(res *v1.AdmissionRequest, cli client.Client, cfg *config.Config, controllerCfg *controllerCfg.ChaosControllerConfig, metrics *metrics.ChaosControllerManagerMetricsCollector) *v1.AdmissionResponse ` in file `pkg/webhook/inject/inject.go`.

If you think other files also need refactor (and it does actually need), it's better to make another issue and PR for that. :D

afzal442:
thanks @STRRL, for your suggestion. But I wondered how I can move package level variable `log` as a field.

Amshu-k:
@g1eny0ung - could you please help me with this issue?
I'm looking at the codebase and this directory `pkg/webhook/inject` doesn't exist anymore? Am I missing something here?

g1eny0ung:
@Amshu-k I've found that this issue is outdated because the `inject.go` was removed in https://github.com/chaos-mesh/chaos-mesh/pull/3886, I forgot about this PR, so I reopened this issue again. You don't miss anything. I'll go ahead and close this issue now.

PR #2917, opened 2022-02-18
Refactor `pkg/webhook/inject` to use pkg level log

afzal442:
Signed-off-by: afzal442 <afzal442@gmail.com>

Closes: #2891 

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


pkg/webhook/inject/inject.go
```diff
@@ -51,8 +60,14 @@ const (
 // Inject do pod template config inject
 func Inject(res *v1.AdmissionRequest, cli client.Client, cfg *config.Config, controllerCfg *controllerCfg.ChaosControllerConfig, metrics *metrics.ChaosControllerManagerMetricsCollector) *v1.AdmissionResponse {
```
Andrewmatilde, line 61: I perfer 
```
func (i *Injector)Inject(..)
```.

File: pkg/webhook/inject/inject.go [hunk 2]
```diff
@@ -51,8 +60,14 @@ const (
...
 	var pod corev1.Pod
+	i = Injector{
+		logger:    log.L(),
```
Andrewmatilde, line 64: I think changes here will lost info : "inject-webhook", in old logger.

pkg/webhook/inject/inject.go
afzal442: I too, thanks. 😄

pkg/webhook/inject/inject.go
afzal442:
```suggestion
		logger:    ctrl.Log.WithName("inject-webhook"),
```
What do I need ?

Andrewmatilde: It's OK.

pkg/webhook/inject/inject.go
```diff
@@ -25,24 +25,33 @@ import (
...
-var log = ctrl.Log.WithName("inject-webhook")
-
 var ignoredNamespaces = []string{
 	metav1.NamespaceSystem,
 	metav1.NamespacePublic,
 }
 
+type Injector struct {
+	logger    logr.Logger
+	err       error
```
STRRL, line 47: please remove this unused `err`.

pkg/webhook/inject/inject.go
```diff
@@ -25,24 +25,33 @@ import (
...
+	c         client.Client
```
STRRL, line 48: I prefer to use `kubeclient` instead of `c` :)

pkg/webhook/inject/inject.go
```diff
@@ -25,24 +25,33 @@ import (
...
+	cfg       *config.Config
+	ctrlerCfg *controllerCfg.ChaosControllerConfig
+}
+
+var i Injector
```
STRRL, line 53: please do not use the global value `i`

pkg/webhook/inject/inject.go
```diff
@@ -25,24 +25,33 @@ import (
...
 )
 
-var log = ctrl.Log.WithName("inject-webhook")
-
 var ignoredNamespaces = []string{
 	metav1.NamespaceSystem,
 	metav1.NamespacePublic,
 }
 
+type Injector struct {
```
STRRL, line 45: please create a constructor function for Injector

STRRL (changes requested):
After introduced the struct `Injector`, we should transform all the related functions as the method of struct `Injector`.

And consider using a factory pattern when initialize Inject with different `kubeclient` and other configurations but with the same `logger`.

pkg/webhook/inject/inject.go
```diff
@@ -34,25 +33,44 @@ import (
...
 )
 
+func NewInjector(kubeclient client.Client, cfg *config.Config, controllerCfg *controllerCfg.ChaosControllerConfig, metrics *metrics.ChaosControllerManagerMetricsCollector) *Injector {
+	return &Injector{
+		kubeclient:    kubeclient,
+		cfg:           cfg,
+		ControllerCfg: controllerCfg,
+		Metrics:       metrics,
+	}
+}
```
STRRL, line 64: please initialize logger in constructor

pkg/webhook/inject.go
```diff
@@ -32,6 +32,8 @@ import (
 
 var log = ctrl.Log.WithName("inject-webhook")
 
+var i inject.Injector
+
```
STRRL, line 36: please do not use this global variable `i`.

pkg/webhook/inject.go
```diff
@@ -55,7 +57,7 @@ func (v *PodInjector) Handle(ctx context.Context, req admission.Request) admissi
 	log.Info("Get request from pod:", "pod", pod)
 
 	return admission.Response{
-		AdmissionResponse: *inject.Inject(&req.AdmissionRequest, v.client, v.Config, v.ControllerCfg, v.Metrics),
+		AdmissionResponse: *i.Inject(&req.AdmissionRequest, v.client, v.Config, v.ControllerCfg, v.Metrics),
```
STRRL, line 60: do not use global value `i`.

consider calling the constructor to instant new `Injector` instance.

 you could modify the fields of `webhook.PodInjector`, appending new field `logger`, so you could access logger when you instant new `Injector` instance.

pkg/webhook/inject/inject.go
```diff
@@ -34,25 +33,44 @@ import (
...
+
 // Inject do pod template config inject
-func Inject(res *v1.AdmissionRequest, cli client.Client, cfg *config.Config, controllerCfg *controllerCfg.ChaosControllerConfig, metrics *metrics.ChaosControllerManagerMetricsCollector) *v1.AdmissionResponse {
+func (i Injector) Inject(res *v1.AdmissionRequest, cli client.Client, cfg *config.Config, controllerCfg *controllerCfg.ChaosControllerConfig, metrics *metrics.ChaosControllerManagerMetricsCollector) *v1.AdmissionResponse {
```
STRRL, line 67: please change the signature of method `Inject`, as `Injector` already contains fields with type `client.Cient`, `config.Config`. You could use them instead of acquiring new methods parameters.

STRRL, line 67: please use pointer receiver instead of value receiver

pkg/webhook/inject/inject.go
```diff
@@ -137,55 +155,55 @@ func Inject(res *v1.AdmissionRequest, cli client.Client, cfg *config.Config, con
 }
 
 // Check whether the target resource need to be injected and return the required config name
-func injectRequired(metadata *metav1.ObjectMeta, cli client.Client, cfg *config.Config, controllerCfg *controllerCfg.ChaosControllerConfig) (string, bool) {
+func (i Injector) injectRequired(metadata *metav1.ObjectMeta, cli client.Client, cfg *config.Config, controllerCfg *controllerCfg.ChaosControllerConfig) (string, bool) {
```
STRRL, line 158: ditto, please use pointer receiver instead of value receiver.

please use fields inside of `Injector` instead of acquiring them as method parameters

STRRL (changes requested):
PTAL

STRRL:
Hi @afzal442 do you need some help?

pkg/webhook/inject/inject.go
```diff
@@ -25,34 +25,55 @@ import (
...
+		kubeclient:    kubeclient,
+		cfg:           config,
+		ControllerCfg: controllerCfg,
+		Metrics:       metrics,
+	}
+}
+
 // Inject do pod template config inject
-func Inject(res *v1.AdmissionRequest, cli client.Client, cfg *config.Config, controllerCfg *controllerCfg.ChaosControllerConfig, metrics *metrics.ChaosControllerManagerMetricsCollector) *v1.AdmissionResponse {
+func (i *Injector) Inject(logger logr.Logger, res *v1.AdmissionRequest, cfg *config.Config, cli client.Client, controllerCfg *controllerCfg.ChaosControllerConfig, metrics *metrics.ChaosControllerManagerMetricsCollector) *v1.AdmissionResponse {
```
Andrewmatilde, line 70: I think the common serial of program is :
1. i := NewInjector(...)
2. i.Inject(...)

And then you can remove the first logger param here.

afzal442:
Any clue why unit test failing here?

cc @Andrewmatilde @STRRL

STRRL:
> Any clue why unit test failing here?
> 
> cc @Andrewmatilde @STRRL

There are some unstable/flaky tests in the unit tests, I would rerun them.

pkg/webhook/inject/inject_test.go
```diff
@@ -27,14 +27,16 @@ import (
 	"github.com/chaos-mesh/chaos-mesh/pkg/webhook/config"
 )
 
+var i Injector
```
STRRL, line 30: do not introduce a new package level variable in the test. instant new one in the test codes.

STRRL: You could initialize a new Injector between `Describe` and `Context` statements.

pkg/webhook/inject/inject.go
```diff
@@ -309,12 +329,12 @@ func createPatch(pod *corev1.Pod, inj *config.InjectionConfig, annotations map[s
...
 	// TODO: remove injecting commands when sidecar container supported
 	// set commands and args
-	patch = append(patch, setCommands(pod.Spec.Containers, inj.PostStart)...)
+	patch = append(patch, i.setCommands(pod.Spec.Containers, inj.PostStart)...)
 
 	return json.Marshal(patch)
 }
 
-func setCommands(target []corev1.Container, postStart map[string]config.ExecAction) (patch []patchOperation) {
+func (i Injector) setCommands(target []corev1.Container, postStart map[string]config.ExecAction) (patch []patchOperation) {
```
STRRL, line 337: please use a pointer as method receiver for **all** methods.

STRRL (changes requested):
Hi @afzal442 PTAL.

Andrewmatilde (approved):
LGTM

Andrewmatilde:
PTAL , @STRRL

STRRL:
Hi @afzal442, please resolve the conflicts

pkg/webhook/inject.go
```diff
@@ -33,6 +33,12 @@ import (
 // +kubebuilder:webhook:path=/inject-v1-pod,mutating=false,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=vpod.kb.io
 
 // PodInjector is pod template config injector
+
+var (
+	i   *inject.Injector
+	log logr.Logger
+)
```
STRRL, line 40: do not use global variable

STRRL (changes requested):
Hi @afzal442, let me talk more details about what should we do:

We should apply one of the refactoring patterns called "refactor combine function into struct/class", reference: https://refactoring.com/catalog/combineFunctionsIntoClass.html

You could see that there are many (exported and unexported) functions under `pkg/webhoook/inject`, and some part of the arguments are same, for example:

- `cli client.Client`
- `cfg *config.Config`
- `ontrollerCfg *controllerCfg.ChaosControllerConfig`
- and so on

We should extract them into a struct (in other language like java, it is class), then make the origin function as the methods of the struct. And also provide a constructor of struct.

So we could use "calling the constructor to build a new instance then call the method" instead of invoking the original function.

/cc @Andrewmatilde

STRRL:
> Hi @afzal442, let me talk more details about what should we do:
> 
> We should apply one of the refactoring patterns called "refactor combine function into struct/class", reference: [refactoring.com/catalog/combineFunctionsIntoClass.html](https://refactoring.com/catalog/combineFunctionsIntoClass.html)
> 
> You could see that there are many (exported and unexported) functions under `pkg/webhoook/inject`, and some part of the arguments are same, for example:
> 
> * `cli client.Client`
> * `cfg *config.Config`
> * `ontrollerCfg *controllerCfg.ChaosControllerConfig`
> * and so on
> 
> We should extract them into a struct (in other language like java, it is class), then make the origin function as the methods of the struct. And also provide a constructor of struct.
> 
> So we could use "calling the constructor to build a new instance then call the method" instead of invoking the original function.
> 
> /cc @Andrewmatilde

Hi @afzal442, if you have another question, feel free to leave comments :)

pkg/webhook/inject/inject_test.go
```diff
@@ -24,17 +26,23 @@ import (
...
+	var i Injector
 
 	Context("Inject", func() {
 		It("should return unexpected end of JSON input", func() {
 			var testClient client.Client
 			var cfg *config.Config
 			var controllerCfg *controllerCfg.ChaosControllerConfig
-			res := Inject(&admissionv1.AdmissionRequest{}, testClient, cfg, controllerCfg, nil)
+			var metrics *metrics.ChaosControllerManagerMetricsCollector
+			var log logr.Logger
```
STRRL, line 42: please initialize this `log`, you could take a look on here as reference:

https://github.com/chaos-mesh/chaos-mesh/blob/master/cmd/chaos-controller-manager/provider/suit_test.go#L81-L82

pkg/webhook/inject/inject_test.go
```diff
@@ -24,17 +26,23 @@ import (
...
+
+			ij := NewInjector(testClient, cfg, controllerCfg, metrics, log)
+			res := ij.Inject(&admissionv1.AdmissionRequest{})
```
STRRL, line 45: Maybe we could also use `i` at line 34 in this testcase.

pkg/webhook/inject/inject_test.go
```diff
@@ -24,17 +26,23 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	controllerCfg "github.com/chaos-mesh/chaos-mesh/pkg/config"
+	"github.com/chaos-mesh/chaos-mesh/pkg/metrics"
 	"github.com/chaos-mesh/chaos-mesh/pkg/webhook/config"
 )
 
 var _ = Describe("webhook inject", func() {
+	var i Injector
```
STRRL, line 34: please initialize this log with constructor, only declaring without initializing makes no sense

STRRL: maybe you need to use different `Injector`s for different testcases.

STRRL (changes requested):

PR #2928, opened 2022-02-23
feat: support cgroup v2 for linux stress experiments

### LINKED ISSUE #2652: Failed to apply chaos: rpc error: code = Unknown desc = controller is not supported
## Bug Report

1.14.3

2.1.0 made by myself
**What did you do?**

add  stresschaos
**What did you see instead?**
image

**Output of chaosctl**

yaml
image

igcherkaev:

Support cgroup v2 driver when running experiments on linux (CPU & Mem). If the host/containers where chaosdaemon is running is in full cgroup v2 mode (unified), users will no longer see:
```
Failed to apply chaos: rpc error: code = Unknown desc = controller is not supported
```
When they run CPU or Memory stress experiments.

Introduced a cgroupv2 manager to add newly created stress processes to the respective cgroup if containers based on what cgroup driver containers are using.

P.S. I did not re-order import statements myself, it must have been done by the format/lint/tidy goals in the Makefile when I ran make check or make test. I can revert these changes, if need be.

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code
- [x] Manual test
  - Start kubernetes 1.21+ with cgroupfs and cgroup v2 driver.
  - Verify what cgroup driver you have on the node and in the container:

    ```bash
    # container
    $ kubectl -n test-dev exec -ti demo-c7c688b8-trr42 -- bash
    bash-4.2# mount| grep cgroup
    cgroup on /sys/fs/cgroup type cgroup2 (ro,nosuid,nodev,noexec,relatime,seclabel)

    # kubernetes node
    lvdkbw503 ~ # mount| grep cgroup
    cgroup2 on /sys/fs/cgroup type cgroup2 (rw,nosuid,nodev,noexec,relatime,seclabel)
    ```

  - Schedule new CPU/Mem stress experiment and verify it's working by checking cgroup assigned to respective processes on the worker nodes.
  - Verify it on the node by using pids of the main process in the container and of the stress-ng/memstress process:

    ```bash
     lvdkbw503 ~ # ps -o cgroup 1476124
     CGROUP
     0::/kubepods/burstable/pod8731e44b-df00-41f9-9c50-
     1c6327a16b74/69604c2dd135683bf0758cf623c05c37f6fe61ab437fd2227b0f584e93eedbc0
     lvdkbw503 ~ # ps -o cgroup 2188035
     CGROUP
     0::/kubepods/burstable/pod8731e44b-df00-41f9-9c50-
     1c6327a16b74/69604c2dd135683bf0758cf623c05c37f6fe61ab437fd2227b0f584e93eedbc0
    ```

Side effects

N/A

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


YangKeao:
This PR cannot close #2652, because the problem in #2652 is actually not for "Unified" mode, but "Hybrid" mode. In https://github.com/chaos-mesh/chaos-mesh/issues/2652#issuecomment-999300301 , only the `blkio` is handled by the cgroupv2 controller.

But it's still fine to support unified group v2 only, for now.

igcherkaev:
> This PR cannot close #2652, because the problem in #2652 is actually not for "Unified" mode, but "Hybrid" mode. In [#2652 (comment)](https://github.com/chaos-mesh/chaos-mesh/issues/2652#issuecomment-999300301) , only the `blkio` is handled by the cgroupv2 controller.
> 
> But it's still fine to support unified group v2 only, for now.

Oh, sorry, I did not know that. @STRRL was talking about cgroup v2 there so I assumed the unified mode. I don't have environment with cgroups in the hybrid mode, and I don't know the root cause of the issue in #2652 and why it's related to blkio, so I'll leave it out of the scope of this PR then.

igcherkaev:
Also, I can backport this fix to `release-2.1` branch with a separate PR - cherry-picking won't work since function signatures have changed.

STRRL:
I tested this patch for different container runtime; it works on contained but does not work on docker.

It seems docker would isolate the "cgroup namespace" for the container but containerd (also crio) would not.

When we call `cgroupsv2.PidGroupPath(<pid>)`, chaos-daemon (in containerd) would return:

```
0::/kubepods/besteffort/poddc367d15-b0dd-4401-86de-84fd7ba7e08b/4d287d5b0651c1bd13ccaecdcfe6f1a65379150a4ec4ae6d46c392b76b812941
```

But chaos-daemon( in docker) would return 

```
0::/../../kubepods-besteffort-pod25df87b9_2545_47fb_94ba_3292ea427f95.slice/docker-41c1b86aa6257d0cd560db3c99c53f632c4909774280dcc5c57a544fde52a340.scope
```

The latter one would not work well. 

Should we also mount the hosts /proc into chaos-daemon like /host-sys? @YangKeao

igcherkaev:
Should I add a condition to apply it only when runtime is containerd? Docker runtime is deprecated in k8s anyway, maybe Chaos Mesh should do the same and begin deprecating docker runtime?

STRRL:
> Should I add a condition to apply it only when runtime is containerd? Docker runtime is deprecated in k8s anyway, maybe Chaos Mesh should do the same and begin deprecating docker runtime?

I think this PR would not require that. I think the support for docker should be finished in the next several PRs.

STRRL:
Hi @igcherkaev , could you resolve the conflicts? Then appending an entry in the `CHANGELOG.md`

igcherkaev:
@STRRL sure, will work on that.

STRRL:
Hi @igcherkaev, the `verify` check failed. `verify` relates to codes linters and code generation. You could execute `make check` on your local machine then commit the changes to fix it.

igcherkaev:
> Hi @igcherkaev, the `verify` check failed. `verify` relates to codes linters and code generation. You could execute `make check` on your local machine then commit the changes to fix it.

Hmm, I thought I'd done that. Did it again and pushed the changes.

YangKeao (approved):
LGTM

STRRL (approved):
LGTM! 

Thanks for your contribution!

igcherkaev:
I think DCO check fails due to rebasing my branch and including commits made by other folks.

STRRL:
> I think DCO check fails due to rebasing my branch and including commits made by other folks.

I would manually set it to PASS. 🥲

PR #2932, opened 2022-02-23
NetworkChaos: Add support for ports in external targets

miedzinski:
Signed-off-by: Dominik Miedziński <dominik.miedzinski@allegro.pl>

This PR adds support for declaring ports in external targets in NetworkChaos experiments.

Chaos-daemon now creates 3 IP sets in pods: `hash:net` (same as before), `hash:net,port` (allowing for testing ip and port pairs) and `list:set` that combines them. Iptables rules now reference the `list:set` IP set.

There are changes to PodNetworkChaos, but NetworkChaos remains unchanged and there's no need to update dashboard.

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Add support for declaring ports in external targets in NetworkChaos experiments.
```


STRRL:
Hi @miedzinski , thanks for your contribution!  That's an awesome feature!

But this PR is really huge, and it would still break the API of Chaos Mesh (with podnetworkchaos). So I am afraid this PR would not be merged recently. 

Currently, the design of NetworkChaos does not care about the port thing, it is designed on the Network Layer(IP Layer). If we want to filter the network traffic with ports(actually it's not difficult with iptables and current code base), we should change our design with at least Transport Layer(consider some protocol, TCP/UDP/...). It would make huge changes to the design of NetworkChaos. I think we would better draft an RFC into [rfcs](https://github.com/chaos-mesh/rfcs) before we do that.

How do you think about it? @miedzinski

miedzinski:
@STRRL, thanks for your response.

You're right there is a problem with backward compatibility of PodNetworkChaos, but I think it's something we can solve:

- There is no need to name all IP sets in PodNetworkChaos. `hash:net` and `hash:net,port` names can be derived in daemon from `list:set` name.
- IP set contents can remain strings and be parsed later.

This way all CRDs are unchanged. There is also a change to protobufs - if you think these shouldn't be changed too, then I can move parsing and resolving external targets to chaos daemon. That would additionally solve [a TODO comment](https://github.com/chaos-mesh/chaos-mesh/blob/b9e3519e8e1efea209655e9e0a23f24236062e83/controllers/podnetworkchaos/netutils/cidr.go#L33).

Regarding the RFC process, this is something I'd rather avoid if possible. Filtering traffic by ports is crucial for adoption of Chaos Mesh in my organization, but we definitely don't intend to make huge changes to design of experiments at this point. I don't think the L3 vs L4 is also an issue, because NetworkChaos already resolves hostnames and that's Application Layer.

Please let me know if you're okay with changes I suggested.

STRRL:
> You're right there is a problem with backward compatibility of PodNetworkChaos, but I think it's something we can solve:
> 
> * There is no need to name all IP sets in PodNetworkChaos. `hash:net` and `hash:net,port` names can be derived in daemon from `list:set` name.
> * IP set contents can remain strings and be parsed later.

> This way all CRDs are unchanged.


We only consider "backward compatibility" now. We could accept appending more fields with default values. But renaming/deleting fields is NOT preferred.

I have another idea based on your suggestion. We could extend `v1alpha1.RawIPSet` and `pb.IPSet` to support more type of `ipset`, instead of using one `v1alpha1.RawIPSet` to introduce the combination of `ipset` rules which contains `hash:ip`, `hash:IP,port` and `list:net`.

```go

const (
	IPSetTypeHashIP     IPSetType = "hash:ip"
	IPSetTypeHashIPPort IPSetType = "hash:ip,port"
	IPSetTypeListNet    IPSetType = "list:net"
	// other IPSet types that we need
)

// RawIPSet represents an ipset on specific pod
type RawIPSet struct {
	// The name of set ipset
	Name string `json:"name"`

	IPSetType IPSetType `json:"ipsetType"`

	// The contents of ipset.
	// Only available when IPSetType is IPSetTypeHashIP
	Cidrs []string `json:"cidrs"`

	// The contents of ipset.
	// Only available when IPSetType is IPSetTypeHashIPPort
	CidrAndPorts []CidrAndPort `json:"cidrAndPorts"`

	// The contents of ipset.
	// Only available when IPSetType is IPSetTypeListNet
	SetNames []string `json:"setNames"`

	// The name and namespace of the source network chaos
	RawRuleSource `json:",inline"`
}
```

We would append three ordered `RawIPset` not one into `PodNetworkChaos`.

before:

```json
[{
  "setName": "chaos1_set_tgt",
  "netPortName": "chaos1_netPort_tgt",
  "netName": "chaos1_net_tgt",
  "cidrs": [
    {
      "cidr": "1.1.1.1",
      "port": 53
    },
    {
      "cidr": "1.2.3.4",
      "port": 0
    }
  ]
}]
```

after: 

```json
[
  {
    "name": "chaos1_ipport_tgt",
    "type": "hash:ip,port",
    "cidrAndPorts": [
      {
        "cidr": "1.1.1.1",
        "port": 53
      }
    ]
  },
  {
    "name": "chaos1_ip_tgt",
    "type": "hash:ip",
    "cidrs": [
      "1.2.3.4"
    ]
  },
  {
    "name": "chaos1_set_tgt",
    "type": "list:set",
    "setName": [
      "chaos1_ipport_tgt",
      "chaos1_ip_tgt"
    ]
  }
]
```


What do you think about it? @miedzinski

miedzinski:
That's a good solution too. I'll send a patch in the next few days.

miedzinski:
There is still one field removed from protobuf, but I believe this is not an issue, isn't it? If yes, I can remove a `oneof` and apply a similar layout to PodNetworkChaos.

STRRL:
> There is still one field removed from protobuf, but I believe this is not an issue, isn't it? If yes, I can remove a `oneof` and apply a similar layout to PodNetworkChaos.

It's not an issue. But I still prefer to keep the same layout to `PodNetworkChaos`. Please update it. ❤️

controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
```diff
@@ -299,15 +299,20 @@ func (impl *Impl) ApplyTc(ctx context.Context, m *podnetworkchaosmanager.PodNetw
 		}
 		targetPods = append(targetPods, pod)
 	}
-	dstIpset := ipset.BuildIPSet(targetPods, externalCidrs, networkchaos, string(tcType[0:2])+ipSetPostFix, m.Source)
-	impl.Log.Info("apply traffic control with filter", "sources", m.Source, "ipset", dstIpset)
+	dstSetIPSet, dstOtherIPSets := ipset.BuildIPSets(targetPods, externalCidrs, networkchaos, string(tcType[0:2])+ipSetPostFix, m.Source)
+	impl.Log.Info("apply traffic control with filter", "sources", m.Source, "ipset", dstSetIPSet)
```
cwen0, line 303: We should print the whole `ipset` info at this log.

controllers/podnetworkchaos/ipset/ipset.go
```diff
@@ -32,24 +32,58 @@ import (
...
-	cidrs := externalCidrs
+// BuildIPSets builds IP sets with provided pod ip list.
+// It returns the set IP set and other IP sets.
+func BuildIPSets(pods []v1.Pod, externalCidrs []v1alpha1.CidrAndPort, networkchaos *v1alpha1.NetworkChaos, namePostFix string, source string) (v1alpha1.RawIPSet, []v1alpha1.RawIPSet) {
+	setName := GenerateIPSetName(networkchaos, "set_"+namePostFix)
+	netName := GenerateIPSetName(networkchaos, "net_"+namePostFix)
+	netPortName := GenerateIPSetName(networkchaos, "netport_"+namePostFix)
+
+	cidrs := []string{}
+	cidrandPorts := []v1alpha1.CidrAndPort{}
```
cwen0, line 43: s/cidrandPorts/cidrAndPorts/g

controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
```diff
@@ -299,15 +299,20 @@ func (impl *Impl) ApplyTc(ctx context.Context, m *podnetworkchaosmanager.PodNetw
 		}
 		targetPods = append(targetPods, pod)
 	}
-	dstIpset := ipset.BuildIPSet(targetPods, externalCidrs, networkchaos, string(tcType[0:2])+ipSetPostFix, m.Source)
-	impl.Log.Info("apply traffic control with filter", "sources", m.Source, "ipset", dstIpset)
+	dstSetIPSet, dstOtherIPSets := ipset.BuildIPSets(targetPods, externalCidrs, networkchaos, string(tcType[0:2])+ipSetPostFix, m.Source)
```
cwen0, line 302: Why should different IPSet types be handled separately here, and can it be handled together?

controllers/chaosimpl/networkchaos/partition/impl.go
```diff
@@ -360,12 +360,18 @@ func (impl *Impl) SetDrop(ctx context.Context, m *podnetworkchaosmanager.PodNetw
...
+		m.T.Append(ipSet)
+	}
+
+	m.T.Append(dstSetIPSet)
+
 	m.T.Append(v1alpha1.RawIptables{
 		Name:      iptable.GenerateName(pbChainDirection, networkchaos),
 		Direction: chainDirection,
-		IPSets:    []string{dstIpset.Name},
+		IPSets:    []string{dstSetIPSet.Name},
```
cwen0, line 374: We should append `dstOtherIPSets` to here.  @YangKeao PTAL

pkg/chaosdaemon/ipset_server.go
```diff
@@ -32,6 +32,10 @@ const (
 	ipsetExistErr        = "set with the same name already exists"
 	ipExistErr           = "it's already added"
 	ipsetNewNameExistErr = "a set with the new name already exists"
+
+	SetIPSet     = "list:set"
```
cwen0, line 36: Why not use `v1alpha1.SetIPSet` directly?

controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
miedzinski: The `list:set` set is the only one that needs to be referenced by iptables, as it stores all other IP sets, but we still need to return all IP sets, because chaos-daemon must create all of them.

If the `BuildIPSets` returned a single array, the caller would need to search for `list:set` type or assume that it's the first element of array. I think using separate return value is more elegant.

controllers/chaosimpl/networkchaos/partition/impl.go
miedzinski: There is no need to - that's the whole point of `list:set` IP set. Matching against set type matches all other IP sets, so there is no need to duplicate iptables rules.

controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
cwen0: Yes, I agree with using separate return value is more elegant. So can we use two functions to generate implement `BuildIPSet` logic?

controllers/chaosimpl/networkchaos/partition/impl.go
cwen0: OK, Got it, This is my fault! Can you add some comments to explain the different IPSet types?  I just found this in the test case.

cwen0:
```
					{
						Name:     "ipset-set-name",
						Type:     "list:set",
						SetNames: []string{"set-1", "set-2"},
					},
					{
						Name:  "ipset-net-name",
						Type:  "hash:net",
						Cidrs: []string{"0.0.0.0/24"},
					},
					{
						Name: "ipset-net-port-name",
						Type: "hash:net,port",
						CidrAndPorts: []*pb.CidrAndPort{{
							Cidr: "1.1.1.1/32",
							Port: 80,
						}},
					},
``` 
From this case, the `list:set` is not an ipset type. It's just a set of ipset's name.

cwen0: I search the ipset doc, `list:set` also can be used as an ipset type. Ignore this `It is not appropriate to define it as an ipset type. ` advice.

controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
miedzinski: I'm not sure I follow. Can you explain in more detail what do you want to change? Right now `BuildIPSets` builds all IP sets definitions, including a `list:set` set that stores all other sets. Would you rather want the callers to assemble all sets type by type? That'd probably result in more code duplication, as it's done in at least two places (tc and partition).

cwen0: We can use `BuildIPSets` to generate `NetIPSet` and `NetPortIPSet`, and then use these to generate `list:set`.

cwen0: I just think `dstOtherIPSets` is confusing and ugly.

controllers/podnetworkchaos/netutils/cidr.go
```diff
@@ -43,29 +48,50 @@ func ResolveCidrs(names []string) ([]string, error) {
...
-	_, ipnet, err := net.ParseCIDR(name)
+func ResolveCidr(name string) ([]v1alpha1.CidrAndPort, error) {
+	var toResolve string
+	var port uint16
+
+	if host, portStr, err := net.SplitHostPort(name); err == nil {
+		toResolve = host
+		port64, err := strconv.ParseUint(portStr, 10, 16)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse port %w", err)
```
cwen0, line 59: `errors.Errorf("parse port %w", err) ` , You can refer to https://github.com/chaos-mesh/rfcs/blob/main/text/2021-09-27-refine-error-handling.md.

miedzinski: Done!

api/v1alpha1/podnetworkchaos_types.go
```diff
@@ -55,18 +55,50 @@ type PodNetworkChaosSpec struct {
...
 
+// CidrAndPort represents CIDR and port pair
+type CidrAndPort struct {
+	Cidr string `json:"cidr"`
+
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=65535
+	Port uint16 `json:"port"`
+}
+
```
STRRL, line 101: We should also append fields to introduce protocol, like `tcp`, `udp` or `icmp`.  When specify ipset rule without protocol, it would only specify with `tcp`.

controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
STRRL: > Yes, I agree with using separate return value is more elegant. So, can we use two functions to generate implement BuildIPSet logic?

> I just think dstOtherIPSets is confusing and ugly.

Do we really need separate it into two functions? Or only rename `dstOtherIPSets` to `dstIPSets` is good enough. 🥲

api/v1alpha1/podnetworkchaos_types.go
STRRL: example usage  like `192.168.1.1,udp:53`

controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
cwen0: `dstIPSets` is ok, and separating into two functions is also necessary.

pkg/chaosdaemon/iptables_server.go
```diff
@@ -94,10 +94,10 @@ func (iptables *iptablesClient) setIptablesChain(chain *pb.Chain) error {
 	var matchPart string
 	var interfaceMatcher string
 	if chain.Direction == pb.Chain_INPUT {
-		matchPart = "src"
+		matchPart = "src,dst"
```
YangKeao, line 97: Why do we need this modification? If I'm not wrong, it means this chain applies, iff both the source and destination are in the ipset, which seems not to be what we want?

miedzinski: This means it must match a pair of source IP address and destination port. Please refer to [man iptables-extensions - section set, `--match-set` parameter](https://ipset.netfilter.org/iptables-extensions.man.html) and [man ipset - section introduction, set dimensions](https://ipset.netfilter.org/iptables-extensions.man.html).

pkg/chaosdaemon/iptables_server.go
```diff
@@ -94,10 +94,10 @@ func (iptables *iptablesClient) setIptablesChain(chain *pb.Chain) error {
...
 		interfaceMatcher = "-i"
 	} else if chain.Direction == pb.Chain_OUTPUT {
-		matchPart = "dst"
+		matchPart = "dst,dst"
```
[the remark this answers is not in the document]
miedzinski: `dst,dst` means it must match destination IP address and destination port. Previously it matched destination IP address only.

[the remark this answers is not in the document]
miedzinski: This is unrelated to netem, as packet filtering is done using iptables and ipsets. Simply put, this change means iptables must match destination ports in addition to source/destination IP addresses. Please take a look at documentation linked in https://github.com/chaos-mesh/chaos-mesh/pull/2932#discussion_r828962532.

controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
miedzinski: I don't know what to do then. Please tell me whether I should revert the change and change the variable name or leave it as is.

api/v1alpha1/podnetworkchaos_types.go
miedzinski: This is something I'd rather avoid right now for a couple of reasons:

- This PR is already huge. Filtering by protocols needs more changes to IP sets and/or iptables rules.
- Adding a support for filtering by protocols is outside the scope of this PR.
- Defaulting to TCP would break backward compatibility. As of now Chaos Mesh is protocol agnostic.

controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
cwen0: The current implementation is ok, just keep it.

api/v1alpha1/podnetworkchaos_types.go
STRRL: Got that!

STRRL:
Hi @miedzinski , could you append an entry in section `Unreleased / Added` of CHANGELOG.MD?

miedzinski:
> Hi @miedzinski , could you please append an entry in section `Unreleased / Added` of CHANGELOG.MD?

@STRRL Done!

pkg/chaosdaemon/iptables_server.go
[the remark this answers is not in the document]
STRRL: When applying tc rules, we need to dispatch some iptables rules, see:

https://github.com/chaos-mesh/chaos-mesh/blob/d0b0236962d30aa897e651f4d9741d383e3ce5d7/pkg/chaosdaemon/tc_server.go#L129-L151

 https://github.com/chaos-mesh/chaos-mesh/blob/d0b0236962d30aa897e651f4d9741d383e3ce5d7/pkg/chaosdaemon/tc_server.go#L274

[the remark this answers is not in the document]
STRRL: > This is unrelated to netem, as packet filtering is done using iptables and ipsets. Simply put, this change means iptables must match destination ports in addition to source/destination IP addresses. Please take a look at documentation linked in [#2932 (comment)](https://github.com/chaos-mesh/chaos-mesh/pull/2932#discussion_r828962532).

Thanks! I would take a look!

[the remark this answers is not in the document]
STRRL: Always using `dst` make sense, because we could only determine the destination/server port.

pkg/chaosdaemon/iptables_server.go
STRRL: I think we need more documentation to describe it but it still makes sense:

- with `action: delay/loss/...` (which requires `tc`) and `direction: from`, `externalTargets` would not works
- with `action: partition`, `direction: from`, `externalTagets: [10.96.0.0/16:8080]`, if would drop any packet from `10.96.0.0/16` to `<pod-ip>:8080`.

STRRL (commented):
LGTM!

Thanks for your contribution! 🤩

cwen0 (approved):
LGTM

PR #2942, opened 2022-02-25
Make time chaos completely recoverable

Andrewmatilde:
Time chaos cannot completely recover now.
Time chaos cannot combine chaos effect now.

Main: 

1.  Add totally recover code to recover the chaos.
2.  Use tasks manager to combine time chaos effect , to ensure : 
             After time_skew or recover_time_skew the status of target process will conform to the expected.
 Example: time_skew(1h) + time_skew(1h) + recover_(time_skew(1h)) -> time_skew(1h)

Other:
Add a nice common error package in `pkg/cerr/..`

### Checklist

Tests

- [x] Unit test
- [x] E2E test


Andrewmatilde:
PTAL !
/cc @YangKeao @STRRL  @iguoyr @Hexilee @WangXiangUSTC

Andrewmatilde:
PTAL @YangKeao @STRRL  @iguoyr @Hexilee @WangXiangUSTC

controllers/chaosimpl/timechaos/impl.go
```diff
@@ -69,6 +69,8 @@ func (impl *Impl) Apply(ctx context.Context, index int, records []*v1alpha1.Reco
 		Sec:         sec,
 		Nsec:        nsec,
 		ClkIdsMask:  mask,
+		Uid:         string(obj.GetUID()) + string(decodedContainer.Pod.GetUID()),
```
cwen0, line 72: g/Uid/UID/g

controllers/chaosimpl/timechaos/impl.go
```diff
@@ -69,6 +69,8 @@ func (impl *Impl) Apply(ctx context.Context, index int, records []*v1alpha1.Reco
...
+		PodId:       string(decodedContainer.Pod.GetUID()),
```
cwen0, line 73: s/PodId/PodID/g

pkg/chaosdaemon/tasks/pod_handler.go
```diff
@@ -0,0 +1,118 @@
...
+		rwLock: sync.RWMutex{},
+	}
+}
+
+func (p *PodProcessMap) Read(podPID PodID) (SysPID, error) {
+	p.rwLock.RLock()
+	defer p.rwLock.RUnlock()
+	sysPID, ok := p.m[podPID]
+	if !ok {
+		return SysPID(0), ErrNotFoundSysPID
```
cwen0, line 58: `return SysPID(0), errors.WithStack(ErrNotFoundSysPID)`

pkg/chaosdaemon/tasks/pod_handler.go
```diff
@@ -0,0 +1,118 @@
...
+	return string(p)
+}
+
+type ChaosOnPOD interface {
+	Injectable
+	Recoverable
+}
+
+type PodProcessMap struct {
+	m      map[PodID]SysPID
```
cwen0, line 42: Can we use `sync.Map` directly?

pkg/chaosdaemon/tasks/pod_handler.go
```diff
@@ -0,0 +1,118 @@
...
+var ErrNotPodPID = errors.New("pid is not PodPID")
+var ErrPodProcessMapNotInit = errors.New("PodProcessMap not init")
+
+type PodID string
+
+func (p PodID) ToID() string {
+	return string(p)
+}
+
+type ChaosOnPOD interface {
```
cwen0, line 36: Add more comments for this file.

pkg/chaosdaemon/tasks/pod_handler.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+func (p *PodHandler) Inject(pid PID) error {
+	podPID, ok := pid.(PodID)
+	if !ok {
+		return ErrNotPodPID
+	}
+	if p.PodProcessMap == nil {
+		return ErrPodProcessMapNotInit
```
cwen0, line 90: Use `errors.WithStack`

pkg/chaosdaemon/tasks/pod_handler.go
cwen0: Refer this https://github.com/chaos-mesh/rfcs/blob/main/text/2021-09-27-refine-error-handling.md

Andrewmatilde: ErrNotFoundSysPID already have a stack. It is created by errors.New()

pkg/chaosdaemon/tasks/pod_handler.go
Andrewmatilde: ErrPodProcessMapNotInit already have a stack. It is created by errors.New()

pkg/chaosdaemon/tasks/pod_handler.go
Andrewmatilde: Added.

cwen0: s/ChaosOnPod/g

pkg/chaosdaemon/time_server_test.go
```diff
@@ -15,93 +15,79 @@
...
-
-			_, err := s.RecoverTimeOffset(context.TODO(), &pb.TimeRequest{
-				ContainerId: "containerd://container-id",
-			})
-			Expect(err).ToNot(BeNil())
-			Expect(err.Error()).To(Equal(errorStr))
-		})
-	})
-})
+//var _ = Describe("time server", func() {
```
cwen0, line 18: We need new tests instead of ignoring them directly!

pkg/chaosdaemon/tasks/task_config_manager.go
```diff
@@ -137,10 +150,10 @@ func (m TaskConfigManager) SumTaskConfig(uid UID) (TaskConfig, error) {
...
-		main: taskRaw.main,
-		data: taskRaw.data.DeepCopy(),
+		Main: taskRaw.Main,
+		Data: taskRaw.Data.DeepCopy(),
 	}
-	uids := m.GetUIDsWithPID(task.main)
+	uids := m.GetUIDsWithPID(task.Main)
 
 	for _, uidTemp := range uids {
 		if uid == uidTemp {
```
cwen0, line 159: Why ignore the same `uid`? In my opinion, this is exactly what we need.

pkg/chaosdaemon/tasks/task_config_manager.go
```diff
@@ -105,25 +107,36 @@ func (m TaskConfigManager) DeleteTaskConfig(id UID) error {
...
+	if t.Main != pid {
+		return ErrDiffPID
+	}
+	return nil
+}
+
+// SumTaskConfig will sum the TaskConfig with a same TaskConfig.Main.
 // If developers want to use it with type T, they must implement Addable for *T.
 // IMPORTANT: Just here , we do not assume A.Add(B) == B.Add(A).
 // What SumTaskConfig do : A := new(TaskConfig), A.Add(B).Add(C).Add(D)... , A marked as uid.
```
cwen0, line 142: I thins `MergeTaskConfig` is a better name.

pkg/chaosdaemon/tasks/task_manager.go
```diff
@@ -37,13 +38,13 @@ type Recoverable interface {
 }
 
 // Creator init an Injectable with values.
-// We use it in a case that TaskConfig.data init an Injectable task here.
+// We use it in a case that TaskConfig.Data init an Injectable task here.
 type Creator interface {
```
cwen0, line 42: I think we can define an interface instead of `Creator`, `Addable`, `Object`, `Assign`, and `TaskExecutor`. 
```
type TaskExecutor interface {
	Merge() error 
        Assign() error 
        ...
}
```

pkg/chaosdaemon/tasks/task_config_manager.go
```diff
@@ -105,25 +107,36 @@ func (m TaskConfigManager) DeleteTaskConfig(id UID) error {
...
 	t, ok := m.TaskConfigMap[id]
 	if !ok {
 		return TaskConfig{}, ErrUIDNotFound
 	}
-	return t.data, nil
+	return t, nil
 }
 
 func (m TaskConfigManager) GetUIDsWithPID(id PID) []UID {
 	uIds := make([]UID, 0)
```
cwen0, line 119: s/uIds/uids/g

pkg/chaosdaemon/tasks/pod_handler.go
Andrewmatilde: Fixed here , I find most of error in this package can be write with a better generic format after golang version uodate to 1.18, so I will finally update in this PR after the bump PR merged.
An example:
```golang 1.17
// Now
func NotFound(type_name string) error {
	return errors.Errorf("%s not found.", type_name)
}
```

```golang 1.18
// golang 1.18
func NotFound[T any]() error {
	var inner T
	return errors.New(fmt.Sprintf("%T not found", inner))
}
```

pkg/chaosdaemon/time_server_test.go
Andrewmatilde: Actually , I want to delete all the test here ,because these test is weird. The test in e2e-test & unit test in pkg/time will cover these test.

pkg/chaosdaemon/tasks/task_config_manager.go
Andrewmatilde: I agree with you.

Andrewmatilde: Fixed.

pkg/chaosdaemon/tasks/task_manager.go
Andrewmatilde: I split the interface because developer choose to implement some of these interface. If developer do not need some part of function , they do not need to inplement the base interface (sometimes they may not have the ability to implement it).
I think it is a kind of falling back ,if we merge the interface.

pkg/chaosdaemon/time_server.go
```diff
@@ -18,65 +18,88 @@ package chaosdaemon
...
+	podProcessMap tasks.PodProcessMap
+	manager       tasks.TaskManager
+	logger        logr.Logger
+}
+
+func (s *TimeChaosServer) SetPodProcess(podID tasks.PodID, sysID tasks.SysPID) {
+	s.podProcessMap.Write(podID, sysID)
+}
+
+func (s *TimeChaosServer) SetTimeOffset(uid tasks.UID, pid tasks.PID, config time.Config) error {
```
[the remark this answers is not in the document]
Andrewmatilde: I agree with you.
I changed it :
PID -> interface IsID {ToID}
UID -> taskID

pkg/chaosdaemon/tasks/pod_handler.go
Andrewmatilde: All Fixed.

pkg/chaosdaemon/tasks/pod_handler.go
Andrewmatilde: All fixed.

pkg/chaosdaemon/tasks/pod_handler.go
```diff
@@ -0,0 +1,123 @@
...
+func (p *PodProcessMap) Write(podPID PodID, sysPID SysPID) {
+	p.rwLock.Lock()
+	defer p.rwLock.Unlock()
+	p.m[podPID] = sysPID
+}
+
+// PodHandler implements injecting & recovering on a kubernetes POD.
+type PodHandler struct {
+	PodProcessMap *PodProcessMap
+	Main          ChaosOnPOD
```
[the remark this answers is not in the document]
Andrewmatilde: I agree with you.
I changed it :
PodHandler.Main -> PodHandler.SubProcess
ProcessGroupHandler.Main -> ProcessGroupHandler.LeaderProcess

Andrewmatilde:
PTAL again, @cwen0 .

pkg/chaosdaemon/tasks/pod_handler.go
```diff
@@ -0,0 +1,122 @@
...
+	if !ok {
+		return SysPID(0), ErrNotFoundSysID.WithStack().Err()
+	}
+	return sysPID, nil
+}
+
+func (p *PodProcessMap) Write(podPID PodID, sysPID SysPID) {
+	p.rwLock.Lock()
+	defer p.rwLock.Unlock()
+	p.m[podPID] = sysPID
```
YangKeao, line 66: Will this `podPID` be recycled? or is it a memory leak?

pkg/chaosdaemon/tasks/task_config_manager.go
```diff
@@ -18,147 +18,163 @@ package tasks
...
+		return ErrDiffID.Wrapf("expect: %v, input: %v", t.Main, pid).Err()
+	}
+	return nil
+}
+
+// MergeTaskConfig will sum the TaskConfig with a same TaskConfig.Main.
+// If developers want to use it with type T, they must implement Mergeable for *T.
+// IMPORTANT: Just here , we do not assume A.Merge(B) == B.Merge(A).
+// What MergeTaskConfig do : A := new(TaskConfig), A.Merge(B).Merge(C).Merge(D)... , A marked as uid.
+func (m TaskConfigManager) MergeTaskConfig(uid TaskID) (TaskConfig, error) {
```
YangKeao, line 146: `MergeTaskConfig` is based on the `pid` right? It collects every config on a same `pid` and merge them together, but why it receives a `uid` and get the `pid` from it? Can we pass the `pid` directly to this function?

controllers/chaosimpl/timechaos/impl.go
YangKeao, line 73: What will happen if I start two `TimeChaos` to inject different containers in the same pod? Is it expected for them to have same `PodId`?

pkg/chaosdaemon/tasks/pod_handler.go
Andrewmatilde: I agree with it , it is a mem leak .

Andrewmatilde: I fixed it.

pkg/chaosdaemon/tasks/task_config_manager.go
Andrewmatilde: It is because we can get a stable First Task base on this, meaning given two tasks uid A.uid & B.uid , the Merge order of MergeTaskConfig(A.uid) A.Merge(B) is differnt from the order of MergeTaskConfig(B.uid) : B.Merge(A).
If just give a pid we cannot divide these situation.

YangKeao: So the `Merge` operation is not required to be commutative?

Andrewmatilde: Yes.

controllers/chaosimpl/timechaos/impl.go
Andrewmatilde: Sorry , I will fix it. When I start the work , it is not a problem because Timechaos have no container selector.

pkg/chaosdaemon/tasks/task_config_manager.go
YangKeao: No! As the golang map is not ordered, you cannot assume the order of the following merge (even though the first one is stable). For example, you may have `B.Merge(A).Merge(C)` or `B.Merge(C).Merge(A)`. The `Merge` has to be commutative to make them the same.

If we have the commutativity, can we use the `pid` instead of `uid` here?

pkg/chaosdaemon/tasks/pod_container_handler.go
```diff
@@ -0,0 +1,129 @@
...
+	p.rwLock.RLock()
+	defer p.rwLock.RUnlock()
+	sysPID, ok := p.m[PodContainerName]
+	if !ok {
+		return SysPID(0), ErrNotFoundSysID.WithStack().Err()
+	}
+	return sysPID, nil
+}
+
+func (p *PodContainerNameProcessMap) Write(PodContainerName PodContainerName, sysPID SysPID) {
```
YangKeao, line 63: I don't understand why this struct is necessary? This maps the "podcontainername" to a pid, but the pid can also be got from `containerId`. Whenever I got a podcontainername, the containerId is usually also touchable, doesn't it?

pkg/chaosdaemon/tasks/process_group_handler.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	childPids, err := util.GetChildProcesses(uint32(sysPID), gp.Logger)
+	if err != nil {
+		return cerr.NotFound("child process").WrapErr(err).Err()
+	}
+
+	for _, childPID := range childPids {
+		childSysPID := SysPID(childPID)
+		if childProcessChaos, ok := gp.childMap[childSysPID]; ok {
```
YangKeao, line 127: Maybe we could iterate through the `childMap` directly (without finding child processes first), it should be able to handle exited processes.

pkg/chaosdaemon/tasks/pod_container_handler.go
Andrewmatilde: But the `containerId` is not a stable key, if the container restart `containerId` will change. In chaos-daemon we need a stable ID to mark the process tasks applid on.
If use  `containerId` : container restart  and `containerId` changed from "0" to "1", we need apply all the tasks with `containerId` again to refresh the tasks.

pkg/chaosdaemon/tasks/task_manager.go
```diff
@@ -100,55 +97,49 @@ func (cm TaskManager) CopyTaskConfigManager() TaskConfigManager {
...
 
 // Create the first task,
 // the New function of TaskExecutor:Creator will only be used here.
 // values is only the import parameter of New function in TaskExecutor:Creator.
-// If it comes a task are already be injected on the PID,
+// If it comes a task are already be injected on the IsID,
 // Create will return ChaosErr.ErrDuplicateEntity.
-func (cm TaskManager) Create(uid UID, pid PID, config TaskExecutor, values interface{}) error {
+func (cm TaskManager) Create(uid TaskID, pid IsID, config TaskExecutor, values interface{}) error {
 	if _, ok := cm.taskMap[pid]; ok {
```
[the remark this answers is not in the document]
Andrewmatilde: I agree with you. I will offer some locker.

pkg/chaosdaemon/tasks/process_group_handler.go
Andrewmatilde: I agree with you.

pkg/chaosdaemon/tasks/task_manager.go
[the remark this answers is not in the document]
Andrewmatilde: Fixed.

pkg/chaosdaemon/tasks/process_group_handler.go
Andrewmatilde: Fixed.

pkg/chaosdaemon/tasks/pod_container_handler.go
YangKeao: I just don't understand why we need to store it. For any "Apply" and "Recover", we can get the pid directly, without reading from any existing map, so why we need to read it from the map (by `podcontainername`)?

Andrewmatilde: If one of process restart with different PID, we need to know which PID is the last one cause we need to apply tasks from the last process to the new process.The `map[podcontainername]` is for merging tasks.

Andrewmatilde: On another hand, I choose store something for merging tasks apply on a same process.

YangKeao: > I choose store something for merging tasks apply on a same process

It seems that the `TaskConfigMap` is used to merge the tasks, and this map is not read during the merge :thinking: . (Please correct me if I was wrong).

Andrewmatilde: `podcontainername` is also stored in `TaskConfigMap`.

Andrewmatilde: `podcontainername` is the value of `TaskConfigMap` & key of `taskmap`. `PodContainerNameProcessMap`  store the relationship of `podcontainername` and working process PID.

pkg/chaosdaemon/tasks/task_config_manager.go
Andrewmatilde:
```
No! As the golang map is not ordered, you cannot assume the order of the following merge (even though the first one is stable). For example, you may have B.Merge(A).Merge(C) or B.Merge(C).Merge(A). The Merge has to be commutative to make them the same.
```

But I can make sure the first one , that was enough.

Andrewmatilde: Using `pid` means rejecting developer with incommutative config , I think it is not a good idea.

Andrewmatilde: Just for time-chaos in `clockIDsMask`. Developers can choose different policy on merging `clockIDsMask` base on this, but not just `clockIDsMask |= clockIDsMask` for every config.

pkg/chaosdaemon/tasks/pod_container_handler.go
YangKeao: > If one of process restart with different PID, we need to know which PID is the last one cause we need to apply tasks from the last process to the new process.The `map[podcontainername]` is for merging tasks.

Everytime we call `Apply` or `Recover`, we actually know the pid.

Andrewmatilde: Yes . And I do not know the problem now.

YangKeao: After discussing with @Andrewmatilde, I understand the problem.

1. The map is used to pass the `pid` to the `PodHandler`, so that the `PodHandler` knows which process to inject.
2. Why don't we pass the `pid` as an argument to the `PodHandler`? Because we cannot replace the argument of `Inject` to the `pid` (or `container_id`), they are not stable.

Every K-V in this map works like a global variable, to tell the `PodHandler` pid. I'm thinking whether there are better way (like using closure or something else) to pass the information to `PodHandler`. (At least, if the abstraction doesn't fit in our logic, we should refractor the abstraction, but not use a "global variable" to bypass it).

Andrewmatilde: But actually the map is not a `global variable`  and we can control the shared scope. I think we can add pid to time-chaos-config and use config.Assign & config.New to transfer the pid, but I think it is a dirtier choise.

YangKeao (approved):
As I cannot come up with a better idea. LGTM

cwen0 (approved):
LGTM

PR #2951, opened 2022-02-28
fix: byteman attach error

### LINKED ISSUE #2588: JVMChaos: target process doesn't respond within 10500ms or hotspot vm not loaded
The error is shown below：
image

zhqqqy:
The jvm-chaos can also be injected when the chaos-mesh is running on a linux version lower than linux 4.1, just a  draft

Close #2588 

Get whether the current linux version running chaos-daemon is less than 4.1, if it is less than 4.1, execute the backup method.
1. Unzip byteman.tar.gz, set the environment variable of shoe byteman_home in the bminstall.sh and bmsubmit.sh scripts to `/usr/local/byteman`
2. Repackage these files into `byteman.tar.gz`
3. Copy byteman.tar.gz to `/usr/local` inside the container to be injected into the experiment and extract it
4. Execute `bminstall.sh` and `bmsubmit.sh `through the client of docker or containerd to perform fault injection on the java process

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


pkg/chaosdaemon/jvm_server_backup.go
```diff
@@ -0,0 +1,171 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if len(output) > 0 {
+		log.Info("tar byteman.tar.gz", "output", string(output))
+	}
+
+	bmInstallCmd := fmt.Sprintf(bmInstallCommandBackUp, req.Port, javaPid)
+
+	output, err = s.crClient.ExecCommandByContainerID(ctx, req.ContainerId, []string{"sh", "-c", bmInstallCmd})
```
YangKeao, line 85: Is it necessary to execute the script through container runtime? The `bpm` package supports setting the namespaces (e.g. `.SetNS(pid, bpm.MountNS)` above).

zhqqqy: I tried it, but got a problem with environment variables like `JAVA_HOME`。I remember that a similar issue was raised under the [nsexec](https://github.com/chaos-mesh/nsexec) project before. https://github.com/chaos-mesh/nsexec/issues/1

pkg/chaosdaemon/jvm_server_backup.go
```diff
@@ -0,0 +1,171 @@
...
+
+	if err != nil {
+		// this error will occured when install agent more than once, and will ignore this error and continue to submit rule
+		errMsg1 := "Agent JAR loaded but agent failed to initialize"
+
+		// these two errors will occured when java version less or euqal to 1.8, and don't know why
```
Gallardot, line 91:
```suggestion
		// these two errors will occured when java version less or equal to 1.8, and don't know why
```

Gallardot (changes requested):
Great jobs! Thanks for your contributing! I will test this feature in my test environment later.

PS: you can also fix CI issues first 😁.

pkg/chaosdaemon/jvm_server_backup.go
YangKeao: I prefer to get the environment variables from the procfs (`/proc/PID/environ`) or container runtime, set the environment variables of the subprocess, then use the `nsexec`.

zhqqqy: I guess you mean is this 
``` go
	//todo
	processBuilder = bpm.DefaultProcessBuilder("sh", "-c", bmInstallCmd).SetContext(ctx)
	if err = setEnvsToProcess(processBuilder, strconv.Itoa(int(pid))); err != nil {
		return nil, err
	}
	processBuilder = processBuilder.SetNS(pid, bpm.MountNS).SetStdin(agentFile)
	output, err = processBuilder.Build().CombinedOutput()
```
set Envs to Process
```go
func setEnvsToProcess(processBuilder *bpm.ProcessBuilder, pid string) error {
	//todo ,get env by pid from /proc/pid/environ
	bEnvs, err := util.GetEnvsByProcess(strconv.Itoa(int(pid)))
	if err != nil {
		return err
	}
      //split  envs content by []byte{0}
	for _, env := range util.SplitEnvs(bEnvs) {
		envMap := strings.Split(env, "=")
		processBuilder.SetEnv(envMap[0], envMap[1])
	}
	return nil
}
```
right?

zhqqqy: prompt after I try
> 	{"error": "fork/exec /usr/local/bin/nsexec: invalid argument"}

YangKeao: > I guess you mean is this

Yes 🍻 

> {"error": "fork/exec /usr/local/bin/nsexec: invalid argument"}

Have you uploaded the code? Maybe I could debug for you 🤔 . I guess there is something unexpected inside the environment variables.

zhqqqy: code has been commited,thanks for your help。🌹

YangKeao: @zhqqqy I have fixed (and added some tests), and the code has been pushed, please check whether it works 😄 .

zhqqqy: Unfortunately, rule submit failed, prompt `Connection refused`
image
the byteman has been attahed to target proocess
image

pkg/chaosdaemon/jvm_server_backup.go
```diff
@@ -0,0 +1,217 @@
...
+func (s *DaemonServer) InstallJVMRulesBackUp(ctx context.Context,
+	req *pb.InstallJVMRulesRequest) (*empty.Empty, error) {
+	pid, err := s.crClient.GetPidFromContainerID(ctx, req.ContainerId)
+	if err != nil {
+		log.Error(err, "GetPidFromContainerID")
+		return nil, err
+	}
+
+	//todo get pid in the container，
+	processBuilder := bpm.DefaultProcessBuilder("sh", "-c", "ps -ef | grep java | grep -v grep | awk 'NR==1 {print $2}'").SetContext(ctx).SetNS(pid, bpm.MountNS).SetNS(pid, bpm.PidNS)
```
YangKeao, line 46: The result of this script is not trustable. For example, the return value of this script in `examples/jvm/app.yaml` is `root`:

```
chaos-daemon-server     get java pid    {"output": "root\n"}
```

I wonder why we cannot suggest the first process is the java process?

pkg/chaosdaemon/jvm_server_backup.go
```diff
@@ -0,0 +1,217 @@
...
+	log.Info("get java pid", "output", string(output))
+
+	javaPid, err := strconv.Atoi(strings.Replace(string(output), "\n", "", -1))
+	if err != nil {
+		return nil, err
+	}
+
+	// todo: Need to write the BYTEMAN_HOME environment variable in bminstall.sh and bmsubmit.sh
+	// or do this in code ?
+	agentFile, err := os.Open("/usr/local/bin/byteman.tar.gz")
```
YangKeao, line 60: This file is removed in the `Dockerfile` 🤔

pkg/chaosdaemon/jvm_server_backup.go
```diff
@@ -0,0 +1,217 @@
...
+	"strings"
+
+	"github.com/chaos-mesh/chaos-mesh/pkg/bpm"
+	pb "github.com/chaos-mesh/chaos-mesh/pkg/chaosdaemon/pb"
+	"github.com/golang/protobuf/ptypes/empty"
+)
+
+const (
+	bytemanHome            = "/usr/local/byteman/"
+	bmInstallCommandBackUp = bytemanHome + "bin/bminstall.sh -b -Dorg.jboss.byteman.transform.all -Dorg.jboss.byteman.verbose -Dorg.jboss.byteman.compile.to.bytecode -p %d %d"
```
YangKeao, line 33: This script requires the target container to have bash, which needs to be clarified.

pkg/chaosdaemon/jvm_server_backup.go
YangKeao: Fixed. The `bmsubmit.sh` script should be run inside the target network namespace (see my latest commit).

pkg/chaosdaemon/jvm_server_backup.go
YangKeao: @zhqqqy I removed this logic in my latest commit. If you think it's not reasonable, feel free to add it back with a comment of the situation when the java process is not the first one.

zhqqqy: > The result of this script is not trustable. For example, the return value of this script in `examples/jvm/app.yaml` is `root`:
> 
> ```
> chaos-daemon-server     get java pid    {"output": "root\n"}
> ```
> 
> I wonder why we cannot assume the first process is the java process?

Because in our internal system, the process mounted by the container is an environment initialization script. When the container starts, some necessary programs, such as `databus, crond, nscd`, etc., will be initialized first, and then the java project will be started.

pkg/chaosdaemon/jvm_server_backup.go
zhqqqy: > Fixed. The `bmsubmit.sh` script should be run inside the target network namespace (see my latest commit).

great, works perfectly!

pkg/chaosdaemon/jvm_server_backup.go
zhqqqy: how about this?
```
ps -eo pid,comm | grep java | awk 'NR==1 {print $1}'
```

pkg/chaosdaemon/jvm_server_backup.go
```diff
@@ -0,0 +1,209 @@
...
+	output, err = processBuilder.Build().CombinedOutput()
+	if err != nil {
+		return nil, err
+	}
+	if len(output) > 0 {
+		log.Info("tar byteman.tar.gz", "output", string(output))
+	}
+
+	// Assuming the first process is the java process
+	bmInstallCmd := fmt.Sprintf(bmInstallCommandBackUp, req.Port, 1)
```
WangXiangUSTC, line 73: could find the first java process, refer to https://github.com/chaos-mesh/chaos-mesh/pull/2904

pkg/chaosdaemon/jvm_server_backup.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/golang/protobuf/ptypes/empty"
+)
+
+const (
+	bytemanHome            = "/usr/local/byteman/"
+	bmInstallCommandBackUp = bytemanHome + "bin/bminstall.sh -b -Dorg.jboss.byteman.transform.all -Dorg.jboss.byteman.verbose -Dorg.jboss.byteman.compile.to.bytecode -p %d %d"
+	bmSubmitCommandBackUp  = bytemanHome + "bin/bmsubmit.sh -p %d -%s %s"
+)
+
+func (s *DaemonServer) InstallJVMRulesBackUp(ctx context.Context,
```
WangXiangUSTC, line 38: How about checking whether java is in `PATH` at the beginning of this function?

WangXiangUSTC (commented):
Thanks for your contribution, It's extremely useful!

pkg/chaosdaemon/jvm_server_backup.go
zhqqqy: I have studied this code, and all he looks for is the pid of the host where the chaos-daemon is located, but here byteman needs to attach the pid inside the container, and needs to read the /proc directory inside the container

pkg/chaosdaemon/jvm_server_backup.go
zhqqqy: Then I think the `PATH ` environment variable should be analyzed first, and if it does not exist, an error will be thrown, and if it exists, the execution will continue.

iguoyr:
@zhqqqy Hi, please fix CI error first, and if you need some help feel free to leave any messages.

zhqqqy:
> @zhqqqy Hi, please fix CI error first, and if you need some help feel free to leave any messages.

can help me resolve conflicts files?

STRRL:
I guess that you use `rebase` instead of `merge` so your commits are spliced after the commits (which look like the origin commits on the master branch but committed by you).

And you have force pushed your codes, I tried to checkout the commit 7ae15fa  (which looks like the commit before rebase) but I failed., it does not exist on the remote repository.

 I think it would only exist on your local git repo, unless you run the `git gc`.

So, there are two ways to fix the current situation:

- find the commit which is the commit before the rebase, you could use the comments on github or `git reflog` on you local git repo to find it 
- manually cherry-pick all the commit to the new branch;

STRRL:
If you do not so familiar with git, I suggest the second way, it could be finished by you or by us. :)

If you need any help, feel free to tell us. :heart: @zhqqqy

zhqqqy:
> If you do not so familiar with git, I suggest the second way, it could be finished by you or by us. :)
> 
> If you need any help, feel free to tell us. ❤️ @zhqqqy

I'm sorry, because I haven't responded to some of my personal affairs for a long time, because I have left the company, and my local files no longer exist. Can you help me fix this problem directly. thank you very much

g1eny0ung:
This PR has had a long-time of no response.

Since there are too many conflicts, I suggest closing this PR and re-submitting your changes into a new one.

PR #2955, opened 2022-03-01
fix: add `envFollowKubernetesPattern` to handle k8s-like format env in helm templates

### LINKED ISSUE #2924: helm: range `env` should support `valueFrom`
## Feature Request

Yes, currently the `env`'s `value` in helm charts doesn't support `valueFrom`, sometimes it's needed if someone wants to pass a `Secret` or `ConfigMap` as the ref.

Can generate `valueFrom` to real YAML.

None.

An example: https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables.

g1eny0ung:

Close #2924

```mustache
{{/*
Handle env variables.

TODO: in the future, we would like to use the k8s-like format for defining environment variables.
So the `envFollowKubernetesPattern` will become to `env`.
And the original way of writing env will be removed.
Ref: https://github.com/chaos-mesh/chaos-mesh/pull/2955.
*/}}
{{- define "chaos-mesh.helpers.list-env-vars" -}}
{{- with .envFollowKubernetesPattern }}
{{ toYaml . }}
{{- end }}
{{- range $key, $val := .env }}
- name: {{ $key | upper }}
  value: {{ $val | quote }}
{{- end }}
{{- end }}
```

### Related changes

  - [x] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


STRRL:
We could set `valueFrom` like

```yaml
dashboard:
  env:
    SOME_SECRET:
      valueFrom:
        secretKeyRef:
          name: some-secret
          key: token
```


It does not follow the `K: V`, neither follows the kubneretes like `name: <name>` then `value: <value>`

Maybe we could create another new field  near the origin `env` field, and follows the kubernetes-like format.

g1eny0ung:
> So in the PR, we could set `valueFrom` like
> 
> ```yaml
> dashboard:
>   env:
>     SOME_SECRET:
>       valueFrom:
>         secretKeyRef:
>           name: some-secret
>           key: token
> ```
> 
> It does not follow the `K: V`, neither follows the kubneretes like `name: <name>` then `value: <value>`.  😰
> 
> Maybe we could create another new field near the origin `env` field, and follows the kubernetes-like format.

But our current `env` also doesn't conform to the k8s-like format. I have no problem with using another field, but maybe we should make `env` a k8s-like format in a future release? (means that the `key: value` format can no longer be used in `env`)

g1eny0ung:
Updated. @STRRL PTAL.

STRRL (approved):

helm/chaos-mesh/templates/_helpers.tpl
```diff
@@ -1,42 +1,61 @@
...
+
+{{/*
+Handle env variables.
+
+TODO: in the future, we would like to use the k8s-like format for defining environment variables.
+So the `envFollowKubernetesPattern` will become to `env`.
+And the original way of writing env will be removed.
+Ref: https://github.com/chaos-mesh/chaos-mesh/pull/2955.
+*/}}
+{{- define "chaos-mesh.helpers.list-env-vars" -}}
```
iguoyr, line 11: According to [helm chart best practice](https://helm.sh/docs/chart_best_practices/values/#naming-conventions), maybe we should use `listEnvVars` instead of `list-env-vars`.

helm/chaos-mesh/templates/chaos-daemon-daemonset.yaml
```diff
@@ -80,10 +81,7 @@ spec:
             - /etc/chaos-daemon/cert/tls.key
           {{- end }}
           env:
-            {{- range $envKey, $envVal := .Values.chaosDaemon.env }}
-            - name: {{ $envKey | upper }}
-              value: {{ $envVal | quote }}
-            {{- end }}
+            {{- include "chaos-mesh.helpers.list-env-vars" .Values.chaosDaemon | indent 12 }}
```
iguoyr, line 84: `indent xx` here will generate unnecessary whitespaces.

helm/chaos-mesh/templates/chaos-dashboard-deployment.yaml
```diff
@@ -61,10 +62,7 @@ spec:
           command:
             - /usr/local/bin/chaos-dashboard
           env:
-            {{- range $envKey, $envVal := .Values.dashboard.env }}
-            - name: {{ $envKey | upper }}
-              value: {{ $envVal | quote }}
-            {{- end }}
+            {{- include "chaos-mesh.helpers.list-env-vars" .Values.dashboard | indent 12 }}
```
iguoyr, line 65: ditto

helm/chaos-mesh/templates/controller-manager-deployment.yaml
```diff
@@ -61,10 +62,7 @@ spec:
         command:
           - /usr/local/bin/chaos-controller-manager
         env:
-          {{- range $envKey, $envVal := .Values.controllerManager.env }}
-          - name: {{ $envKey | upper }}
-            value: {{ $envVal | quote }}
-          {{- end }}
+          {{- include "chaos-mesh.helpers.list-env-vars" .Values.controllerManager | indent 10 }}
```
iguoyr, line 65: ditto

iguoyr (commented):
Rest LGTM

helm/chaos-mesh/templates/_helpers.tpl
g1eny0ung: Now fixed!

helm/chaos-mesh/templates/chaos-daemon-daemonset.yaml
g1eny0ung: Hmmm...weird behavior here. If the indent is removed, then the indentation of yaml will generate an error. But even if I write like this:

```yaml
          env:
{{- include "chaos-mesh.helpers.listEnvVars" .Values.dashboard | indent 12 }}
```

There will still be 12 spaces. It seems that this is a bug of helm. 🤔

iguoyr: Well, it works fine on my local environment. (helm version: 3.5.3)

Could you show me some error messages?

g1eny0ung: This helps me. I'm using helm `v3.8.1` in my local, it's updated by homebrew silently. I will try to downgrade helm and generate it again.

g1eny0ung: Same behavior in `v3.5.2`. Remove `indent 12` will cause **Error: YAML parse error on chaos-mesh/templates/chaos-dashboard-deployment.yaml: error converting YAML to JSON: yaml: line 59: did not find expected key**

Then generated will be like:

```yaml
          env:
- name: CLEAN_SYNC_PERIOD
  value: "12h"
- name: DATABASE_DATASOURCE
  value: "/data/core.sqlite"
```

iguoyr: hmmm, I wonder if you try this:

```yaml
          env:
            {{- include "chaos-mesh.helpers.listEnvVars" .Values.dashboard }}
```

or this:

```yaml
          env:
{{- include "chaos-mesh.helpers.listEnvVars" .Values.dashboard }}
```

iguoyr: Oops, I found that I test the `chaos-daemon` on my local environment which the `env` is empty! 😿 

And I find a solution (got inspiration from [here](https://github.com/bitnami/charts/blob/c353b59c08c1e3026efff64d8d6d8561f0836670/bitnami/airflow/templates/_git_helpers.tpl)):
1. use `{{- end -}}` in the last line of a template
```yaml
{{- define "chaos-mesh.helpers.listEnvVars" -}}
...
{{- end -}}
```

2. use `trim` and `nindent` to control whitespaces:
```yaml
{{- include "chaos-mesh.helpers.listEnvVars" .Values.dashboard | trim | nindent 12 }}
```

BTW, I think keeping `{{- xxx -}}`(both `-`) in templates is a good practice, if we don't want to take more care of controlling whitespaces.

g1eny0ung: Thx! It works. I will update this PR soon. 😄

g1eny0ung: > BTW, I think keeping {{- xxx -}}(both -) in templates is a good practice, if we don't want to take more care of controlling whitespaces.

Makes sense. I just saw useless `-` at the end of some control statements and removed them.

g1eny0ung: Resolved.

g1eny0ung: A note: due to daemon has an value `env: {}` in `values.yaml`. The output will be:

```yaml
env:

  - name: TZ
     value: ${timezone}
```

iguoyr: > A note: due to daemon has an value `env: {}` in `values.yaml`. The output will be:
> 
> ```yaml
> env:
> 
>   - name: TZ
>      value: ${timezone}
> ```

Maybe we could check whether the field is empty (like `{{ if .Values.dashboard.env }}`), and then decide to use the template.

g1eny0ung: Completed. ✅ There should be no further questions now.

iguoyr (approved):
LGTM

Issue #2888, opened 2022-02-11
Refine logging in `pkg/selector/pod`, refactor these functions as methods of a Struct, then move package level variable `log` as a field


STRRL:
> Please leave comments with your idea if you are interested in this issue. And comment `/assign` to operate self-assign before you work on it.

- replace with log.L()
- refactor the function to methods

STRRL:
closed by #3002

PR #3002, opened 2022-03-10
chore: refine logging in pkg/selector/pod

STRRL:

It's a part of https://github.com/chaos-mesh/chaos-mesh/issues/2888;

These changes are small and easy to review. 

- using `log.L()` instead of logger in ctroller-runtime

### Related changes

  - [x] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


fewdan (approved):
LGTM

iguoyr:
@STRRL How about adding a changelog for this PR?

STRRL:
> @STRRL How about adding a changelog for this PR? Rest LGTM :)

Updated! PTAL

iguoyr (approved):
LGTM

pkg/selector/pod/selector.go
```diff
@@ -159,7 +158,7 @@ func selectSpecifiedPods(ctx context.Context, c client.Client, spec v1alpha1.Pod
 	for ns, names := range spec.Pods {
 		if !clusterScoped {
 			if targetNamespace != ns {
-				log.Info("skip namespace because ns is out of scope within namespace scoped mode", "namespace", ns)
+				log.L().WithName("pod-selector").Info("skip namespace because ns is out of scope within namespace scoped mode", "namespace", ns)
```
Andrewmatilde, line 161: It seems L is a deprecated func. Can we create a new logger?

STRRL: That's still a temporary solution. Because we have nowhere to store the logger.

Another solution is to change the function signature of `SelectAndFilterPods`, passing a logger from the caller, but changing the function signature seems a little much for this migration..


`log.L()` is designed to be used with this situation.

PR #2992, opened 2022-03-08
chaosctl forced recovery supports httpchaos, iochaos, stresschaos, and networkchaos

Hexilee:
### What problem does this PR solve?

Support forced recovery for some types of chaos.

#### NetworkChaos
Clean rules in iptables chains CHAOS-INPUT and CHAOS-OUTPUT; clean all tc qdisc rules. 

#### StressChaos
Kill stress-ng and menStress by SIGTERM.

#### IOChaos
Kill toda by SIGTERM.

#### HTTPChaos
Kill tproxy by SIGTERM.

### Related changes

- [ ] Need to update `chaos-mesh/website`
- [ ] Need to update `Dashboard UI`
- Need to **cheery-pick to release branches**
  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [x] E2E test
- [ ] No code
- [x] Manual test (add steps below)

```bash
> chaosctl recover httpchaos <PODS>
> chaosctl recover httpchaos -n <NAMESPACE>
> chaosctl recover httpchaos -l app=nginx
```

Side effects

- [ ] Breaking backward compatibility

### Release note 

```text
-  Add recover subcommand for chaosctl .
```

### DCO

If you find the DCO check fails, please run commands like below (Depends on the actual situations. For example, if the failed commit isn't the most recent) to fix it:

```shell
git commit --amend --signoff
git push --force
```

PR #3006, opened 2022-03-10
refactor(pkg/selector/pod): extract methods into struct LegacyPodSelector

STRRL:

- make codes in pkg/selector/pod cleaner and easier to read.
- it does not change the logic of the codes

- `container.SelectImpl` now need a dependency with `pod.SelectImpl` instead of call `pod.SelectAndFilterPods` directlly 
- extract function `SelectAndFilterPods`, `SelectPods`, `CheckPodMeetSelector` and related unexported functions as methods of struct `LegacyPodSelector`

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [x] Unit test
- [x] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/pkg/selector/pod/legacy.go
+++ b/pkg/selector/pod/legacy.go
@@ -1,0 +1,252 @@
+// Copyright 2022 Chaos Mesh Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package pod
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/chaos-mesh/chaos-mesh/api/v1alpha1"
+	"github.com/chaos-mesh/chaos-mesh/pkg/mock"
+	"github.com/chaos-mesh/chaos-mesh/pkg/selector/generic"
+	genericnamespace "github.com/chaos-mesh/chaos-mesh/pkg/selector/generic/namespace"
+	"github.com/chaos-mesh/chaos-mesh/pkg/selector/generic/registry"
+)
+
+// LegacyPodSelector is the temporary place to hold the legacy logic for selecting then filtering the Pod.
+type LegacyPodSelector struct {
+	// c is the kubernetes client which we usually use.
+	c client.Client
+	// r is actually the kubernetes client without indexing and caching, only used/required with fieldSelector.
+	r client.Reader
+	// clusterScoped means selector from all namespace.
+	clusterScoped bool
+	// targetNamespace means select from this namespace, only work with clusterScoped is true.
+	targetNamespace string
+	// enableFilterNamespace means select only from namespace annotated with generic.InjectAnnotationKey.
+	enableFilterNamespace bool
+}
+
+func NewLegacyPodSelector(c client.Client, r client.Reader, clusterScoped bool, targetNamespace string, enableFilterNamespace bool) *LegacyPodSelector {
+	return &LegacyPodSelector{c: c, r: r, clusterScoped: clusterScoped, targetNamespace: targetNamespace, enableFilterNamespace: enableFilterNamespace}
+}
+
+// SelectAndFilterPods returns the list of pods that filtered by selector and SelectorMode
+// Deprecated: use pod.SelectImpl as instead
+func (lps *LegacyPodSelector) SelectAndFilterPods(ctx context.Context, spec *v1alpha1.PodSelector) ([]v1.Pod, error) {
+	if pods := mock.On("MockSelectAndFilterPods"); pods != nil {
+		return pods.(func() []v1.Pod)(), nil
+	}
+	if err := mock.On("MockSelectedAndFilterPodsError"); err != nil {
+		return nil, err.(error)
+	}
+
+	selector := spec.Selector
+	mode := spec.Mode
+	value := spec.Value
+
+	pods, err := lps.SelectPods(ctx, selector)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(pods) == 0 {
+		err = errors.New("no pod is selected")
+		return nil, err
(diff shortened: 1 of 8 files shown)


pkg/selector/pod/legacy.go
```diff
@@ -0,0 +1,252 @@
...
+
+	"github.com/chaos-mesh/chaos-mesh/api/v1alpha1"
+	"github.com/chaos-mesh/chaos-mesh/pkg/mock"
+	"github.com/chaos-mesh/chaos-mesh/pkg/selector/generic"
+	genericnamespace "github.com/chaos-mesh/chaos-mesh/pkg/selector/generic/namespace"
+	"github.com/chaos-mesh/chaos-mesh/pkg/selector/generic/registry"
+)
+
+// LegacyPodSelector is the temporary place to hold the legacy logic for selecting then filtering the Pod.
+type LegacyPodSelector struct {
```
Hexilee, line 35: Why not merge `LegacyPodSelector` and `SelectImpl` into a structure? As this `LegacyPodSelector` is also a breaking change to package `pkg/selector/pod`.

Hexilee (commented):
Rest LGTM

pkg/selector/pod/legacy.go
STRRL: Do you mean embed `LegacyPodSelector` in to `SelectImplSelectImpl`? Or only move all the methods of `LegacyPodSelector` into methods of `SelectImpl`?

The former one makes not soo many changes, we could make it. 

But for the latter one, it would make more trouble when we want to purge `LegacyPodSelector` in the future. Actually, one of my main purposes is to tidy the usage of `LegacyPodSelector`, for purging it in the future.

STRRL: I think that bringing breaking changes on `pkg/selector/pod` is acceptable, unlike kubernetes CRD API, we do not make any promise on `pkg/selector/pod` (and either most of other packages ), we actually do not intend other users use it as a library.

Hexilee: Yes, breaking changes don't matter. I prefer to embed `LegacyPodSelector` into `SelectImpl`, then we can ignore the existence of `LegacyPodSelector` (only construct `SelectImpl`).

Hexilee: > Yes, breaking changes don't matter. I prefer to embed `LegacyPodSelector` into `SelectImpl`, then we can ignore the existence of `LegacyPodSelector` (only construct `SelectImpl`).

In this solution, we should privatize the `NewLegacyPodSelector`.

STRRL: 🤔 But it would make the original dependency on `LegacyPodSelector` "extended" to dependency on `pod.SelectImpl`, which bring 2 problems:

- it would use `selector pod.SelectImpl` as `select.SelectPods` **directly**, which might break the convention to using   `Selector`(in `pkg/selector/selector.go`) and each `SelecImpl` in each package, beacse it seems `pkg/selector` is designed to using `Selector`(in `pkg/selector/selector.go`) instead of designed to using `SelectImpl` directly. Or maybe I miss something? PTAL @iguoyr @YangKeao 
- the current constructor of `pod.SelectImpl` is coupled with `fx` dependency injection library.

Hexilee: I think the `fx` dependency injection doesn't prevent us from constructing an object manually, especially outside of controller-manager (like the apiserver). For applications embedded in controller-manager, we can also make them by `fx` (we will construct ctrlserver by `fx` in #2992).

STRRL: Oh, that's right, the issue about `fx` could be resolved. What do you think about the former one?

Hexilee: > It seems `pkg/selector` is designed to using `Selector`(in `pkg/selector/selector.go`) instead of designed to using `SelectImpl` directly. Or maybe I miss something? PTAL @iguoyr @YangKeao

I don't think so, as we expose `pod.SelectImpl` as a public type, and provide it by `fx`, we allow other components to use `pod.SelectImpl` straightly (and its methods are common enough to be used). If we only allow other components to use `Selector`, we should make `pod.SelectImpl` private and only expose an interface.

STRRL: >  If we only allow other components to use Selector, we should make pod.SelectImpl private and only expose an interface.

The current codes indeed expose these `SelectImpl` structs, but I think that is not the original intent...  🙈🙈🙈

iguoyr: I think it's fine to use `SelectImpl` directly or not. But if we do not want to use `SelectImpl` directly, maybe we should not embed `pod.SelectImpl` into `container.SelectImpl` at the same time (as they are under two different packages)? 🤔 

https://github.com/chaos-mesh/chaos-mesh/blob/d77e86e6d75d585815f4bf8a95484ca5e612f0e6/pkg/selector/container/selector.go#L31-L37

Hexilee: > I think it's fine to use `SelectImpl` directly or not. But if we do not want to use `SelectImpl` directly, maybe we should not embed `pod.SelectImpl` into `container.SelectImpl` at the same time (as they are under two different packages)? 🤔
> 
> https://github.com/chaos-mesh/chaos-mesh/blob/d77e86e6d75d585815f4bf8a95484ca5e612f0e6/pkg/selector/container/selector.go#L31-L37

Why not merge package `selector/pod` into `selector/container`? As the `selector/container` strongly depend on `selector/pod`.

STRRL: > But if we do not want to use SelectImpl directly, maybe we should not embed pod.SelectImpl into container.SelectImpl at the same time (as they are under two different packages)?

Please do not be restricted by golang's syntax, golang does not provide access levels other than `exported` or `unexported`. Exporting the struct but only for internal usage still commonly exists. 

I want to figure out whether `SelectImpl` is recommended to use without the limitation of  who is the invoker.

STRRL: > Why not merge package selector/pod into selector/container? As the selector/container strongly depend on selector/pod.

I even prefer to flat the whole `pkg/selector/*` into one layer of the package. 🤣

Hexilee: > I even prefer to flat the whole `pkg/selector/*` into one layer of the package. 🤣

Good idea😏

iguoyr: > Exporting the struct but only for internal usage still commonly exists.

Yes, but I think it's not a best practice. I agree with @Hexilee, if still using `SelectImpl`, that means it could be invoked by outside packages. The struct `Selector` helps to reflect the detail `xxxSelector`, but it doesn't prevent to invoke `xxxSelector` directly. And if we don't want to invoke directly, maybe move `pkg/selector/container` package or create a legacyPodSelector when `Select` in `pkg/selector/container` are better choices.

What do you think? @YangKeao

STRRL:
I am going to close this PR because we have lost the context of doing this.

PR #3019, opened 2022-03-14
chore: new component checkbox for boolean value

STRRL:

- new component  `CheckBox` in mui-extends for representing a boolean value

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


ui/packages/mui-extends/src/Checkbox/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+  helperText?: string
+  /**
+   * Validation error message, would be shown on the bottom of the checkbox.
+   */
+  errorMessage?: string
+}
+
+export default (props: CheckboxProps) => {
+  return (
+    <div>
```
g1eny0ung, line 44: `<div></div>` is redundant.

ui/packages/mui-extends/src/Checkbox/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+      <FormControl error={props.errorMessage !== ''} required={true}>
+        <FormLabel component="legend">{props.label}</FormLabel>
```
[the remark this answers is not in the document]
g1eny0ung: Except for a checkbox group.

ui/packages/mui-extends/src/Checkbox/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+import { FormControl, FormControlLabel, FormHelperText, FormLabel } from '@mui/material'
+import { default as MuiCheckbox, CheckboxProps as MuiCheckboxProps } from '@mui/material/Checkbox'
+
+export type CheckboxProps = MuiCheckboxProps & {
+  /**
+   * The name of ths input field when using it in a form.
+   */
+  name?: string
```
g1eny0ung, line 25: `MuiCheckboxProps` has already defined a `name` field. So this part is redundant.

ui/packages/mui-extends/src/Checkbox/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <div>
+      <FormControl error={props.errorMessage !== ''} required={true}>
+        <FormLabel component="legend">{props.label}</FormLabel>
+        <FormControlLabel
+          control={
+            <MuiCheckbox
+              checked={props.checked}
+              onChange={props.onChange}
+              disabled={props.disabled}
+              inputProps={{ name: props.name }}
```
g1eny0ung, line 53: As above says, `name` can be handled with spread props, like:

```tsx
export default ({ label, helperText, ...rest}: CheckboxProps) => {
  return (
    // ...
    <FormControlLabel
      label={label}
      control={<MuiCheckbox {...rest} />} />
    // ...
  )
}
```

ui/packages/mui-extends/src/Checkbox/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+            ></MuiCheckbox>
+          }
+          label={props.helperText || props.label}
+        ></FormControlLabel>
+
+        <FormHelperText>{props.errorMessage}</FormHelperText>
```
g1eny0ung, line 59: I think you wanna render an optional error message when a checkbox has an error? 🤔 Then I think the following would be a better way:

```jsx
<FormHelperText error={props.error}>
  {error && props.errorMessage ? props.errorMessage : props.helperText}
</FormHelperText>
```

ui/packages/mui-extends/src/Checkbox/index.tsx
```diff
@@ -0,0 +1,63 @@
...
+        <FormControlLabel
+          control={
+            <MuiCheckbox
+              checked={props.checked}
+              onChange={props.onChange}
+              disabled={props.disabled}
+              inputProps={{ name: props.name }}
+            ></MuiCheckbox>
+          }
+          label={props.helperText || props.label}
```
g1eny0ung, line 56: If you want to fill the `helperText` when `label` is unset, then theoretically it should be reversed:

```
props.label || props. helperText
```

g1eny0ung: But actually `label` is required.

ui/packages/mui-extends/src/Checkbox/index.tsx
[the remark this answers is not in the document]
STRRL: I built this component with the reference prototype on figma:

image

It seems the text `Continous` is the `label`, and the next text near the checkbox is `help text`.

So I made this. Does it make sense?

ui/packages/mui-extends/src/Checkbox/index.tsx
STRRL: That could work with most other components like TextField.  It's also related to the prototype, if the text right beside of checkbox means `helpText`, I think maybe showing errorMessage in other places is better.

Another thing is we do not have both `error` and `errorMessage`, we have only `errorMessage` now. Maybe declaring both of them is better?

ui/packages/mui-extends/src/Checkbox/index.tsx
STRRL: 🤣 Just conversely, I want to use label as the fallback when helperText is not provided, because label is required.

ui/packages/mui-extends/src/Checkbox/index.tsx
STRRL: Updated! PTAL

ui/packages/mui-extends/src/Checkbox/index.tsx
STRRL: Updated! PTAL

ui/packages/mui-extends/src/Checkbox/index.tsx
STRRL: Updated! PTAL

ui/packages/mui-extends/src/Checkbox/index.tsx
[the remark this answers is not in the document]
g1eny0ung: Got it. It's a `subtitle`. It's no need to write in a form component.

A little more intuitive, with markdown the above prototype can be translated as:

```md
### Continous

Checkbox
```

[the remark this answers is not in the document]
STRRL: I get that!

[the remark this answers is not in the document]
STRRL: Updated! PTAL

ui/packages/mui-extends/src/Checkbox/index.tsx
g1eny0ung: > That could work with most other components like TextField. It's also related to the prototype, if the text right beside of checkbox means `helpText`, I think maybe showing errorMessage in other places is better.
> 
> Another thing is we do not have both `error` and `errorMessage`, we have only `errorMessage` now. Maybe declaring both of them is better?

Follow https://mui.com/components/checkboxes/#label, it means a `label`, not a `helperText`. I can provide an example of how the mui represents an error when it has occurred:

https://github.com/chaos-mesh/chaos-mesh/blob/821d65d78e45eb340cec9816154b0b4f86ba1c48/ui/app/src/components/NewExperimentNext/Step2.tsx#L148

Usually, the `helperText` will perform two responsibilities, one for rendering a helper text and another is rendering an error message when an error occurred.

And the `error` will tell a form component if there has an error. You can find the `error` props in https://mui.com/api/text-field/.

STRRL: Got that,  I would update later.

STRRL: Updated! PTAL! :D

ui/packages/mui-extends/src/Checkbox/index.tsx
STRRL: This conversation has been outdated as codes already changed.

STRRL: and we would not borther on this problem.

g1eny0ung (approved):
LGTM

fewdan (approved):
LGTM

PR #3036, opened 2022-03-17
selector: percent selection should select at least one

YangKeao:

This bug was originally found in `NetworkChaos`:

If the user is using the `maxPercentMode` for both `.` and `.Target`, and blessed by the goddess of fortune, one of them doesn't select any pod. The other one will be injected for all traffic, which is not expected. After considering some time, I thought it would be reasonable to generate at least one target in the selector (if it's at least able to.)

In this PR, I modified two things:

1. Replace the `Floor` in selector into `Ceil`
2. Check the emptiness for every selector, but not only for the total records.


pkg/selector/generic/mode.go
```diff
@@ -85,7 +86,7 @@ func FilterObjectsByMode(mode v1alpha1.SelectorMode, value string, count int) ([
 		}
 
 		percentage := getRandomNumber(maxPercentage + 1) // + 1 because Intn works with half open interval [0,n) and we want [0,n]
-		num := int(math.Floor(float64(count) * float64(percentage) / 100))
+		num := int(math.Ceil(float64(count) * float64(percentage) / 100))
```
STRRL, line 89: Should we remove the `maxPercentage+1` and update the comments?

YangKeao: No. Consider the special situation where the `maxPercentage` is `100`, then the `percentage` will be `0-100` (as the `getRandomNumber` always returns the number in `[0,101)`), which is actually expected.

YangKeao: @iguoyr PTAL

iguoyr (approved):
LGTM

cwen0 (approved):
LGTM @YangKeao You need to update docs for this.

PR #3038, opened 2022-03-17
bump gqlgen to v0.17.2

Hexilee:
Signed-off-by: xixi <i@hexilee.me>

### What problem does this PR solve?

[go 1.18](https://tip.golang.org/doc/go1.18) is released! It introduces generics, which may improve our development effectiveness.

However, the gqlgen v0.13 cannot work on go 1.18, so we should upgrade it first.

### Related changes

- [ ] Need to update `chaos-mesh/website`
- [ ] Need to update `Dashboard UI`
- Need to **cheery-pick to release branches**
  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code
- [ ] Manual test (add steps below)

Side effects

- [ ] Breaking backward compatibility

### Release note 

```text
Please add a release note.

You can safely ignore this section if you don't think this PR needs a release note.
```

### DCO

If you find the DCO check fails, please run commands like below (Depends on the actual situations. For example, if the failed commit isn't the most recent) to fix it:

```shell
git commit --amend --signoff
git push --force
```

PR #2986, opened 2022-03-07
e2e test on GitHub Action

STRRL:
Signed-off-by: STRRL <str_ruiling@outlook.com>

### What problem does this PR solve?

This PR try to move e2e tests from self-hosted Jenkins to GitHub Action, the latter one provides a better network environment.

### What's changed and how it works?

- create a new GitHub Action workflow
- using a test reporter to generate test report

known issue:

- test reporter would insert the test report to an incorrect position. https://github.com/dorny/test-reporter/issues/67

### Related changes

- [ ] Need to update `chaos-mesh/website`
- [ ] Need to update `Dashboard UI`
- Need to **cheery-pick to release branches**
  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [x] E2E test
- [ ] No code
- [x] Manual test (add steps below)

Side effects

- [ ] Breaking backward compatibility

### Release note 

```text
Please add a release note.

You can safely ignore this section if you don't think this PR needs a release note.
```

### DCO

If you find the DCO check fails, please run commands like below (Depends on the actual situations. For example, if the failed commit isn't the most recent) to fix it:

```shell
git commit --amend --signoff
git push --force
```

PR #3055, opened 2022-03-24
bump go to 1.18

Hexilee:

[go 1.18](https://tip.golang.org/doc/go1.18) is released! It introduces generics, which may improve our development effectiveness.

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


Hexilee:
wait for #3038

STRRL:
please append an description in CHANGELOG

Hexilee:
> please append an description in CHANGELOG

Done

YangKeao (approved):
LGTM

Andrewmatilde:
Please fix build error in e2e-test.

STRRL:
> Please fix build error in e2e-test.

It does not look like a build error, images could be built successful, and [the test report](https://ci.pingcap.net/blue/organizations/jenkins/chaos-mesh-e2e/detail/chaos-mesh-e2e/2199/tests) shows several messages after kind startup:

```
gzip: stdin: unexpected end of file
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
tar: Error is not recoverable: exiting now
```

I downloaded the logs but find nothing useful...

I prefer to think that `kind` failed to load image from host docker into kind docker.

Anyhow, I would try to trigger the e2e tests again.

Andrewmatilde:
> > Please fix build error in e2e-test.
> 
> It does not look like a build error, images could be built successful, and [the test report](https://ci.pingcap.net/blue/organizations/jenkins/chaos-mesh-e2e/detail/chaos-mesh-e2e/2199/tests) shows several messages after kind startup:
> 
> ```
> gzip: stdin: unexpected end of file
> tar: Unexpected EOF in archive
> tar: Unexpected EOF in archive
> tar: Error is not recoverable: exiting now
> ```
> 
> I downloaded the [logs](https://ci.pingcap.net/blue/organizations/jenkins/chaos-mesh-e2e/detail/chaos-mesh-e2e/2199/artifacts/) but find nothing useful...
> 
> I prefer to think that `kind` failed to load image from host docker into kind docker.
> 
> Anyhow, I would try to trigger the e2e tests again.

I agree with you , but I think it is not a good idea to skip the error now , the problem may happen in any other PRs after this merge.

STRRL:
> I agree with you , but I think it is not a good idea to skip the error now , the problem may happen in any other PRs after this merge.

Which error? Did I miss anything? 🤪

Andrewmatilde:
> > I agree with you , but I think it is not a good idea to skip the error now , the problem may happen in any other PRs after this merge.
> 
> Which error? Did I miss anything? 🤪

I think the error is , the e2e-test is totally unstable.

STRRL:
> I think the error is , the e2e-test is totally unstable.

We would migrate e2e tests onto GitHub Action which includes a much better networking environment. It would reduce the failure caused by the networking issue. 🥲

> even after migration, I think we also need time to turn the performance of it. 

But the logic of e2e tests itself make it would not be stable as unit tests: it makes asserts based on time. And I do NOT think it would be better in short future(several weeks). 🥲

After we received the computing resources from CNCF, I think we could start to think about how to make the CI experience better, like test reporting, and statistics.

Andrewmatilde:
> 

I agree with you . It seems at least two PR have e2e-test problem now. Can we just ingore these error & merge ?

STRRL:
> I agree with you . It seems at least two PR have e2e-test problem now. Can we just ingore these error & merge ?

I suggest NEVER do that. IMO, bypassing these CI would bring the challenge/break the determined checks. 🥲🥲🥲

I think we should suffer from it, build the correct recognition of CI, which is the last line between human's wild thoughts and engineering. 

Until we improve/fix the performance of CI.

STRRL:
Hi @Hexilee, I think we should update go version with `actions/setup-go`: .github/workflows/integration_test.yml:22

There are still several recipes in makefile still not using `dev-env`.

/cc @YangKeao

Hexilee:
> 

Oops, updated.

STRRL (approved):
LGTM!

Hexilee:
E2E tests Failed too many times, maybe we should wait for #2986

STRRL:
> E2E tests Failed too many times, maybe we should wait for #2986

It seems the networking Jenkins back to normal, let me take another retry

Issue #3069, opened 2022-03-28
allow to modify the `oom_score_adj` value of the stress process


YangKeao:
## Feature Request

**Is your feature request related to a problem? Please describe:**

I want to use the `StressChaos` to simulate the situation when pod is killed by the oom killer, but I cannot as the `oom_score_adj` of the stress process is higher than the workload (and I don't want to modify the score adj of the target container process). 

**Describe the feature you'd like:**

Allow the user to set the `oom_score_adj` of the stress process. We could use to `choom` to implement this feature.

- [x] Implement this feature https://github.com/chaos-mesh/chaos-mesh/pull/3070
- [ ] Add documents for the additional field.

PR #3070, opened 2022-03-28
stress: add field to set the oom score

YangKeao:

#3069, but still need to modify the document after this PR merged.

Add a new field to represent the `oom_score_adj` value of the mem stress process. The `chaos-daemon` will use `choom` to execute the process and set the `oom_score_adj`. If this value is set to `-1000`, the oom killer will always kill the target container.

### Related changes


controllers/chaosimpl/stresschaos/impl.go
```diff
@@ -74,13 +74,17 @@ func (impl *Impl) Apply(ctx context.Context, index int, records []*v1alpha1.Reco
...
+	req := pb.ExecStressRequest{
 		Scope:           pb.ExecStressRequest_CONTAINER,
 		Target:          containerId,
 		CpuStressors:    cpuStressors,
 		MemoryStressors: memoryStressors,
 		EnterNS:         true,
-	})
+	}
+	if stresschaos.Spec.Stressors.MemoryStressor != nil {
+		req.OomScoreAdj = int64(stresschaos.Spec.Stressors.MemoryStressor.OOMScoreAdj)
```
FingerLeader, line 85: Why use `int64` here? I found that everything else uses `int` except for proto.

YangKeao: Because the size of `int` depends on the implementation but is not part of the go language, the `gRPC` only allows using `int32` or `int64`.

But anyway, I should use `int32` here. I have submitted a commit to turn it into `int32`.

cwen0:
@YangKeao Can you fix CI first?

YangKeao:
> @YangKeao Can you fix CI first?

Fixed.

cwen0 (approved):
LGTM @YangKeao Remember to create an issue to trace the implementation of this feature on chaos dashboard.

PR #3078, opened 2022-03-30
Delete the CI Test install.sh

FingerLeader:

Because `script_test.yml` uses the previous image, this can cause the CI to fail when adding new flags to the components. And `install.sh` will be checked in [integration test](https://github.com/chaos-mesh/chaos-mesh/blob/master/.github/workflows/integration_test.yml#L52]), there is no need for repeat testing.
Remove `script_test.yml`


cwen0 (approved):
LGTM

WangXiangUSTC (approved):
LGTM

PR #3120, opened 2022-04-11
Fix clusterrole name in DNS server

anuragpaliwal80:

Fix the cluster role name for DNS server 

Use the correct cluster role name

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/helm/chaos-mesh/templates/dns-deployment.yaml
+++ b/helm/chaos-mesh/templates/dns-deployment.yaml
@@ -44,7 +44,7 @@ spec:
 {{ toYaml . | indent 8 }}
     {{- end }}
     spec:
-      serviceAccountName: chaos-dns-server
+      serviceAccountName: {{ .Values.dnsServer.serviceAccount }}
       affinity:
         podAntiAffinity:
           preferredDuringSchedulingIgnoredDuringExecution:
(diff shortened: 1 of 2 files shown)


iguoyr (approved):
LGTM

fewdan (approved):
LGTM

STRRL:
I notice that the service account  `{{ .Values.dnsServer.serviceAccount }}` (which binded with ClusterRole `{{ .Release.Name }}-chaos-dns-server-target-namespace`) actually has no usage.

It seems that we could purge the `dns-rbac.yaml`.

/cc @WangXiangUSTC

WangXiangUSTC:
> I notice that the service account `{{ .Values.dnsServer.serviceAccount }}` (which binded with ClusterRole `{{ .Release.Name }}-chaos-dns-server-target-namespace`) actually has no usage.
> 
> It seems that we could purge the `dns-rbac.yaml`.
> 
> /cc @WangXiangUSTC

It is used here https://github.com/chaos-mesh/chaos-mesh/blob/59cfaba0e27a9b798a64fe1e99c5fc98ffbd887d/helm/chaos-mesh/templates/dns-deployment.yaml#L47
But I fill it with a static value by mistake, need to update it with `{{ .Values.dnsServer.serviceAccount }}` 😅

WangXiangUSTC (approved):
LGTM

Issue #3155, opened 2022-04-19
create JVMChaos in workflow failed


WangXiangUSTC:
## Bug Report

**What version of Chaos Mesh are you using?**

v2.1.3

**What did you do?**

create JVMChaos in workflow

**What did you expect to see?**
create success

**What did you see instead?**

failed, the error logs:

```txt
2022/04/15 08:10:31 http: panic serving 172.18.0.2:29032
: interface conversion: interface {} is *v1alpha1.Workfl
ow, not *v1alpha1.JVMChaos
goroutine 2487 [running]:
net/http.(*conn).serve.func1()
        /usr/local/go/src/net/http/server.go:1825 +0xb8
panic({0x1a09220, 0x400031f5f0})
        /usr/local/go/src/runtime/panic.go:844 +0x260
github.com/chaos-mesh/chaos-mesh/api/v1alpha1.(*JVMChaos
Spec).Default(0x1ab6340?, {0x1ca96c0?, 0x4000da9800?}, 0
x0?)
```

PR #3156, opened 2022-04-19
fix the bug that create JVMChaos failed in workflow

WangXiangUSTC:

fix issue #3155 

`JVMChaosSpec`'s root may not be `JVMChaos`, it can also be `Workflow`.

### Related changes

  - [x] release-2.1
  - [ ] release-2.0

### Checklist

Tests

- [ ] Unit test
- [x] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


STRRL (approved):
LGTM!

STRRL:
Validation with `root` might be dangerous, and there are also other similar usages like `IOErrno`, `EbsVolume`, which would be a potential issue.   

FYI @YangKeao

Andrewmatilde:
IMO, we should also reduce code with reflction.

Andrewmatilde (approved):
LGTM

PR #3170, opened 2022-04-21
add more status for record

### LINKED ISSUE #3127: more clear status to tell if the experiment succeed
## Feature Request

from existing status, we can hardly to tell if the experiment succeed, or just duration exceeded. e.g.
```yaml
  spec:
    clockIds:
    - CLOCK_REALTIME
    duration: 10s
    mode: one
    selector:
      labelSelectors:
        app: nginx
      namespaces:
      - xlgao
    timeOffset: 10s
  status:
    conditions:
    - reason: ""
      status: "True"
      type: AllRecovered
    - reason: ""
      status: "False"
      type: Paused
    - reason: ""
      status: "True"
      type: Selected
    - reason: ""
      status: "False"
      type: AllInjected
    experiment:
      containerRecords:
      - id: xlgao/nginx-644467cc8b-dv25r/centos
        phase: Not Injected
        selectorKey: .
      desiredPhase: Stop
```

if we do not see the events, we may not know if the record had been injected in the duration.

we need a clearer way to know if the record had been injected in the duration.

since a experiment is a stateful resource, I think we should add more fields in status. for this case ,maybe add a bool field in status like `durationExceeded` or something 

xlgao-zju:

 Close #3127

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [x] E2E test
- [ ] No code

create experiment, pause and resume it

image

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


xlgao-zju:
/cc @cwen0 @STRRL @YangKeao @Hexilee 
for review

api/v1alpha1/common_types.go
```diff
@@ -75,6 +76,12 @@ type Record struct {
 	Id          string `json:"id"`
 	SelectorKey string `json:"selectorKey"`
 	Phase       Phase  `json:"phase"`
+	// InjectedCount is a counter to record the sum of successful injections
+	InjectedCount int `json:"injectedCount"`
+	// RecoveredCount is a counter to record the sum of successful recoveries
+	RecoveredCount int `json:"recoveredCount"`
+	// Events are the essential details about the injections and recoveries
+	Events []*RecordEvent `json:"events,omitempty"`
```
STRRL, line 84:
```suggestion
	// Events are the essential details about the injections and recoveries
	Events []RecordEvent `json:"events,omitempty"`
```

I think the element in this slice could not be `nil`, and we would not update the element, so we do not require to use "slice of pointers".

api/v1alpha1/common_types.go
```diff
@@ -86,6 +93,33 @@ const (
...
+
+type RecordEventStage string
+
+const (
+	// Injection means this event is recorded, when we inject the chaos
+	// typically, when we call impl.Apply()
+	Injection RecordEventStage = "Injection"
+	// Recovery means this event is recorded, when we recover the chaos
+	// typically, when we call impl.Recover()
+	Recovery RecordEventStage = "Recovery"
```
STRRL, line 120: Maybe we could use the words "Apply" and "Recover" (instead of "Injection" and "Recovery" )as stage names for reducing thinking?

api/v1alpha1/common_types.go
```diff
@@ -86,6 +93,33 @@ const (
 	Injected Phase = "Injected"
 )
 
+type RecordEvent struct {
+	Type      RecordEventType  `json:"type"`
+	Stage     RecordEventStage `json:"stage"`
+	Message   string           `json:"message,omitempty"`
+	Timestamp *metav1.Time     `json:"timestamp"`
+}
```
STRRL, line 101: please append comments for these fields. :D

STRRL (commented):
rest LGTM!

STRRL:
after discussing with @xlgao-zju, the capacity of `events` might be a problem **now**

There are two reasons that would make `events increase fast:

- "one-to-many" chaos-records relation; when a chaos experiment uses a selector for selecting a big mount of targets, there would be many "records" in the status
- uncontrolled retry/reconcile when failed to `Apply`/`Recover`; we would **always requeue** the reconcile request if the operation failed, and only using the [default rate limiter](https://github.com/kubernetes-sigs/controller-runtime/blob/c162794a9b12dea31c076820938a779873e93957/pkg/controller/controller.go#L145-L147) of controller-runtime. So there would be many events with "similar" messages appearing in a short time.

Maybe we need an aggregation or rotate policy in the future. I still stay positive about the effect of this problem, I think we do not need to consider it in this PR.

How do you think about it? @YangKeao @Hexilee @xlgao-zju

xlgao-zju:
at the beginning, I append the events, when we apply failed. and then I get too many error messages. and the reason is:

apply failed  -> update the status -> the update will let the object be reconciled -> apply failed

since this is circle, so we will get so so many record.events 😂 

in order to avoid thi bug, I did those:
- make the common CRD's status as a sub resource(or modify the status will increase the generation of the resource)
- for common CRDs, use `WithEventFilter` to filter the events,  pick up the event only when label/annotation/spec changes
-  for the CRDs controlled by common CRDs, there is no filter. since when the child CR status changes, we need to reconcile the according parent CR
- add another condition step after record step, since if we will not reconcile the object, if only the status changes, the conditions calculated before record step may not be correct after the record step

the PR is ready for review
@STRRL @YangKeao @Hexilee

xlgao-zju:
kindly ping @STRRL @YangKeao @Hexilee for review

STRRL:
PTAL @YangKeao , this PR introduces Status Subresources for all the chaos CRD. But I forgot why do we NOT use status subresources before. Maybe after navirna we could use subresrouce without any blocking?🤔

STRRL:
> after discussing with @xlgao-zju, the capacity of `events` might be a problem **now**
> 
> There are two reasons that would make `events increase fast:
> 
> * "one-to-many" chaos-records relation; when a chaos experiment uses a selector for selecting a big mount of targets, there would be many "records" in the status
> * uncontrolled retry/reconcile when failed to `Apply`/`Recover`; we would **always requeue** the reconcile request if the operation failed, and only using the [default rate limiter](https://github.com/kubernetes-sigs/controller-runtime/blob/c162794a9b12dea31c076820938a779873e93957/pkg/controller/controller.go#L145-L147) of controller-runtime. So there would be many events with "similar" messages appearing in a short time.
> 
> Maybe we need an aggregation or rotate policy in the future. I still stay positive about the effect of this problem, I think we do not need to consider it in this PR.
> 
> How do you think about it? @YangKeao @Hexilee @xlgao-zju

it seems this issue is NOT resolved properly, I could still reproduce it with these manifests:

```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
spec:
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - image: nginx:latest
        imagePullPolicy: Always
        name: nginx
        resources: {}
---
kind: IOChaos
metadata:
  name: io-delay-example
spec:
  action: latency
  mode: all
  selector:
    labelSelectors:
      app: nginx
  volumePath: /not-existed
  path: /not-existed/**/*
  delay: "10ms"
  percent: 10
```

STRRL:
It seems the predicates DO work. But the soo many events appended and resoruceVersion increased near to thousands. I am still trying to figure it out with logs. 🤔~/temp/chaos-controller-manager.log

xlgao-zju:
look like the controller runtime do not set gvk in the object....
image
image

so we can use type assertions to determine if it is a podxxxchaos

STRRL:
A little wired, the e2e test cases of networkchaos failed, I still try to figure out the reason.

xlgao-zju:
@STRRL looks like the network chaos has not been deleted, before we create a new one... so we get this error
```golang
object is being deleted: networkchaos.chaos-mesh.org "network-chaos-1" already exists
      occurred

      /home/runner/work/chaos-mesh/chaos-mesh/e2e-test/e2e/chaos/networkchaos/network_delay.go:139
```

STRRL:
> @STRRL looks like the network chaos has not been deleted, before we create a new one... so we get this error
> 
> ```go
> object is being deleted: networkchaos.chaos-mesh.org "network-chaos-1" already exists
>       occurred
> 
>       /home/runner/work/chaos-mesh/chaos-mesh/e2e-test/e2e/chaos/networkchaos/network_delay.go:139
> ```

I think I might be figure out the reasons: we should also trigger the Reconcile with the changes on "DeletionTimestamp":

```
		predicaters := []predicate.Predicate{
			predicate.LabelChangedPredicate{},
			predicate.AnnotationChangedPredicate{},
			predicate.GenerationChangedPredicate{},
			DeletionTimestampUpdatePredicate{},
		}
```

```
type DeletionTimestampUpdatePredicate struct {
	predicate.Funcs
}

func (DeletionTimestampUpdatePredicate) Update(e event.UpdateEvent) bool {
	return e.ObjectNew.GetDeletionTimestamp() != e.ObjectOld.GetDeletionTimestamp()
}
```

I have tested it on my local cluster, it passed the e2e tests.  : )

STRRL:
Also, I found that running e2e tests on a local machine becomes not so-fluent now, maybe it's time to update the developer documentation and makefile again. (I would raise a new issue tomorrow.

xlgao-zju:
> we should also trigger the Reconcile with the changes on "DeletionTimestamp":
@STRRL will add. but if `DeletionTimestamp` will not trigger the Reconcile, how could we delete other chaos(since other chaos got passed).

STRRL:
> > we should also trigger the Reconcile with the changes on "DeletionTimestamp":
> 
> @STRRL will add. but if `DeletionTimestamp` will not trigger the Reconcile, how could we delete other chaos(since other chaos got passed).

I am afraid that other tests cases only assert the "effect of chaos experiment has been gone" but not make assert on "the custom resource should be deleted in a properly duration"

So the behavior of failed testcase that I observed is, that one NetworkChaos has been fully recovered (condition AllRecovered is True), but when I execute `kubectl delete`, it would be stucked.

It seems 2 failed test cases ("[NetworkChaos] [NetworkPartition] [Schedule]"and "[NetworkChaos] [Netem] [Schedule]") are the only test cases cover these situation: "deleting one chaos custom resource, then create another resource with same name"

xlgao-zju:
@STRRL after add `DeletionTimestamp` predicate, I think we can pass all the test case. and as you said, if we can run e2e test smoothly on local, it must be very cool. ❤️

STRRL:
Update: the required changes might be much more than "DeletionTimestamp predicate" 🤔. 

Even though it could pass all the e2e test cases. The example under `exampl/nginx` still could NOT be deleted as expected.

It seems we still require changes on `Status.Experiment.Records[].Phase` to trigger the Reconcile.

xlgao-zju:
> It just works with "DeletionTimestamp predicate", but why. 😇

set DeletionTimestamp -> reconcile -> delete child cr resource and set phase `recover/wait` -> child cr is deleted, and get a delete event -> reconcile -> finish

@STRRL like this?

xlgao-zju:
> Once Reconcile was triggered by "Setup DeletionTimeStamp", the rest process could be triggered by the updates of "PodXXXChaos".

yep, that's why we should pick up all the child CRDs events. 😆

STRRL:
> > It just works with "DeletionTimestamp predicate", but why. innocent
> 
> set DeletionTimestamp -> reconcile -> delete child cr resource and set phase `recover/wait` -> child cr is deleted, and get a delete event -> reconcile -> finish
> 
> @STRRL like this?

Mostly. But the child PodXXChaos custom resource would not be deleted sometime, for PodHTTPChaos, it would only be updated.

xlgao-zju:
looks like I can remove the `RecordPhaseChangePredicate` which I just add, haha

STRRL (commented):
rest LGTM!

controllers/common/step.go
```diff
@@ -24,5 +24,5 @@ import (
 )
 
 func AllSteps() []pipeline.PipelineStep {
-	return []pipeline.PipelineStep{finalizers.Step, desiredphase.Step, condition.Step, records.Step}
+	return []pipeline.PipelineStep{finalizers.Step, desiredphase.Step, condition.Step, records.Step, condition.Step}
 }
```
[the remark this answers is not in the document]
xlgao-zju: @STRRL  when I develop  this patch, i tried to remove the first `condition `, but it will be a bug. when we pause a experiment(which keep failing to inject), the object will be requeued in the `records` step, and not enter the `condition` step. then, the `Paused` condition will still be false...

so, I just add another `condition` step after the `records` step. and did not remove the first one.

xlgao-zju:
@STRRL updated according to your comment. PTAL

YangKeao:
> PTAL @YangKeao , this PR introduces Status Subresources for all the chaos CRD. But I forgot why do we NOT use status subresources before. Maybe after navirna we could use subresource without any blocking?thinking

@STRRL Before @Hexilee piped all controllers in a single pipeline, they are separated controllers and will trigger each other (which has been proved to be a whimsical idea/decision). At that time, using subresource is impossible because the status change will not trigger any other `Reconcile`.

 Only considering the `*Chaos` itself, it will be likely to work after setting the subresource, as it only has one controllers. But other controllers (like `Workflow` and `Schedule`) will listen to the `*Chaos`'s status (at least `Schedule` will), I'm not sure whether it will work.

I'll manually test `Schedule` in this PR (though there are already some tests about it).

YangKeao:
It seems that a more prudent way is to only set predicate and ignore the changes happened on events (and I think that's why we want to set `SubResource` in this PR?), but it's also a great chance for us to think about the `subresource` and decide whether to use it.

STRRL:
> At that time, using subresource is impossible because the status change will not trigger any other Reconcile.

🤔, well, it seems the status changes could also trigger the reconcile, until we setup predicate like `predicate.GenerationChangedPredicate`, `predicate.AnnotationChangedPredicate`, etc...

> they are separated controllers and will trigger each other (which has been proved to be a whimsical idea/decision).

off-topic: I still think running separated controllers would work with well-designed status and reconcilers. Using "pipeline" is just a compatible and easy-implemented solution.

STRRL:
> It seems that a more prudent way is to only set predicate and ignore the changes happened on events

+1; If we want to merge this PR quickly, I think using that predicate is easier than setup the status subresource.

cc @xlgao-zju

YangKeao:
> thinking, well, it seems the status changes could also trigger the reconcile, until we setup predicate like predicate.GenerationChangedPredicate, predicate.AnnotationChangedPredicate, etc...

@STRRL Sorry I made a mistake. It will.

xlgao-zju:
> It seems that a more prudent way is to only set predicate and ignore the changes happened on events

If we want to ignore the changes of `events` without subresource, I think we should also ignore the change of `resourceVersion` and `generation`.

> but it's also a great chance for us to think about the subresource and decide whether to use it.

agree

xlgao-zju:
@STRRL @YangKeao I have remove the "sub resource" part. and add a predicate to skip the events which ONLY object.status.experiment.records[].events changed

PTAL

controllers/common/step.go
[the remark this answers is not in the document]
STRRL: Maybe this change is not required after we do not use the Status Subresrouce

controllers/common/fx.go
```diff
@@ -145,3 +153,54 @@ func Bootstrap(params Params) error {
...
+// Update implements UpdateEvent filter for update deletionTimestamp.
+func (StatusRecordEventsChangePredicate) Update(e event.UpdateEvent) bool {
+	objNew, ok := e.ObjectNew.DeepCopyObject().(v1alpha1.StatefulObject)
+	if !ok {
+		return true
+	}
+	objOld, ok := e.ObjectOld.DeepCopyObject().(v1alpha1.StatefulObject)
+	if !ok {
+		return true
+	}
```
STRRL, line 189: I prefer to return `false` here.

Because we would use `Or()` for all the predicates, so each predicate would implement the "restricted" way. For this one, it should only allow:
- the event about "Update" on certain Chaos Resource, `v1alpha1.StatefulObject` 
- there are only changes on `status.Experiment.Records[].Events`

STRRL (commented):
rest LGTM!

PTAL @xlgao-zju

xlgao-zju:
@STRRL updated

controllers/common/fx.go
```diff
@@ -145,3 +153,54 @@ func Bootstrap(params Params) error {
...
+	return false
+}
+
+// StatusRecordEventsChangePredicate skip the update event,
+// when we Only update object.status.experiment.records[].events
+type StatusRecordEventsChangePredicate struct {
+	predicate.Funcs
+}
+
+// Update implements UpdateEvent filter for update deletionTimestamp.
```
YangKeao, line 180: It seems that the comment is not related with the implementation? It has nothing to do with `deletionTimestamp`

YangKeao (commented):
Rest LGTM

controllers/common/fx.go
xlgao-zju: oposs ... I forget to update this... `deletionTimestamp` will introduced by this comment https://github.com/chaos-mesh/chaos-mesh/pull/3170#issuecomment-1120231330

xlgao-zju:
@YangKeao Updated

YangKeao (approved):
LGTM

STRRL (approved):
LGTM!

Issue #3173, opened 2022-04-21
Native ARM64 CI/CD


YangKeao:
Thanks for CNCF, we got a powerful ARM64 machine and bootstrapped some github action runner. Now we can add / migrate some building / testing tasks to the native environments :smile:.

- [ ] Add integration test / unit test for arm
- [ ] Add format check for arm
- [ ] Migrate the nightly image release to the arm (or possible release image for every commit in master)

lukas8219:
Guys - are we ever continuing the project of having ARM builds for `chaosd` binary?

STRRL:
> Guys - are we ever continuing the project of having ARM builds for `chaosd` binary?

Hi @lukas8219, we're not put lots of effort on `chaod` so far, but contribution is welcome! 🤗

PR #3177, opened 2022-04-22
run ci and integration test task on arm

YangKeao:

Run ci workflow on arm


.github/workflows/integration_test.yml
```diff
@@ -7,16 +7,24 @@ on:
...
+    runs-on: ${{ fromJson('{"x86_64":"ubuntu-latest", "aarch64":["self-hosted", "Linux", "ARM64"]}')[matrix.arch] }}
     steps:
       - uses: actions/checkout@master
         with:
           # Must use at least depth 2!
           fetch-depth: 2
 
       - name: Create kind cluster
-        uses: helm/kind-action@v1.2.0
+        uses: helm/kind-action@4c7909140acfc81a05fc96fed8fea6673ba8ce80
```
WangXiangUSTC, line 21: Why use this commit instead of a released version? It might need a comment.

WangXiangUSTC (commented):
rest LGTM

.github/workflows/ci.yml
```diff
@@ -45,13 +45,14 @@ jobs:
...
     needs: changes
     if: ${{ needs.changes.outputs.go == 'true' }}
-    runs-on: ubuntu-latest
     strategy:
       matrix:
+        arch: [x86_64, aarch64]
         job:
           - verify
           - build
           - test
```
STRRL, line 54: it seems we should also append  `fail-fast: false`

YangKeao:
Blocked by https://github.com/chaos-mesh/chaos-mesh/pull/3258. Though we can use an older version of `kubectl`, using the latest one is more elegant.

YangKeao:
The latest problem is that the kind in dind fail to connect to the network outside.

STRRL:
> The latest problem is that the kind in dind fail to connect to the network outside.

What does it mean? 🙈

YangKeao:
> > The latest problem is that the kind in dind fail to connect to the network outside.
> 
> What does it mean? see_no_evil

We run the action in a dind environment, and start a kind cluster in it. However, the kind cluster seems failed to connect to the internet, and cannot download the image from the dockerhub (or any other mirrors). Only the pod with preloaded image can be booted.

YangKeao:
Some problems found in this PR is also recorded in https://github.com/chaos-mesh/chaos-mesh/issues/3173

.github/workflows/integration_test.yml
```diff
@@ -7,16 +7,40 @@ on:
...
+      - name: Setup Helm
+        uses: azure/setup-helm@v2.1
+
+      - name: Setup Kubectl
+        uses: azure/setup-kubectl@v2.0
+
+      - name: Magic Kind DNS Fix
+        if: ${{ matrix.arch == 'arm64' }}
+        run: |
+          docker exec chart-testing-control-plane /bin/bash -c "sed -e 's/nameserver \(.*\)/nameserver 8.8.8.8/g' /etc/resolv.conf > /etc/resolv.conf.sed"
```
YangKeao, line 36: The network problem in kind is all about DNS.

`kind` has some magic scripts to fix the DNS loop in coredns when running inside a docker, see https://github.com/kubernetes-sigs/kind/blob/main/images/base/files/usr/local/bin/entrypoint#L375, and so does `minikube`. The awful thing is that: these scripts run, iptables rules are also applied successfully, but the DNS just don't work.

I have entered the environment manually, and dig/nslookup/tcpdump/... the DNS request with a lot of tools, but both the redirection source/target doesn't respond the DNS request. To further understand the problem, we could trace with the iptables, enable debug log of dockerd... It will take a lot more time to find the root, so I choose a much simpler way to solve: replace the dns server to 8.8.8.8 ...

test/integration_test/run.sh
```diff
@@ -41,12 +41,18 @@ fi
...
+            continue
+        fi
+
         run $script
     done
 else
     for name in $test_case; do
         script="$test_dir/$name/run.sh"
-	echo "run $script"
+	    echo "run $script"
```
WangXiangUSTC, line 55: can remove the space

test/integration_test/run.sh
```diff
@@ -41,12 +41,18 @@ fi
 
 if [ "$test_case" == "*" ]; then
     for script in $test_dir/$test_case/run.sh; do
+        # jvmchaos and chaosd are not supported in aarch64
+        # TODO: support JVMChaos / chaosd in aarch64, and remove this check
```
WangXiangUSTC, line 45: OK, I will fix them later

WangXiangUSTC (approved):
rest LGTM

test/integration_test/run.sh
YangKeao: I'll also replace the initial tab with spaces (as this file is indented with spaces), to keep align with former and later codes.

WangXiangUSTC:
LGTM

cwen0:
@YangKeao Remember to update the configuration of ti-chi-bot before merging this pr. https://github.com/ti-community-infra/configs/blob/main/prow/config/config.yaml

cwen0:
@YangKeao Remember to update the configuration of ti-chi-bot before merging this pr. https://github.com/ti-community-infra/configs/blob/main/prow/config/config.yaml

cwen0 (approved):
LGTM

PR #3186, opened 2022-04-25
update memStress

### LINKED ISSUE #2999: different behaviour between memStress and stress-ng after OOM Killing
## Bug Report

N/A

memStress in `master` branch
(using stress-ng with `2.1.1` as controlled group)

**What did you do?**

create a StressChaos with extremely large size aims to oom kill:

```yaml
kind: StressChaos
metadata:
  name: burn-cpu
spec:
  mode: one
  selector:
    labelSelectors:
      "app": "nginx"
  stressors:
    memory:
      size: 1000Gi
      workers: 1
  duration: "5m"
``` 

`memStress` should act like `stress-ng`:
- keep allocating memory even after oom killing

**What did you see instead?**

 `memStress` was killed, and StressChaos would also failed to recover

It relates to the process model of `stress-ng`, stress-ng would create new sub-processes (not sure process or thread) to allocate the memory, so oom killer would only kill the sub-process, then `stress-ng` main process would keel spawning new sub process.

But `memStress` would only work in one process, after oom killing, it did gone.

I think we could update the memStress, using `stress-ng` process model as a reference, for keeping the consistent behavior.

**Output of chaosctl**

FingerLeader:

Close #2999 

 Use the new `memStress` to make stressChaos behave more like it does when using ·stress-ng`

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


xlgao-zju:
@FingerLeader since stress-ng will allocate the `size`, not the `size * workers`. so, what's the behavior of `memStress`?

FingerLeader:
> @FingerLeader since stress-ng will allocate the `size`, not the `size * workers`. so, what's the behavior of `memStress`?

I have changed `memStress`'s behavior, now these workers will splite the `size`.

STRRL (approved):
LGTM!

STRRL:
PTAL

/cc @Hexilee @Andrewmatilde

Andrewmatilde (approved):
LGTM

PR #3223, opened 2022-05-06
fix(ui): isolate scopes when creating NetworkChaos

### LINKED ISSUE #3224: fix(workflow-ng): isolate `target` field and `Scope` when creating `NetworkChaos` in UI

g1eny0ung:

Close #3224

This PR distinguishes between different scopes when creating `NetworkChaos`.

Also, you can also see some non-related changes, most of these are needed for testing. So the core changes of this PR I have marked below:

https://github.com/chaos-mesh/chaos-mesh/blob/f4a20a74e354f2ac22ea95edc848a71919912367/ui/app/src/components/Scope/index.tsx#L64-L66

It checks if the `scope` key is the target field of the `NetworkChaos`.

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

CHANGELOG

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,7 +26,7 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
 
 ### Fixed
 
-- Isolate `target` field and `Scope` when creating `NetworkChaos` [#3223](https://github.com/chaos-mesh/chaos-mesh/issues/3221)
+- Isolate `target` field and `Scope` when creating `NetworkChaos` in UI [#3223](https://github.com/chaos-mesh/chaos-mesh/issues/3221)
 
 ### Security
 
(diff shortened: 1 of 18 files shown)


STRRL (commented):
LGTM

inside the component Scope, there might be a little mess about which kind of selector should we present, in lines 64 - 69; In the future, is there possible to extract one more components for render wiith leaving glue codes here? 🙈

STRRL (approved):

g1eny0ung:
> LGTM
> 
> inside the component Scope, there might be a little mess about which kind of selector should we present, in lines 64 - 69; In the future, is there possible to extract one more components for render wiith leaving glue codes here? 🙈

@STRRL Yes! To reduce unnecessary conditions, another base component can be wrapped up here and I will create an issue to track it.

WangXiangUSTC (approved):
LGTM

PR #3226, opened 2022-05-07
Use template execute rather than http.ResponseWriter.Write

anuragpaliwal80:
Signed-off-by: Anurag Paliwal <anuragpaliwal@gmail.com>

Use template execute rather than http.ResponseWriter.Write

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/e2e-test/cmd/e2e_helper/main.go
+++ b/e2e-test/cmd/e2e_helper/main.go
@@ -20,6 +20,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"html/template"
 	"io"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


anuragpaliwal80:
retest

STRRL (commented):
Actually, I prefer NOT to merge this PR, as instead, mark this issue on the lift as ignored.

In e2e-helper, we want to build a bytes-in-bytes-out ping-pong server for detecting the network connection, by receiving with UDP and sending with HTTP Resposne.

So using `html/template` to escape the original bytes would be break the bytes-in-bytes-out exception.

/cc @Hexilee @YangKeao

Hexilee:
> Actually, I prefer NOT to merge this PR, as instead, mark this issue on the lift as ignored.
> 
> In e2e-helper, we want to build a bytes-in-bytes-out ping-pong server for detecting the network connection, by receiving with UDP and sending with HTTP Resposne.
> 
> So using `html/template` to escape the original bytes would be break the bytes-in-bytes-out exception.
> 
> /cc @Hexilee @YangKeao

I agree. The bodies are not in HTML format, we should not escape them by HTML template.

PR #3231, opened 2022-05-07
[BugBash] SA1019: Remove deprecated pkg/log

anuragpaliwal80:
Remove deprecated pkg/log

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
 ### Changed
 
 - Helm charts: update validate-auth to chaos-mesh-validation-auth [#3193](https://github.com/chaos-mesh/chaos-mesh/pull/3193)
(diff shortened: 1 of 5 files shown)


STRRL:
`log.L()` is still the compatible solution for the login, ref: https://github.com/chaos-mesh/rfcs/blob/main/text/2021-12-09-logging.md

We marked it as deprecated for mention that new codes should not use `log.L()`.

And using logger provided by controller-runtime is actually the old/deprecated one, because we have nearly no control over it.
 
In general, the priorities of should be "`logger` in the struct" > `log.L()` > any other global loggers(package level logger, logger from klog/controller-runtime, etc..)

STRRL:
so, I am going to close this PR. :)

If you have any other questions, free feel to comment and reopen it! :D

PR #3236, opened 2022-05-07
[BugBash] SC2164: Fix cd fails case issue in bootstrap.sh

anuragpaliwal80:
Fix cd fails case issue in bootstrap.sh

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/ui/scripts/bootstrap.sh
+++ b/ui/scripts/bootstrap.sh
@@ -51,7 +51,7 @@ elif [[ ! -f $CHAOS_DASHBOARD_BIN/chaos-dashboard ]]; then
 
   cd ..
   IN_DOCKER=1 make images/chaos-dashboard/bin/chaos-dashboard
-  cd -
+  cd ... || return
 else
   echo "Already build chaos-dashboard."
 fi


ui/scripts/bootstrap.sh
```diff
@@ -51,7 +51,7 @@ elif [[ ! -f $CHAOS_DASHBOARD_BIN/chaos-dashboard ]]; then
 
   cd ..
   IN_DOCKER=1 make images/chaos-dashboard/bin/chaos-dashboard
-  cd -
+  cd ... || return
```
YangKeao, line 54: As this is the last line of the script, maybe we could remove this `cd`? @g1eny0ung 

`cd ... || return` looks like an invalid bash script (`cd` doesn't accept `...`).

g1eny0ung: Sorry I missed this comment. I agree with you. this line can be removed. @anuragpaliwal80 Please help update it when you have time? Thx!

g1eny0ung (changes requested):

PR #3243, opened 2022-05-07
[BugBash] SC2034: Specify export for unused var

anuragpaliwal80:
Specify export for unused var

### Related changes

  - [ ] release-2.1
  - [ ] release-2.0

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


tools/schedule-migration/migrate.sh
```diff
@@ -17,7 +17,7 @@
 set -e
 
 NAMESPACES=$(kubectl get namespace | sed '1d' | awk '{print $1}')
-SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
+export SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
```
g1eny0ung, line 20: After my review, I found that this line can be removed because `SCRIPT_DIR` is not used.

hack/lib.sh
```diff
@@ -39,7 +39,7 @@ KUBEBUILDER_VERSION=2.3.2
 KUSTOMIZE_BIN=$OUTPUT_BIN/kustomize
 KUSTOMIZE_VERSION=3.10.0
 KUBETEST2_VERSION=v0.1.0
-KUBETSTS2_BIN=$OUTPUT_BIN/kubetest2
+export KUBETSTS2_BIN=$OUTPUT_BIN/kubetest2
```
g1eny0ung, line 42: This var doesn't need to be exported, as it will be loaded at:

https://github.com/chaos-mesh/chaos-mesh/blob/42fcabc19f5268cf0e0808d6ea7f2171fba2f28a/hack/e2e.sh#L28

g1eny0ung (changes requested):

Issue #3204, opened 2022-04-29
Bug: Network Partition with both target & external target will block from target to external target


Andrewmatilde:
## Bug Report

**What version of Chaos Mesh are you using?**
master

**What did you do?**
apply network chaos like :
```yaml
kind: NetworkChaos
apiVersion: chaos-mesh.org/v1alpha1
metadata:
  namespace: default
  name: shenan-net
spec:
  selector:
    namespaces:
      - default
    pods:
      default:
        - e2e-helper
  mode: all
  action: partition
  direction: to
  target:
    selector:
      namespaces:
        - default
      pods:
        default:
          - e2e-helper1
    mode: all
  externalTargets:
    - www.baidu.com
```

**What did you expect to see?**

curl www.baidu.com in e2e-help1 works.

**What did you see instead?**

curl www.baidu.com in e2e-help1 failed.

PR #3308, opened 2022-05-29
Fix Network Partition with both target & external target will block from target to external target 

Andrewmatilde:
Signed-off-by: andrewmatilde <davis6813585853062@outlook.com>

#3204

Add an externalCidrs param to SetDrop.
And block selected target pods get it.

### Related changes

  - [x] release-2.2
  - [ ] release-2.1

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code
- [x] Manual test

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


Andrewmatilde:
PTAL @cwen0 @STRRL @YangKeao

STRRL:
How about creating a testcase for that?

(maybe with e2e test?

Andrewmatilde:
> How about creating a testcase for that?
> 
> (maybe with e2e test?

I found it will make a large pull request for e2e-testing since the e2e test of external target have not been writen. So I plan to make a new pull request for it later.

Andrewmatilde:
And there already have a problem hardly to solve. That is the DNS of external target is not stable which may lead to continuing failure in e2e-test.

STRRL:
> I found it will make a large pull request for e2e-testing since the e2e test of external target have not been writen. So I plan to make a new pull request for it later.

please create an issue for that : )

> And there already have a problem hardly to solve. That is the DNS of external target is not stable which may lead to continuing failure in e2e-test.

maybe we could use Pod's IP as the external target when we write e2e testcase :P

Andrewmatilde:
> please create an issue for that : )
OK
>maybe we could use Pod's IP as the external target when we write e2e testcase :P
That's a good idea.

PR #3312, opened 2022-05-30
obsolete the `pid` and `startTime` fields in iochaos, httpchaos and stresschaos

Hexilee:

In https://github.com/chaos-mesh/chaos-mesh/pull/2918, we bring each background process an uid to avoid reusing pid. However, in view of the compatibility of API, we cannot remove the field [pid](https://github.com/chaos-mesh/chaos-mesh/blob/2eca5c17b1604a670d14b16ba89f8d500d4bf68c/api/v1alpha1/podiochaos_types.go#L232).

In this PR, we obsolete the `pid` and `startTime` fields in iochaos, httpchaos and stresschaos, and locate the background processes by the `uid` field.

### Related changes

  - [ ] release-2.2
  - [ ] release-2.1

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [x] E2E test
- [ ] No code

Side effects

### Release note 

```text
Please add a release note.

```

### DCO

```shell
git commit --amend --signoff
git push --force
```


api/v1alpha1/podhttpchaos_types.go
```diff
@@ -26,14 +26,20 @@ type PodHttpChaosSpec struct {
 
 // PodHttpChaosStatus defines the actual state of PodHttpChaos.
 type PodHttpChaosStatus struct {
+	//+kubebuilder:deprecatedversion:warning="Pid is deprecated, please use Uid instead"
 	// Pid represents a running tproxy process id.
 	// +optional
 	Pid int64 `json:"pid,omitempty"`
```
STRRL, line 32:
```suggestion
	//+kubebuilder:deprecatedversion:warning="Pid is deprecated, please use Uid instead"
	// Pid represents a running tproxy process id.
	// Deprecated: Pid is deprecated, please use Uid instead
	// +optional
	Pid int64 `json:"pid,omitempty"`
```

we should also add a godoc style "Deprecated: " comment.

STRRL: and `+kubebuilder:deprecatedversion:warning` marker is supported by `controller-gen > 0.6.0`;  we should upgrade it in dev-env;

see https://github.com/kubernetes-sigs/controller-tools/pull/541 and https://github.com/kubernetes-sigs/controller-tools/releases/tag/v0.6.0

controllers/podiochaos/controller.go
```diff
@@ -153,18 +151,17 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		Volume:      obj.Spec.VolumeMountPath,
 		ContainerId: containerID,
 
-		Instance:  obj.Status.Pid,
-		StartTime: obj.Status.StartTime,
-		EnterNS:   true,
+		InstanceUid: obj.Status.Uid,
+		StartTime:   obj.Status.StartTime,
+		EnterNS:     true,
```
STRRL, line 156: I think we could also mark these old fields as deprecated in `.proto`.

Or just use "reserved" to remove it directly. 

refer: 
- https://developers.google.com/protocol-buffers/docs/proto3#options, see `deprecated` in section "Options"
- https://developers.google.com/protocol-buffers/docs/proto3#reserved

STRRL (commented):
rest LGTM
PTAL @Hexilee

api/v1alpha1/podhttpchaos_types.go
Hexilee: The godoc-style comments are appended. It seems there are breaking changes in controller-gen v0.6.0, I'd like to upgrade it in another PR.

controllers/podiochaos/controller.go
Hexilee: I prefer to use  "reserved".

Hexilee: Reserved

Hexilee:
PTAL again @STRRL

PR #3396, opened 2022-06-27
[multicluster] chaos injection and status synchronization

FingerLeader:
This PR implements chaos injection and status synchronization in remote clusters.

According to the [rfc/44](https://github.com/chaos-mesh/rfcs/pull/44), this PR deploys chaos with `cloudregistry`, and before deploying, uses `predicates` to filter crds containing `remotecluster`.


YangKeao:
https://github.com/chaos-mesh/chaos-mesh/pull/3414 Includes mutliple modification for this PR, and this PR, practically, needs a lot of modifications to work, so it's relatively meaningless to review it. Should we close it? @FingerLeader

mdnyeemakhtar:
Hello,

We are running chaos mesh version 2.5.1. Below is our use case and remote cluster management is a perfect solution for us however we facing issues while running remote cluster chaos.

We have multi-cluster setup where Cluster A manages Cluster B. And we have automation framework which can apply chaos mesh experiments to cluster B via cluster A using Remote Cluster management since it does not have direct access to cluster B. However since chaos status is not synched back to Cluster A experiment CR our automation can't see the actual status of experiment that ran on cluster B.

Since this PR was closed and PR #3453 which is mentioned in the comment above has `Synchronize the status` on TODO list, is this feature in the works in a different PR? @YangKeao @FingerLeader.

PR #3405, opened 2022-06-28
stresschaos: run pause before choom

YangKeao:

To modify the `oom_score_adj`, we used the `choom` program. Then the whole process would be: `choom pause nsexec memStress`. However, if the `choom` runs not that fast, when the chaos-daemon is sending SIGCONT to the subprocess, it may find the comm is `choom` and thought this process has resumed.

```go
comm, err := util.ReadCommName(cmd.Process.Pid)
if err != nil {
	return nil, err
}
if comm != "pause\n" {
	log.Info("pause has been resumed", "comm", comm)
	break
}
```

So, the `pause` should always be the first command to execute.

Reorder the command executed by bpm.


YangKeao:
> Shall we add some e2e test to cover this bug?

StressChaos itself has e2e test, but this bug will only occur iff the chaos-daemon runs faster than choom, so it cannot be caught 100% (actually, we doesn't come with this problem (in e2e test or user's response) after it has been merged several monthes, which is suprising :cry:  )

Hexilee (approved):
> > Shall we add some e2e test to cover this bug?
> 
> StressChaos itself has e2e test, but this bug will only occur iff the chaos-daemon runs faster than choom, so it cannot be caught 100% (actually, we doesn't come with this problem (in e2e test or user's response) after it has been merged several monthes, which is suprising 😢 )

Fine, LGTM!

cwen0 (approved):
LGTM

PR #3445, opened 2022-07-12
fix: refactor service level in pkg/dashboard/apiserver/experiment

ConnieHan2019:
### What problem does this PR [solve?](https://github.com/chaos-mesh/chaos-mesh/issues/2875)
#2875 

### Related changes

  - [ ] release-2.2
  - [ ] release-2.1

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### DCO

```shell
git commit --amend --signoff
git push --force
```


pkg/dashboard/apiserver/experiment/experiment.go
```diff
@@ -60,12 +60,14 @@ func NewService(
...
 	config *config.ChaosDashboardConfig,
 	scheme *runtime.Scheme,
+	log logr.Logger,
 ) *Service {
 	return &Service{
 		archive: archive,
 		event:   event,
 		config:  config,
 		scheme:  scheme,
+		log: log,
```
xlgao-zju, line 70: use `u.Log.WithName("experiments")` would be better

xlgao-zju (changes requested):
and we can remove the Line 48, since we will not use  the var `log` anymore.

xlgao-zju:
you can run `make check` before commit.

pkg/dashboard/apiserver/experiment/experiment.go
g1eny0ung: This is a legacy method to log something, so this part is not a problem.

xlgao-zju (approved):
LGTM

PR #3484, opened 2022-07-26
Use `net.Interfaces()` to implement `getAllInterfaces()`

FingerLeader:

Close [#187](https://github.com/chaos-mesh/chaosd/issues/187)

Mainly for Chaosd's use.
Use `net.Interfaces()` ranther than `ip addr -j` to implement `getAllInterfaces()`, which can aovid errors caused by difference behaviors of `ip` command-line tool in different OS.

### Related changes

  - [ ] release-2.2
  - [ ] release-2.1

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### DCO

```shell
git commit --amend --signoff
git push --force
```


pkg/chaosdaemon/tc_server.go
```diff
@@ -73,34 +73,18 @@ func generateQdiscArgs(action string, qdisc *pb.Qdisc) ([]string, error) {
 	return args, nil
 }
 
-func getAllInterfaces(ctx context.Context, log logr.Logger, pid uint32, enterNS bool) ([]string, error) {
-	processBuilder := bpm.DefaultProcessBuilder("ip", "-j", "addr", "show").SetContext(ctx)
-	if enterNS {
-		processBuilder = processBuilder.SetNS(pid, bpm.NetNS)
```
YangKeao, line 79: It's getting interfaces in the remote network namespace. Using `net.Interfaces()` directly will break Chaos Mesh.

YangKeao: We could use `net.Interfaces()` as a fallback, if `enterNS` is false.

FingerLeader: I tested `getAllInterfaces()`, it actually get the remote NICs.

FingerLeader: FYI image

YangKeao: @FingerLeader I don't understand. You don't even tell him the target pid, how could it get interfaces in the network namespace of target process?

BTW, a normal single-stack container should only have `["lo", "eth0"]` :thinking: .

FingerLeader: I think that's because when deamon server calls the `SetTcs()`, the namespace has already been set to the target namespace.

FingerLeader: image
You can see when "handling tc request", `SetTcs`'s namespacedName is "default/ubuntu".

YangKeao: > the namespace has already been set to the target namespace.

I'm pretty sure it doesn't work like this. The `chaos-daemon` never changes itself's net namespace. The interfaces of `chaos-daemon` and target net namespaces could be nearly the same (as it doesn't use host network), you cannot tell it's the target net namespace only by the interfaces.

YangKeao: `chaos-daemon` spawns `nsexec` and `nsexec` will switch the net namespace and execute following commands (e.g. `tc`), so it never changes the net namespace of `chaos-daemon`.

FingerLeader: I checked ip of `eth0`, `net.Interfaces()` does get the wrong data 😢

FingerLeader:
PTAL /cc @YangKeao @Andrewmatilde

Andrewmatilde (approved):
LGTM

pkg/chaosdaemon/tc_server.go
```diff
@@ -74,33 +75,37 @@ func generateQdiscArgs(action string, qdisc *pb.Qdisc) ([]string, error) {
...
+		log.Info("get interfaces from ip command", "ifaces", ifaces)
+	} else {
+		interfaces, err := net.Interfaces()
+		if err != nil {
+			return []string{}, errors.New("fail to read ifname from net.Interfaces()")
+		}
+		for _, iface := range interfaces {
+			ifaces = append(ifaces, iface.Name)
+		}
+		log.Info("get interfaces from net.Interfaces()", "ifaces", ifaces)
```
YangKeao, line 106: Move this log out of the `if` condition.

YangKeao (commented):
Rest LGTM

pkg/chaosdaemon/tc_server.go
FingerLeader: There are two logs in branches, and they are different in sources ([L97](https://github.com/FingerLeader/chaos-mesh/blob/0212a5f7879da2a0baa9eeccc24cb33df0a868d5/pkg/chaosdaemon/tc_server.go#L97))

YangKeao: Sorry. I understand :facepalm:

YangKeao (approved):
LGTM

PR #3473, opened 2022-07-22
Protect chaos available namespace and filter namespaces if needed

brnck:
Signed-off-by: brnck <a.berneckas@gmail.com>

### What problem does this PR solve?

Close #3452 

### What's changed and how it works?

This PR fixes endpoint `/api/common/chaos-available-namespaces` when **SecurityMode** and ** enableFilterNamespace** - both set to true.

### Related changes

- [ ] Need to update `chaos-mesh/website`
- [ ] Need to update `Dashboard UI`
- Need to **cheery-pick to release branches**
  - [ ] release-2.2
  - [ ] release-2.1

### Checklist

CHANGELOG

- [x] I have updated the `CHANGELOG.md`
- [ ] I have labeled this PR with "no-need-update-changelog"

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code
- [X] Manual test (add steps below)

Steps to test:
* Deploy chaos-mesh
* `curl <domain.tld>:<port>/api/common/chaos-available-namespaces` - will return all namespaces
* Set `enableFilterNamespace` to true and redeploy.
* `curl <domain.tld>:<port>/api/common/chaos-available-namespaces` - will return `[]`
* `kubectl annotate ns chaos-mesh chaos-mesh.org/inject=enabled`
* `curl <domain.tld>:<port>/api/common/chaos-available-namespaces` - will return `["chaos-mesh"]`
* Set `SecurityMode` to true and redeploy.
* `curl <domain.tld>:<port>/api/common/chaos-available-namespaces` - will return empty body
* Create token using **token generator** feature in dashboard UI
* `curl -H "Authorization: Bearer <generated_token>" <domain.tld>:<port>/api/common/chaos-available-namespaces` - will return `["chaos-mesh"]`

Side effects

- [ ] Breaking backward compatibility

### DCO

If you find the DCO check fails, please run commands like below (Depends on the actual situations. For example, if the failed commit isn't the most recent) to fix it:

```shell
git commit --amend --signoff
git push --force
```

Issue #3452, opened 2022-07-14
Return only annotated namespaces when filter is enabled


brnck:
## Feature Request

**Is your feature request related to a problem? Please describe:**

I’ve enabled filter namespace in the controller:

```
# If enabled, only pods in the namespace annotated with `"chaos-mesh.org/inject": "enabled"` could be injected
enableFilterNamespace: true
```

When logged in to the UI I can see all available namespaces in the cluster. For the record - only one namespace has this annotation (demo-bookapp)

![image](https://user-images.githubusercontent.com/14329469/178921195-da2f993e-ce4e-42e9-962d-fcc40d89998c.png)

But as you can see full list of namespaces is visible in the UI

**Describe the feature you'd like:**

Ideally, I'd like Chaosmesh to return only those namespace that has the annotation `"chaos-mesh.org/inject": "enabled"`.

Even though we can't run chaos tests against the namespaces that do not have this annotation, reducing information noise would improve the UX. It would, also, simplify things because when a cluster has >100 namespaces, but only, let's say, 5 namespaces have this annotation, there is no point in returning rest of the namespaces

Return only namespaces that have the annotation will remove information noise in a couple of places:

<img width="1294" alt="image" src="https://user-images.githubusercontent.com/14329469/178922510-77f88d8d-116f-4e8e-90af-3a7ff527f4c8.png">

**Describe alternatives you've considered:**

N/A

**Teachability, Documentation, Adoption, Migration Strategy:**

N/A

nioshield:
@STRRL the feature have any plan ?

brnck:
@nioshield I'm planning to contribute and resolve this by myself. If everything is okay, I'll start working on this tomorrow or Thursday

PR #3595, opened 2022-09-02
Remove token checker when listing namespaces

cwen0:
Signed-off-by: Cwen Yin <cwenyin0@gmail.com>

`chaos-available-namespaces` API will be used before entering Token. If we add this checker will make generating token action fail.  Caused by https://github.com/chaos-mesh/chaos-mesh/pull/3473

Remove token checker when listing namespaces. 

### Related changes

  - [x] release-2.3
  - [x] release-2.2

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
 - Fix "sidecar config" e2e test cases run failed in some scenario.[#3564](https://github.com/chaos-mesh/chaos-mesh/pull/3564)
 - Fix recover bug when setting force recover to true [#3578](https://github.com/chaos-mesh/chaos-mesh/pull/3578)
 - Fix Integration test with bumping kubectl version. [#3589](https://github.com/chaos-mesh/chaos-mesh/pull/3589)
+- Fix generate token failed on chaos dashboard [#3595](https://github.com/chaos-mesh/chaos-mesh/pull/3595)
 
 ### Security
 
--- a/pkg/dashboard/apiserver/common/common.go
+++ b/pkg/dashboard/apiserver/common/common.go
@@ -227,7 +227,7 @@ func (s *Service) getChaosAvailableNamespaces(c *gin.Context) {
 		}
 		namespaces = make(sort.StringSlice, 0, len(nsList.Items))
 		for _, ns := range nsList.Items {
-			if s.conf.EnableFilterNamespace && !namespace.CheckNamespace(context.TODO(), kubeCli, ns.Name, u.Log) {
+			if s.conf.EnableFilterNamespace && !namespace.CheckNamespace(context.TODO(), s.kubeCli, ns.Name, u.Log) {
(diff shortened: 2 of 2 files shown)


STRRL:
> If we add this checker will make generating token action fail.

What does this mean?

g1eny0ung (approved):
LGTM

cwen0:
> > If we add this checker will make generating token action fail.
> 
> What does this mean?

image
image

g1eny0ung:
> > If we add this checker will make generating token action fail.
> 
> What does this mean?

There is some missing context here.

This is because we need to get the available namespaces before we can get the token. The error will happen in this place (https://chaos-mesh.org/docs/manage-user-permissions/#create-user-accounts-and-bind-permissions). Users need to get the available ns in advance when generating the RBAC templates.

cwen0:
> > If we add this checker will make generating token action fail.
> 
> What does this mean?

This checker will make generating RBAC template fail.

STRRL:
> > > If we add this checker will make generating token action fail.
> > 
> > 
> > What does this mean?
> 
> There is some missing context here.
> 
> This is because we need to get the available namespaces before we can get the token. The error will happen in this place (https://chaos-mesh.org/docs/manage-user-permissions/#create-user-accounts-and-bind-permissions). Users need to get the available ns in advance when generating the RBAC templates.

Is there any other solutions to archive #3452 and also avoid breaking things? @g1eny0ung @cwen0 🤔🤔

cwen0:
> > > > If we add this checker will make generating token action fail.
> > > 
> > > 
> > > What does this mean?
> > 
> > 
> > There is some missing context here.
> > This is because we need to get the available namespaces before we can get the token. The error will happen in this place (https://chaos-mesh.org/docs/manage-user-permissions/#create-user-accounts-and-bind-permissions). Users need to get the available ns in advance when generating the RBAC templates.
> 
> Are there any other solutions to archive #3452 and also avoid breaking things? @g1eny0ung @cwen0 🤔🤔

We can add an other API for generating RBAC templates.

brnck:
> > > > > If we add this checker will make generating token action fail.
> > > > 
> > > > 
> > > > What does this mean?
> > > 
> > > 
> > > There is some missing context here.
> > > This is because we need to get the available namespaces before we can get the token. The error will happen in this place (https://chaos-mesh.org/docs/manage-user-permissions/#create-user-accounts-and-bind-permissions). Users need to get the available ns in advance when generating the RBAC templates.
> > 
> > 
> > Are there any other solutions to archive #3452 and also avoid breaking things? @g1eny0ung @cwen0 🤔🤔
> 
> We can add an other API for generating RBAC templates.

It would be great not to show all namespaces to an unauthorized user as well as not break the current feature to generate tokens.

Showing all namespaces for unauthorized users poses some security risks I am afraid

PR #3642, opened 2022-09-17
Update changelog for prev release

STRRL:

append missing changelog entry

- copy/paste changelogs for patch version from each branch
- if one entry has been released in previous patch version, remove it from `Unreleased` section and `2.<y+1>.0` section

### Related changes

  - [ ] release-2.3
  - [ ] release-2.2

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [x] No code

Side effects

### DCO

```shell
git commit --amend --signoff
git push --force
```


CHANGELOG.md
```diff
@@ -212,10 +297,78 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
...
-- Fix `real_gettimeofday` on arm64 [#2849](https://github.com/chaos-mesh/chaos-mesh/pull/2849)
-- Fix Github Action `upload-image` [#2935](https://github.com/chaos-mesh/chaos-mesh/pull/2935)
-- Fix JVMChaos to handle the situation that the container which holds the JVM rules has been deleted [#2981](https://github.com/chaos-mesh/chaos-mesh/pull/2981)
-- Fix typo in comments for Chaos API [#3109](https://github.com/chaos-mesh/chaos-mesh/pull/3109)
+
+### Security
+
+- Nothing
+
+## [v2.1.8] - 2020-08-30
```
g1eny0ung, line 305: Should be `2022-08-30`.

CHANGELOG.md
```diff
@@ -212,10 +297,78 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
 - Fix chaos-kernel build, mark bcc version [#2693](https://github.com/chaos-mesh/chaos-mesh/pull/2693)
 - Fix wrong field name of PhysicalMachineChaos on Chaos Dashboard [#2724](https://github.com/chaos-mesh/chaos-mesh/pull/2724)
 - Fix field descriptions of GCPChaos [#2791](https://github.com/chaos-mesh/chaos-mesh/pull/2791)
-- Fix `real_gettimeofday` on arm64 [#2849](https://github.com/chaos-mesh/chaos-mesh/pull/2849)
-- Fix Github Action `upload-image` [#2935](https://github.com/chaos-mesh/chaos-mesh/pull/2935)
-- Fix JVMChaos to handle the situation that the container which holds the JVM rules has been deleted [#2981](https://github.com/chaos-mesh/chaos-mesh/pull/2981)
-- Fix typo in comments for Chaos API [#3109](https://github.com/chaos-mesh/chaos-mesh/pull/3109)
```
g1eny0ung, line 218: Why delete these?

CHANGELOG.md
STRRL: Updated!

CHANGELOG.md
STRRL: This bug-fixes already carried with `2.1.5`, which is earlier than the `2.2.0`, so they should be deleted from `Unreleased` section before we release 2.2.0. So I remove them.

g1eny0ung: @STRRL We should keep these original changes (if they have cherry-picks into old versions).

Also, if we change these, they won't be consistent with the release notes.

STRRL: Yes. You're right, keep tracking the cherrypicked PR and making it keep consistent would cost so much.....

I would append them back.

BTW, maybe we should not manually CHANGELOG.md anymore, instead of using PR's title...

g1eny0ung: > BTW, maybe we should not manually CHANGELOG.md anymore, instead of using PR's title...

🤔 You mean some lines haven't been consistent with the original PR titles? I think it's fine as long as you can describe what PR does because this is somewhat difficult to constrain everyone to do so.

CHANGELOG.md
```diff
@@ -63,20 +63,44 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
...
-- fix NetworkChaos fail with @ifXX in the device [#3605](https://github.com/chaos-mesh/chaos-mesh/pull/3605)
 - Fix IOChaos `containerNames` field in UI [#3533](https://github.com/chaos-mesh/chaos-mesh/pull/3533)
 - Fix BlockChaos can't show Chinese name. [#3536](https://github.com/chaos-mesh/chaos-mesh/pull/3536)
-- Add `omitempty` JSON tag to optional fields of the CRD objects. [#3531](https://github.com/chaos-mesh/chaos-mesh/pull/3531)
-- Fix "sidecar config" e2e test cases run failed in some scenario.[#3564](https://github.com/chaos-mesh/chaos-mesh/pull/3564)
 - Fix recover bug when setting force recover to true [#3578](https://github.com/chaos-mesh/chaos-mesh/pull/3578)
-- Fix Integration test with bumping kubectl version. [#3589](https://github.com/chaos-mesh/chaos-mesh/pull/3589)
-- Fix generate token failed on chaos dashboard [#3595](https://github.com/chaos-mesh/chaos-mesh/pull/3595)
-- Remove `limit` action of `BlockChaos` in the dashboard [#3655](https://github.com/chaos-mesh/chaos-mesh/pull/3655)
-- Sync PhysicalMachineChaos to API client and forms [#3660](https://github.com/chaos-mesh/chaos-mesh/pull/3660)
```
g1eny0ung, line 79: These also need to add back.

STRRL: No.. They are not carried with 2.3.1, ref: https://github.com/chaos-mesh/chaos-mesh/releases/tag/v2.3.1

CHANGELOG.md
STRRL: Yes. It could be hard. GitHub could actually generate the CHANGELOG based on PRs, and here are some practices called "Semantic PRs". 👀

CHANGELOG.md
g1eny0ung: Got it.

CHANGELOG.md
g1eny0ung: But auto-generation PRs from GitHub will be generated only before we prepare to release a new version. So it still needs to constrain the format in the CHANGELOG because it will be updated earlier.

g1eny0ung (approved):
LGTM

PR #3647, opened 2022-09-19
ADD HTTP TLS Support 

Andrewmatilde:

Current HTTPChaos cannot inject TLS connections, in this PR we can make it.

Users should provide `cert`, `key` and optional `ca` by a k8s `Secret` resource.

### Related changes

- [x] This change also requires further upadtes to `chaos-mesh/website` (e.g. docs)
- [x] This change also requires further updates to `UI interface`
  - [ ] release-2.3
  - [ ] release-2.2

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### DCO

```shell
git commit --amend --signoff
git push --force
```


Hexilee (approved):
LGTM, please fix the tests

api/v1alpha1/podhttpchaos_types.go
```diff
@@ -181,6 +186,26 @@ const (
...
+	// CertName represents the data name of cert file in secret, `tls.crt` for example
+	CertName string `json:"certName,omitempty"`
+
+	// KeyName represents the data name of key file in secret, `tls.key` for example
+	KeyName string `json:"KeyName,omitempty"`
+
+	// CAName represents the data name of ca file in secret, `ca.crt` for example
+	// +optional
+	CAName *string `json:"caName,omitempty"`
+}
```
STRRL, line 207: - if these fields are required, please remove the `omitempty` tag and append `+kubebuilder:validation:Required`
- if they are optional, please add `+optional` comment

e2e-test/e2e/chaos/httpchaos/http_tls.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	err = f.Close()
+	framework.ExpectNoError(err, "close server.ext file error")
+
+	cmdStr := "openssl x509 -req -in server.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out server.crt -days 3650 -sha256 -extfile server.ext"
+	cmd := exec.Command("bash", "-c", cmdStr)
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		framework.ExpectNoError(err, "run openssl cmd error: "+string(output))
+	}
```
STRRL, line 86: it seems that server.crt already existed in directory `keys` 🤔

api/v1alpha1/podhttpchaos_types.go
```diff
@@ -181,6 +186,26 @@ const (
...
+
+	// SecretNamespace represents the namespace of required secret resource,
+	// should be the same with deployment/chaos-controller-manager in most cases
+	SecretNamespace string `json:"secretNamespace,omitempty"`
+
+	// CertName represents the data name of cert file in secret, `tls.crt` for example
+	CertName string `json:"certName,omitempty"`
+
+	// KeyName represents the data name of key file in secret, `tls.key` for example
+	KeyName string `json:"KeyName,omitempty"`
```
STRRL, line 202: `KeyName` -> `keyName`

api/v1alpha1/podhttpchaos_types.go
```diff
@@ -181,6 +186,26 @@ const (
 	PodHttpResponse PodHttpChaosTarget = "Response"
 )
 
+// PodHttpChaosTLS contains the tls config for HTTPChaos
+type PodHttpChaosTLS struct {
+	// SecretName represents the name of required secret resource
+	SecretName string `json:"secretName,omitempty"`
+
+	// SecretNamespace represents the namespace of required secret resource,
+	// should be the same with deployment/chaos-controller-manager in most cases
```
STRRL, line 195: > should be the same with deployment/chaos-controller-manager in most cases

please remove this comment 🤣 there is no reason to make that suggestion/prediction

pkg/chaosdaemon/pb/chaosdaemon.proto
```diff
@@ -233,6 +233,8 @@ message ApplyHttpChaosRequest {
   int64 startTime = 5;
   bool enterNS = 6;
   string instance_uid = 7;
+
+  string tls = 8;
```
STRRL, line 237: IMHO, using JSON string inside of a protocol is definitely anti-pattern (like multi-value in one column in RDBMS)

it means we are using convention instead of schema/protocol, that's ambiguous.

helm/chaos-mesh/templates/controller-manager-rbac.yaml
```diff
@@ -110,9 +110,12 @@ rules:
...
     verbs: [ "*" ]
-  - apiGroups: [""]
-    resources: ["configmaps"]
-    verbs: ["*"]
+  - apiGroups: [ "" ]
+    resources: [ "configmaps" ]
+    verbs: [ "*" ]
+  - apiGroups: [ "" ]
+    resources: [ "secrets" ]
+    verbs: [ "get" ]
```
STRRL, line 118: these rule about secrets should be added into Role `{{ .Release.Name }}-chaos-controller-manager-target-namespace`, not here.

see L26-L64

STRRL:
AWESOME WORK!!! 

Rest LGTM!

several nits about the codebase, PTAL! ❤️  @Andrewmatilde

cwen0:
@Andrewmatilde PTAL

e2e-test/e2e/chaos/httpchaos/http_tls.go
Andrewmatilde: I will keep them , and then someone can just create a serect directly with it.

pkg/chaosdaemon/pb/chaosdaemon.proto
Andrewmatilde: I think this maybe the anti-pattern of protocol like json. But not the anti-pattern of RPC, especially we use it just like running a span of code in two machine. The main problem is gRPC cannot do tls parsing and unparsing for us.

helm/chaos-mesh/templates/controller-manager-rbac.yaml
Andrewmatilde: Fixed.

api/v1alpha1/podhttpchaos_types.go
Andrewmatilde: Fixed.

api/v1alpha1/podhttpchaos_types.go
Andrewmatilde: Fixed.

pkg/chaosdaemon/pb/chaosdaemon.proto
STRRL: > The main problem is gRPC cannot do tls parsing and unparsing for us.

It seems that the content of `tls` is `[]PodHttpChaosBaseRule`, which is complex but it does not mean there is no way to describe it in protobuf. 

But I think it's too heavy to make this PR carry these changes, so I would not block this PR even this does not look good to me.

Similar issues also happen on stress chaos and jvm chaos, we should refine the protobuf file before it becomes too hard to maintance.

STRRL:
@Andrewmatilde 

please execute `make check` on your local machine :)

Andrewmatilde:
> please execute `make check` on your local machine :)

Fixed.

STRRL:
I noticed that there was already a related documentation update: https://github.com/chaos-mesh/website/pull/362

Do we need a new doc or a post/blog describing how to use this feature? 

cc @cwen0 @Andrewmatilde

Andrewmatilde:
@STRRL Could you give some advice?

STRRL:
> @STRRL Could you give some advice?

How about a blog, an example of using HTTPChaos for services exposed as https ?

STRRL (approved):

PR #3655, opened 2022-09-20
dashboard: remove limit action from BlockChaos and generate specs

YangKeao:

This PR removes the limit action, and run codegen, though I don't know whether these codegen behaves well.

### Related changes

  - [x] release-2.4
  - [x] release-2.3
  - [ ] release-2.2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
 - Fix recover bug when setting force recover to true [#3578](https://github.com/chaos-mesh/chaos-mesh/pull/3578)
 - Fix Integration test with bumping kubectl version. [#3589](https://github.com/chaos-mesh/chaos-mesh/pull/3589)
 - Fix generate token failed on chaos dashboard [#3595](https://github.com/chaos-mesh/chaos-mesh/pull/3595)
+- Remove `limit` action of `BlockChaos` in the dashboard [#3655](https://github.com/chaos-mesh/chaos-mesh/pull/3655)
 
 ### Security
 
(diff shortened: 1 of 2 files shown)


g1eny0ung (changes requested):
It looks like for a long time the `PhysicalMachineChaos` weren't syncing the OpenAPI after the changes, so it's better to fix that with the new PR since I'm not sure if it's causing problems (not sure if it is compatible with `v2.3`) either. 😢

I need to test it before I can guarantee that everything works, so I suggest that this PR does not take any changes other than `ui/app/src/components/NewExperimentNext/data/types.ts` for now.

g1eny0ung:
I will update the CI to check if the OpenAPI changes need to be synced to the API clients and forms in the UI.

YangKeao:
> It looks like for a long time the PhysicalMachineChaos weren't syncing the OpenAPI after the changes, so it's better to fix that with the new PR since I'm not sure if it's causing problems (not sure if it is compatible with v2.3) either. cry
> I need to test it before I can guarantee that everything works, so I suggest that this PR does not take any changes other than ui/app/src/components/NewExperimentNext/data/types.ts for now.

Good! I reset these files. I wonder whether I need to generate files for `BlockChaos`? (I'll need to add a `"blockchaos"` in a variable, and then it can be generated).

g1eny0ung:
> > It looks like for a long time the PhysicalMachineChaos weren't syncing the OpenAPI after the changes, so it's better to fix that with the new PR since I'm not sure if it's causing problems (not sure if it is compatible with v2.3) either. cry
> > I need to test it before I can guarantee that everything works, so I suggest that this PR does not take any changes other than ui/app/src/components/NewExperimentNext/data/types.ts for now.
> 
> Good! I reset these files. I wonder whether I need to generate files for `BlockChaos`? (I'll need to add a `"blockchaos"` in a variable, and then it can be generated).

@YangKeao After I check PhysicalMachineChaos it will be ready, I will do this change this afternoon and I will let you know when it is done.

g1eny0ung (approved):
LGTM

g1eny0ung:
> cherry pick to release-2.3 failed

Manually cherry-pick with #3658.

PR #3716, opened 2022-10-24
fix(chaos daemon): logging kvs as pair

STRRL:
fix panic: odd number of arguments passed as key-value pairs for logging

Signed-off-by: STRRL <im@strrl.dev>

this PR would fix the integration test failure, caused by chaos daemon jvm_sever panic when logging. 

### Related changes

  - [ ] release-2.4
  - [ ] release-2.3
  - [ ] release-2.2

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### DCO

```shell
git commit --amend --signoff
git push --force
```


STRRL:
The JVM Integration test failed because the chaosdaemon crashed/panicked when logging, so JVMChaos is failed to inject.

And JVMChaos  integration test is skipped on ARM64, that's the reason why it only breaks the x86 integration tests.

image

g1eny0ung (approved):
LGTM

PR #3721, opened 2022-10-24
fix: ignore unnecessary workflows when `**.md` or `ui/**` changes

g1eny0ung:

As the title.

### Related changes

  - [ ] release-2.4
  - [ ] release-2.3
  - [ ] release-2.2

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### DCO

```shell
git commit --amend --signoff
git push --force
```


STRRL:
I think we do not need to cherrypick into release-2.2 anymore.

STRRL (approved):
LGTM

STRRL (changes requested):
Oops, I just realized that the Check "E2E Test Passed" is **required** for merging the PR..

g1eny0ung:
> I think we do not need to cherrypick into release-2.2 anymore.

Makes sense. 😌

STRRL:
> Oops, I just realized that the Check "E2E Test Passed" is **required** for merging the PR..

Maybe we should do the same thing like `ci.yaml` and `ci_skip.yaml`.

g1eny0ung:
> Oops, I just realized that the Check "E2E Test Passed" is **required** for merging the PR..

@STRRL We can use the same mechanism as `ci_skip.yaml`. I will update it soon.

.github/workflows/integration_test_skip.yml
```diff
@@ -0,0 +1,21 @@
...
+
+permissions: read-all
+
+jobs:
+  run:
+    name: Integration Test
+    strategy:
+      matrix:
+        arch: [amd64, arm64]
+    runs-on: ${{ fromJson('{"amd64":"ubuntu-latest", "arm64":["self-hosted", "Linux", "ARM64"]}')[matrix.arch] }}
```
g1eny0ung, line 19: @STRRL If we just skip it, shall we just need to write `ubuntu-latest`?

g1eny0ung: My mistake. An integration test is not required for merging PRs. I removed this file.

g1eny0ung:
> > Oops, I just realized that the Check "E2E Test Passed" is **required** for merging the PR..
> 
> @STRRL We can use the same mechanism as `ci_skip.yaml`. I will update it soon.

@STRRL After I tried and checked the Actions docs, I found that either `paths` or `dorny/paths-filter@v2` doesn't meet our need because we want **only markdown files to be changed** to not run the `E2E Test` workflow. Both of them will run if at least one other path matches.  :(

So I discover a new way to detect the changes. Using https://github.com/tj-actions/changed-files/ (support `only_changed` detection) and checking whether only markdown files changed. if it is, the rest of the logic is similar to `ci_skip.yml`.

cwen0 (approved):
LGTM

cwen0:
@STRRL PTAL

Issue #3775, opened 2022-11-11
ci: go verify job has long-standing unresolved warnings


g1eny0ung:
For a long time, our code has not been improved for these warnings. After I checked, I found that we only enabled revive as a lint tool very early in the project, but didn't follow up in the following years.

I am going to improve this problem.

g1eny0ung:
Use `revive@v1.2.4` and merge its [Recommended Configuration](https://github.com/mgechev/revive#recommended-configuration) with our configuration. The result is:

<img width="315" alt="image" src="https://user-images.githubusercontent.com/15034155/201279068-909330e2-9c8e-41c3-9de0-18b6b8538ae9.png">

Many warnings will occur after uncommenting `var-naming`, `package-comments`, and `unused-parameter`. 😢

g1eny0ung:
I opened a PR to solve part of the problem. Several remaining issues are difficult to solve at once.

- [ ] Resolve `var-naming` warnings in CI/go/verify job
- [ ] Resolve `package-comments` warnings in CI/go/verify job
- [ ] Resolve `unused-parameter` warnings in CI/go/verify job
- [ ] Resolve `flag-parameter` warnings in CI/go/verify job

PR #3776, opened 2022-11-11
fix(ci): resolve part of verify warnings and upgrade revive

g1eny0ung:

Related to #3775.

See https://github.com/chaos-mesh/chaos-mesh/issues/3775#issue-1444996947.

### Related changes

  - [ ] release-2.4
  - [ ] release-2.3

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### DCO

```shell
git commit --amend --signoff
git push --force
```


tools/schedule-migration/main.go
```diff
@@ -23,7 +23,7 @@ import (
 	"gopkg.in/yaml.v2"
 )
 
-type Item struct {
+type Object struct {
```
STRRL, line 26: just a little curious, using `Item` as the strut name is a bad smell? or is there revive warnings on here? 

Or it just renames because easy to understand?

STRRL: rest LGTM

g1eny0ung: > Or it just renames because easy to understand?

Yep. I think it would be more appropriate to use `Object`. No warnings here.

The bad smell that occurred in this file is using `new` as the variable name. At this line:

https://github.com/chaos-mesh/chaos-mesh/blob/40080be801e7ea9230813b0ec971e6292c74c278/tools/schedule-migration/main.go#L45

The `new` is a reserved keyword.

STRRL: got that

PR #3888, opened 2023-01-09
introduce `make help` as default make recipe

### LINKED ISSUE #3769: Use code generation instead of `eval` in makefile
Also relates to https://github.com/chaos-mesh/chaos-mesh/issues/1963

Currently, we are using `eval` for rendering targets for "building executable binary" and "building container" images:

https://github.com/chaos-mesh/chaos-mesh/blob/cfbca5e5becfb09b0129a8d9702a930942999730/Makefile#L172-L175

https://github.com/chaos-mesh/chaos-mesh/blob/cfbca5e5becfb09b0129a8d9702a930942999730/Makefile#L211-L220

It's not easy to understand and maintain that makefile with dynamic rendering with `eval`. Also, it would reduce the performance of the bash completion for the makefile.

I suggest writing a go utility application, and generating `binary.generated.mk` and `container-image.generated.mk`, instead of using `eval` in `Makefile`

STRRL:

close #3769

In this PR, we provide a target called `make help` like this:

image

### Related changes

  - [ ] release-2.5
  - [ ] release-2.4

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### DCO

```shell
git commit --amend --signoff
git push --force
```


STRRL:
/cc @g1eny0ung 

PTAL

g1eny0ung:
I will look at it this afternoon and this evening.

hack/local-up-chaos-mesh.sh
```diff
@@ -107,7 +107,7 @@ fi
 
 if [ -z "$SKIP_IMAGE_BUILD" ]; then
     echo "info: building docker images"
-    IMAGE_REGISTRY_PREFIX=$IMAGE_REGISTRY_PREFIX IMAGE_PROJECT=chaos-mesh IMAGE_TAG=$IMAGE_TAG UI=1 SWAGGER=1 make image
+    IMAGE_REGISTRY_PREFIX=$IMAGE_REGISTRY_PREFIX IMAGE_TAG=$IMAGE_TAG UI=1 SWAGGER=1 make image
```
g1eny0ung, line 110:
```suggestion
    IMAGE_TAG=$IMAGE_TAG UI=1 SWAGGER=1 make image
```

If I understand correctly, now we are no longer providing the ability to customize the registry, and some targets with `docker-push` prefix also are being removed.

I agree with these changes, since it's easy to rename the image name with `docker image tag`, we don't need to keep these seemingly convenient, but somewhat redundant targets.

So the `IMAGE_REGISTRY_PREFIX` env should be removed here as well as others.

cmd/generate-makefile/generate_container_image_makefile.go
```diff
@@ -0,0 +1,156 @@
...
+clean-image-built:
+{{- range .Recipes }}
+	rm -f {{ .SourcePath }}/.dockerbuilt
+{{- end }}
+
+`
+
+// containerImageGeneratedMkTemplate is the template for one target, use containerImageRecipeOptions as the context.
+const containerImageRecipeTemplate = `.PHONY: image-{{ .ImageName }}
+image-{{ .ImageName }}:{{ .SourcePath }}/.dockerbuilt ## {{ .Comment }}
```
g1eny0ung, line 46:
```suggestion
image-{{ .ImageName }}: {{ .SourcePath }}/.dockerbuilt ## {{ .Comment }}
```

g1eny0ung (commented):
The rest LGTM. As it's too late, I'll check it again this morning.

cmd/generate-makefile/funcmaps.go
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+package main
+
+import "strings"
+
+var defaultFuncMap = map[string]interface{}{"StringsJoin": StringsJoin}
+
+func StringsJoin(s []string, sep string) string {
+	return strings.Join(s, sep)
```
g1eny0ung, line 22: It is better to check for empty strings.

Makefile
```diff
@@ -49,22 +45,172 @@ ifeq ($(UI),1)
...
+# The awk commands is responsible to read the entire set of makefiles included in this invocation, looking for lines of the file as xyz: ## something, and then pretty-format the target and help. Then, if there's a line with ##@ something, that gets pretty-printed as a category.
+# More info over the usage of ANSI control characters for terminal formatting: https://en.wikipedia.org/wiki/ANSI_escape_code#SGR_parameters
+# More info over awk command: http://linuxcommand.org/lc3_adv_awk.php
+#
+# Notice that we have a little modification on the awk command to support slash in the recipe name:
+# origin: /^[a-zA-Z_0-9-]+:.*?##/
+# modified /^[a-zA-Z_0-9\/\.-]+:.*?##/
+.PHONY: help
+help: ## Display this help
+	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9\/\.-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
```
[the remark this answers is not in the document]
STRRL: Great! Any reason for using `21` instead of `15`, not even longer? Or should use a larger number here?

hack/local-up-chaos-mesh.sh
STRRL: > If I understand correctly, now we are no longer providing the ability to customize the registry, and some targets with `docker-push` prefix also are being removed.

Yeah, that's the thing I want to do: do NOT provide the flags/ENVs for configuring different registry/repo, but **only during the building process**.

It seems the changes in `local-up-chaos-mesh.sh` is NOT a part of building things but installing things. 

So maybe it's better to revert the changes in `local-up-chaos-mesh.sh`.

If necessary,  I would open another PR for similar changes, but for installing scripts like `install.sh`, `local-up-chaos-mesh.sh` and other .sh scripts.

What do you think about it? Reverting changes in `local-up-chaos-mesh.sh`

cmd/generate-makefile/generate_container_image_makefile.go
STRRL: Nice Catch!

cmd/generate-makefile/funcmaps.go
STRRL: `strings.Join` already takes care about that :P

https://github.com/golang/go/blob/e8fe3b77574d9afadab664104a669d8e9f0f4303/src/strings/strings.go#L432-L464

STRRL: I would mark this comment as resovled

hack/local-up-chaos-mesh.sh
g1eny0ung: > What do you think about it? Reverting changes in local-up-chaos-mesh.sh

I still don't think it's necessary because these scripts are used more for debugging purposes, and giving them more configurable items is often redundant. As I said above:

_I agree with these changes, since it's easy to rename the image name with `docker image tag`, we don't need to keep these seemingly convenient, but somewhat redundant targets._

g1eny0ung: @STRRL Only this place is not updated, after the update I will approve this PR.

cwen0:
Cool! I have run this pr on the environment, The `make help` is very useful!

Makefile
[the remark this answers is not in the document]
STRRL: got that.

hack/local-up-chaos-mesh.sh
STRRL: Hi @g1eny0ung , I actually prefer to revert changes in `local-up-chaos-mesh.sh` **in this PR**.

I totally agree with your opinion. But it's a little out of the scope from this PR.

I would do it, but in another new PR. Not carry changes in `.sh` files in the PR.

What do you think about it?

g1eny0ung: > But it's a little out of the scope from this PR.

@STRRL That sounds reasonable. I agree that we can wrap up the `.sh` modifications in the upcoming PR.

g1eny0ung (approved):
LGTM!

g1eny0ung:
@chaos-mesh/maintainers @chaos-mesh/committers Please be aware that any additions or removals of make targets in the `Makefile` must now be done with consideration for modifying the template.

PR #3897, opened 2023-01-13
Install offline Helm Chart for a multi-cluster

nioshield:
Install the specified version Helm Chart offline to multiple clusters

### Related changes

  - [ ] release-2.5
  - [ ] release-2.4

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### DCO

```shell
git commit --amend --signoff
git push --force
```


STRRL:
Hi @nioshield , thanks for your contribution! Awesome patch! 

I am going to review it soon.

helm/chaos-mesh/values.yaml
```diff
@@ -139,6 +139,13 @@ controllerManager:
     retryPeriod: 2s
   # chaosdSecurityMode is enabled for mTLS connection between chaos-controller-manager and chaosd
   chaosdSecurityMode: true
+  # multi cluster install offline helm chart path
+  localHelmChart:
+    enabled: true
+    volume:
+      hostPath:
+        path: /data/helm
+        type: DirectoryOrCreate
```
STRRL, line 148: The default value for `controllerManager.localHelmChart.enabled` is `true`, and the default value of `controllerManager.localHelmChart.volume` is a `HostPath` which most probably does NOT exist on the environment. It would lead to the installation failing with the default values.

What about:

- making `controllerManager.localHelmChart.enabled`  as `false`
- also make the `controllerManager.localHelmChart.volume` as a volume for ConfigMap or Secret, with some comments.

controllers/multicluster/remotecluster/controller.go
```diff
@@ -189,7 +189,7 @@ func (r *Reconciler) ensureHelmRelease(ctx context.Context, obj *v1alpha1.Remote
 	_, err = helmClient.GetRelease(obj.Spec.Namespace, chaosMeshReleaseName)
 	if err != nil {
 		if errors.Is(err, driver.ErrReleaseNotFound) {
-			chart, err := helm.FetchChaosMeshChart(ctx, chaosMeshReleaseVersion)
+			chart, err := helm.FetchChaosMeshChart(ctx, obj.Spec.Version, config.ControllerCfg.LocalHelmChartPath)
```
STRRL, line 192: The logic about fetching  `Status.CurrentVersion` needs updates.

Fields in `Status` means the **observed status** by the controller, it should NOT be setup by simply copying the value from `Spec.Version`.

What about making a refactor on `helm.FetchChaosMeshChart`, derive 2 method like:

- helm.FetchChaosMeshChartFromRepository(ctx context.Context, version string) (chart, version ,err)
- helm.FetchChaosMeshChartFromLocal(ctx context.Context, helmChartPath string) (chart, version ,err)

then setup the `Status.CurrentVersion` with the returned `version` value.

pkg/helm/chart.go
```diff
@@ -41,9 +41,21 @@ func FetchChaosMeshChart(ctx context.Context, version string) (*chart.Chart, err
...
+func GetChaosMeshChartTgzPath(ctx context.Context, version, local string) (string, error) {
+	fileName := fmt.Sprintf("/chaos-mesh-%s.tgz", version)
+	tgzPath := fmt.Sprintf("%s/%s", os.TempDir(), fileName)
+	if local != "" {
+		tgzPath = fmt.Sprintf("%s/%s", local, fileName)
+	}
+
+	if _, err := os.Stat(tgzPath); err != nil {
+		return DownloadChaosMeshChartTgz(ctx, version)
+	}
```
STRRL, line 53: please do not fallback to the `DownloadChaosMeshChartTgz`, just return the error.

the fallback behavior would make it complex and ambiguous.

STRRL (commented):
Basically LGTM!

there are several comments about the codes. PTAL

Thanks! @nioshield

nioshield:
@STRRL fixed according to comments. PTAL

pkg/helm/chart.go
```diff
@@ -29,21 +29,43 @@ import (
...
 	}
 	return requestedChart, nil
 }
 
+func GetChaosMeshChartTgzPath(ctx context.Context, version, local string) (string, error) {
+	fileName := fmt.Sprintf("/chaos-mesh-%s.tgz", version)
+	tgzPath := fmt.Sprintf("%s/%s", os.TempDir(), fileName)
+	if local != "" {
+		tgzPath = fmt.Sprintf("%s/%s", local, fileName)
+	}
```
STRRL, line 59:
```suggestion
	fileName := fmt.Sprintf("chaos-mesh-%s.tgz", version)
	tgzPath := fmt.Sprintf("%s/%s", os.TempDir(), fileName)
	if local != "" {
		tgzPath = fmt.Sprintf("%s/%s", local, fileName)
	}
```

the `/` already existed when concat the `tgzPath`, no need to append it in `fileName`

STRRL (changes requested):
rest LGTM! 

also, please fix the DCO check

STRRL (approved):

g1eny0ung:
Supplement docs in https://github.com/chaos-mesh/website/commit/ba2b18d1d77c8e4268956867b522eca936d0c008.

PR #3973, opened 2023-03-13
feat: pod-level control injection policy

tddhit:

We want to reduce the risk of injecting faults in production by having a fine-grained control injection policy.
ChaosMesh can support namespace-level isolation by enabling "enableFilterNamespace" or specifying "targetNamespace", but our k8s-based platform will force all deployments to use the same specified namespace, so the above policy cannot be used by us. We hope chaosmesh to support pod-level control injection policy

Injection can also be controlled on a per-pod basis, by configuring the chaos-mesh.org/inject label on a pod.

### Related changes

  - [ ] release-2.5
  - [ ] release-2.4

### Checklist

CHANGELOG

Tests

- [] Unit test
- [ ] E2E test
- [ ] No code

1. Set the enableFilterPod of helm/chaos-mesh/values.yaml to true/false
2.  Bring/Remove the annotation “chaos-mesh.org/inject” to the pod
3.  Apply pod-kill-example.yaml 
4. Observe whether the pod is killed

Side effects

### DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/helm/chaos-mesh/templates/chaos-dashboard-deployment.yaml
+++ b/helm/chaos-mesh/templates/chaos-dashboard-deployment.yaml
@@ -77,6 +77,8 @@ spec:
               value: {{ .Values.controllerManager.targetNamespace | quote }}
             - name: ENABLE_FILTER_NAMESPACE
               value: "{{ .Values.controllerManager.enableFilterNamespace }}"
+            - name: ENABLE_FILTER_POD
+              value: "{{ .Values.controllerManager.enableFilterPod }}"
             - name: SECURITY_MODE
               value: "{{ .Values.dashboard.securityMode }}"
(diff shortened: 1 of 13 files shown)


STRRL:
Hi @tddhit, thanks for your contribution!

> Injection can also be controlled on a per-pod basis, by configuring the chaos-mesh.org/inject label on a pod.

IMHO, this feature could be done with the "set-based selector" which could select pods with complex rules:

https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement

STRRL:
I have no idea there are other benefits from this solution.

Feel free to comment if I missed something. :heart:

STRRL:
I am going to close this PR.

Feel free to reopen it if you want to ongoing ideas.

PR #3986, opened 2023-03-29
fix: make `IN_DOCKER=1` effective in build

g1eny0ung:

If `IN_DOCKER` equals `1`, we shouldn't let `image-build-env` as a prerequisite.

### Related changes

  - [x] release-2.5
  - [x] release-2.4

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### DCO

```shell
git commit --amend --signoff
git push --force
```


Makefile
```diff
@@ -135,7 +136,11 @@ GO_TARGET_PHONY :=
 define COMPILE_GO_TEMPLATE
 
 $(1): SHELL:=$(RUN_IN_BUILD_SHELL)
+ifeq ($(IN_DOCKER),1)
```
cwen0, line 139: `IN_DOCKER=1` represent that build in docker. So we should check if IN_DOCKER equal to 0 here.

cwen0, line 139: `IN_DOCKER=1` represents the build action has been run in docker. So we shouldn't use it as a switch for whether to compile in docker

g1eny0ung: Yes. For historical reasons, it represents that build in docker, but for local environment, it still needs to pass `IN_DOCKER=1` to assume **"we are in the container"**, so that you don't have to start docker compilation.

g1eny0ung: In conclusion, after further discussion with @cwen0, we feel that this issue should be reconsidered:

1. Since our default build behavior is to build using containers to avoid compatibility issues, we should no longer provide an `IN_DOCKER` env entry to trigger a local build in `Makefile`.
2. For local build, we should create new targets to handle, like `make chaosctl`.

FYI @chaos-mesh/maintainers @chaos-mesh/committers, hope to get your suggestions. I will create two issues to track and close this PR for now.

g1eny0ung: 1. #3988
2. #3989

PR #3992, opened 2023-03-30
chore: remove `IN_DOCKER` env in `Makefile`

### LINKED ISSUE #3988: Remove `IN_DOCKER` env entry in `Makefile`
Ref: https://github.com/chaos-mesh/chaos-mesh/pull/3986#discussion_r1152731381

Since our default build behavior is to build using containers to avoid compatibility issues, we should no longer provide an `IN_DOCKER` env entry to trigger a local build in `Makefile`.

g1eny0ung:

Close #3988

Since our default build behavior is to build using containers to avoid compatibility issues, we should no longer provide an `IN_DOCKER` env entry to trigger a local build in `Makefile`.

### Related changes

  - [ ] release-2.5
  - [ ] release-2.4

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### DCO

```shell
git commit --amend --signoff
git push --force
```


g1eny0ung:
@cwen0 PTAL.

STRRL:
it seems `IN_DOCKER` still is included in `build/get_env_shell.py`.

we could also remove that.

g1eny0ung:
> it seems `IN_DOCKER` still is included in `build/get_env_shell.py`.
> 
> we could also remove that.

Since we will still be running `get_env_shell.py` in the container when building, if we want to completely remove the `IN_DOCKER` environments, we need to find another way to determine whether the current process is running in the container.

I found a helpful stackoverflow link, I will test it to see if it works properly: https://stackoverflow.com/questions/20010199/how-to-determine-if-a-process-runs-inside-lxc-docker.

g1eny0ung:
> > it seems `IN_DOCKER` still is included in `build/get_env_shell.py`.
> > we could also remove that.
> 
> Since we will still be running `get_env_shell.py` in the container when building, if we want to completely remove the `IN_DOCKER` environments, we need to find another way to determine whether the current process is running in the container.
> 
> I found a helpful stackoverflow link, I will test it to see if it works properly: https://stackoverflow.com/questions/20010199/how-to-determine-if-a-process-runs-inside-lxc-docker.

@STRRL Test in my local and works. Please check again.

cwen0 (approved):
LGTM

Issue #3989, opened 2023-03-30
Provide a way to build binaries locally in `Makefile`


g1eny0ung:
Ref: https://github.com/chaos-mesh/chaos-mesh/pull/3986#discussion_r1152731381

STRRL:
Great! 

Actually, I noticed another similar way to make contributor's local setup more easier: https://github.com/apache/incubator-opendal/pull/1687/files

But our situation is, we have used too many environment variables to configure the building process, which is not easy to passthrough into the docker.

STRRL:
Any other idea about avoiding using too many environment variables as "feature flags" in Makefile?

also cc @yangkeao @cwen0

g1eny0ung:
> Any other idea about avoiding using too many environment variables as "feature flags" in Makefile?
> 
> also cc @YangKeao @cwen0

@STRRL In #4004, I tried to alleviate this problem with a temporary solution. PTAL.

g1eny0ung:
This issue has a milestone and is already available. So I will close it for now.

PR #4004, opened 2023-04-09
feat!: build binaries locally with `local/` prefix targets in `Makefile`

g1eny0ung:

Ref #3989. **Please not that this PR does not fully address this issue.**

> First of all, **TL:DR**, this PR only introduces a temporary way to build **_some binaries_** locally (i.e. it doesn't trigger any docker commands).

As @STRRL says in https://github.com/chaos-mesh/chaos-mesh/issues/3989#issuecomment-1493233625, we have relied on many env variables in the Makefile. Most of them are used by two variables: `RUN_IN_DEV_SHELL` and `RUN_IN_BUILD_SHELL` (as their names, they actually use a Python script (`get_env_shell.py`) to generate the `docker run` command string to make sure the target process run inside `dev-env` or `build-env` containers).

When we run `make xxx` from the command line, these two variables will use a `BASIC_IMAGE_ENV` variable, and the `BASIC_IMAGE_ENV` will contain the versions of `dev-env` and `build-env` in the current branch, which we will compute from a shell script, `hack/env-image-tag.sh`:

```Makefile
ifeq ($(IMAGE_BUILD_ENV_TAG),)
export IMAGE_BUILD_ENV_TAG := $(shell ./hack/env-image-tag.sh build-env)
endif
ifeq ($(IMAGE_DEV_ENV_TAG),)
export IMAGE_DEV_ENV_TAG := $(shell ./hack/env-image-tag.sh dev-env)
endif
```

In `hack/env-image-tag.sh`, we will use the `mikefarah/yq` image to get the versions from `env-images.yaml`. So so far, we have two calls to docker. But if we want to run some targets without needing docker, we still have to perform the steps described above.

The screenshot below is deliberately run without docker to show the process of `make clean`:

image

As you can see, we'll still go and get the versions of `dev-env` and `build-env`, but we won't use them.

---

So in this PR, I tried to avoid using these docker-related variables for the purpose of building locally through:

```Makefile
ifeq (,$(findstring local/,$(MAKECMDGOALS)))
# Do not call the if block when the target include string of "local/"
endif
```

After the changes, when we run `make local/xxx', it tells the Makefile that we want to build some binaries locally, not in containers. This if block will also wrap the calls to [canned recipes](https://www.gnu.org/software/make/manual/html_node/Canned-Recipes.html#Canned-Recipes), as all our defines are aimed at building images.

In summary, although it is now possible to avoid calling Docker-related commands in the make process for local build purposes, this requires us to manually prefix the command with local/, so I call this a temporary solution.

So instead of changing `make clean` above to `make local/clean`, I added a few commands to build binaries **for local debugging purposes (this is also the main purpose of the local build, for production builds we still need to use container builds to ensure compatibility)**, such as:

```Makefile
# Build chaos-controller-manager locally
local/chaos-controller-manager:
	$(GO) build -ldflags '$(LDFLAGS)' -o bin/chaos-controller-manager ./cmd/chaos-controller-manager/main.go

# Build chaos-dashboard locally
local/chaos-dashboard:
	$(CGO) build -ldflags '$(LDFLAGS)' -o bin/chaos-dashboard ./cmd/chaos-dashboard/main.go

# Build chaosctl locally
local/chaosctl:
	$(GO) build -ldflags '$(LDFLAGS)' -o bin/chaosctl ./cmd/chaosctl/main.go
```

And others remain the same (e.g. no local/chaos daemon) because they rely heavily on a different instruction set architecture.

### Related changes

  - [ ] release-2.5
  - [ ] release-2.4

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

- [x] **Breaking backward compatibility**

### DCO

```shell
git commit --amend --signoff
git push --force
```


Makefile
```diff
@@ -352,12 +368,12 @@ swagger_spec: SHELL:=$(RUN_IN_DEV_SHELL)
 swagger_spec: images/dev-env/.dockerbuilt
 	swag init -g cmd/chaos-dashboard/main.go --output pkg/dashboard/swaggerdocs --pd --parseInternal
 
-.PHONY: all clean test install manifests groupimports fmt vet tidy image \
-	docker-push lint generate config \
-	install.sh $(GO_TARGET_PHONY) \
-	manager chaosfs chaosdaemon chaos-dashboard \
-	dashboard dashboard-server-frontend gosec-scan \
-	failpoint-enable failpoint-disable swagger_spec \
+.PHONY: all image clean test install manifests manifests/crd.yaml \
```
g1eny0ung, line 371: I also arranged and removed some phony targets, below are targets that have been removed because they are outdated:

- manager
- chaosfs
- chaosdaemon
- chaos-dashboard
- dashboard
- dashboard-server-frontend
- go_build_cache_directory

g1eny0ung:
This PR will need to be rebased after #3888 is merged.

Makefile
```diff
@@ -212,11 +222,26 @@ test: generate manifests test-utils images/dev-env/.dockerbuilt ## Run unit test
 
 ##@ Advanced building targets
 
-test-utils: timer multithread_tracee pkg/time/fakeclock/fake_clock_gettime.o pkg/time/fakeclock/fake_gettimeofday.o
+# Build chaosctl locally
+local/chaosctl:
+	$(GO) build -ldflags '$(LDFLAGS)' -o bin/chaosctl ./cmd/chaosctl/main.go
```
STRRL, line 227: it seems that the original recipe `chaosctl` was already been build locally, so there is no difference between `make local/chaosctl` and the old `make chaosctl`.

Or is there other things I missed?

Makefile
```diff
@@ -212,11 +222,26 @@ test: generate manifests test-utils images/dev-env/.dockerbuilt ## Run unit test
...
 
-timer:
+local/timer:
 	$(GO) build -ldflags '$(LDFLAGS)' -o bin/test/timer ./test/cmd/timer/*.go
```
STRRL, line 230: Here, also like `chaosctl`, the original recipe `timer` was been locally built.

STRRL (commented):
rest LGTM

Makefile
g1eny0ung: Yes, it is. Only the name has changed.

> After discussing with @STRRL, we decided not to change the old locally built targets.

g1eny0ung:
@STRRL Need to review again. 👀

STRRL (approved):

PR #4008, opened 2023-04-11
chore: refine dashboard config

g1eny0ung:

This PR moves the `TTLConfig` to `pkg/config` and creates a new struct `TTLConfigWithStringTime` to consume the config from env. This change aims to isolate the TTL config definition from the implementation.

### Related changes

  - [x] release-2.5
  - [ ] release-2.4

### Checklist

CHANGELOG

Tests

- [x] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/cmd/chaos-dashboard/main.go
+++ b/cmd/chaos-dashboard/main.go
@@ -29,7 +29,7 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	ctrl "sigs.k8s.io/controller-runtime"
 
-	config "github.com/chaos-mesh/chaos-mesh/pkg/config/dashboard"
+	config "github.com/chaos-mesh/chaos-mesh/pkg/config"
 	"github.com/chaos-mesh/chaos-mesh/pkg/dashboard/apiserver"
 	"github.com/chaos-mesh/chaos-mesh/pkg/dashboard/collector"
 	"github.com/chaos-mesh/chaos-mesh/pkg/dashboard/store"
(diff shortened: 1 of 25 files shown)


pkg/config/dashboard_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+	if config.WorkflowTTL != "336h" {
+		t.Error("WorkflowTTL is not set")
+	}
+
+	parsed, err := config.Parse()
+
+	if err != nil {
+		t.Error("Error parsing config")
+	}
```
STRRL, line 87:
```suggestion
	if err != nil {
		t.Fatal("Error parsing config", err)
	}
```

It's better to use `Fatal`, because there is no way to make the rest assertions if the parsing fails.

And it's better to print the origin error.

STRRL (commented):
rest LGTM

cmd/chaos-dashboard/main.go
```diff
@@ -39,7 +39,7 @@ import (
 )
 
 // @title Chaos Mesh Dashboard API
-// @version 2.2
+// @version 2.5
```
STRRL, line 42: A little curious about where we configure for this `version`?

STRRL:
Also, it requires `make check` to update generated codes.

cmd/chaos-dashboard/main.go
g1eny0ung: Currently no place. I used to modify it manually.

pkg/config/dashboard_test.go
g1eny0ung: I will update it soon.

g1eny0ung:
> Also, it requires `make check` to update generated codes.

Yes, I have updated the swagger docs.

STRRL (approved):
LGTM

PR #4056, opened 2023-05-21
fix: use `*time.Time` to represent finish time

### LINKED ISSUE #1659: dashboard: can't insert experiment in MySQL NO_ZERO_IN_DATE mode

g1eny0ung:

Close #1659

Use `*time.Time` to represent the finish time, this can prevent an invalid date `0000-00-00` being inserted when creating a chaos object.

### Related changes

  - [ ] release-2.5
  - [ ] release-2.4

### Checklist

CHANGELOG

Tests

- [ ] Unit test
- [ ] E2E test
- [ ] No code

Side effects

### DCO

```shell
git commit --amend --signoff
git push --force
```


STRRL:
it might bring breaking changes on db table schema, although we enabled the `AutoMigrate` of gorm.

I would take a try with MySQL later, then I would give my lgtm

g1eny0ung:
> it might bring breaking changes on db table schema, although we enabled the `AutoMigrate` of gorm.
> 
> I would take a try with MySQL later, then I would give my lgtm

Oh...yes, I tested in a new MySQL database and all works fine, but I didn't test on an existing MySQL with old Chaos experiments.

STRRL:
/lgtm

I noticed that gorm already used `null` for that column, no matter whether we use the pointer or not in the go struct.

```sql
create table experiments
(
    id          int unsigned auto_increment
        primary key,
    created_at  datetime      null,
    updated_at  datetime      null,
    deleted_at  datetime      null,
    uid         varchar(255)  null,
    kind        varchar(255)  null,
    name        varchar(255)  null,
    namespace   varchar(255)  null,
    action      varchar(255)  null,
    start_time  datetime      null,
    finish_time datetime      null,
    archived    tinyint(1)    null,
    experiment  varchar(4096) null
);
```

PR #4106, opened 2023-06-29
Allow annotations on chaos-controller-manager and chaos-daemon ServiceAccount

### LINKED ISSUE #4092: AWS Chaos Authorisation using AIM Roles
## Feature Request

When using AWS Chaos, we would prefer not to pass credentials such as `aws_access_key_id` and `aws_secret_access_key`

We would like to use use IAM Roles to authorise chaos-mesh to talk to the AWS API.

AWS provides an integration for this, so long as your service is running in EKS: [IAM Roles for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html)

By creating a service account with the correct permissions, the AWS SDK will pick this up automatically and handle the authorisation without needing secrets or credentials.

All that is needed is a small patch to the helm chart:

```
    - kustomize:
        patchesJson6902:
          - target:
              version: v1
              kind: ServiceAccount
              name: chaos-controller-manager
            patch:
              - op: add
                path: /metadata/annotations
                value:
                  'eks.amazonaws.com/role-arn': "arn:aws:iam::<blah-blah>"
```

This can be better supported by allowing role-arn to be passed in the chart values, and added to the ServiceAccount annotations if present.

philipptanlak-form3:

Currently there is no way to add AWS IAM roles to the service accounts of chaos-daemon and chaos-controller-manager.
This change adds the ability to add annotations, and therefore IAM roles, to the service accounts.

Close #4092

Two new configuration options have been introduced in the helm chart to provide annotations to the service accounts of chaos-daemon and chaos-controller-manager.

As a reference, here is the official AWS guide describing how a service account can assume an IAM role.
- https://docs.aws.amazon.com/eks/latest/userguide/associate-service-account-role.html

## Related changes

- [ ] release-2.6
- [ ] release-2.5

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [x] Manual test

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
 
 ### Added
 
-- Nothing
+- Allow annotations on chaos-controller-manager and chaos-daemon ServiceAccount [#4106](https://github.com/chaos-mesh/chaos-mesh/pull/4106)
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


cwen0 (approved):
LGTM

PR #4133, opened 2023-07-28
Prevent scroll from changing experiment number fields

raymyers:

Currently when entering numeric fields like `Port` into new Experiments, scrolling can change the number in the input box, while the intent is typically to move down the page.

Screenshot 2023-07-28 at 10 44 02 AM

This change adds an onWheel handler to all `InputField` components with type `number`. The handler simply calls `.blur()` on the input element to prevent the value changing.

For instance, this will apply to:
- Port in numerous experiment types
- Workers in Stress Experiment 
- Probability and "max times of failures" in Kernel Experiment
- historyLimit and startingDeadlineSeconds in Schedule
- Node number in New Workflow

## Related changes

- [ ] release-2.6
- [ ] release-2.5

### CHANGELOG

- [X] I have updated the `CHANGELOG.md`

### Tests

- [X] Unit test
- [ ] E2E test
- [X] Manual test

(Manually tested in Chrome 114.0.5735.198)

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
 
 ### Changed
 
+- Prevent mousewheel scroll in the `New Experiment` page from changing numeric input values.
+
 - Change `awschaos/*/impl.go` to use `aws_session_token` with static credentials provider so that users can perform awschaos using [temporary AWS credentials](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html) [#4066](https://github.com/chaos-mesh/chaos-mesh/pull/4066)
 - Enable consistency for ghcr.io registry [#4091](https://github.com/chaos-mesh/chaos-mesh/pull/4091) [#4134](https://github.com/chaos-mesh/chaos-mesh/pull/4134)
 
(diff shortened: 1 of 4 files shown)


g1eny0ung:
I will review this PR within the next two days.

g1eny0ung:
> If desired I could moved this change into ui/app/src/components/FormField/TextField.tsx and it should then cover all of those, but I started with the smaller scope.

@raymyers I agree with that. Yes, if we want to prevent accidental scrolling from happening, then the best thing to do is update the base component: if the type of the input box is a number, then we add the `onWheel` event handler.

raymyers:
> > If desired I could moved this change into ui/app/src/components/FormField/TextField.tsx and it should then cover all of those, but I started with the smaller scope.
> 
> @raymyers I agree with that. Yes, if we want to prevent accidental scrolling from happening, then the best thing to do is update the base component: if the type of the input box is a number, then we add the `onWheel` event handler.

Understood. I've updated this to apply to all TextField components of type number. Please take a look when you get a chance. Thanks!

g1eny0ung (approved):
/lgtm

g1eny0ung:
I submitted a commit to remove the commented code and organize the changelog a bit. Thanks for the contribution!

PR #4199, opened 2023-09-16
refactor: optimize `allInjected` and `allRecovered` states

g1eny0ung:

Optimize `allInjected` and `allRecovered` states when targets are not selected.

Before:

```txt
2023-09-16T02:51:42.517Z	INFO	condition	condition/controller.go:122	updating conditions
{"conditions": [{"type":"Paused","status":"False"},{"type":"Selected","status":"False"},{"type":"AllInjected","status":"True"},{"type":"AllRecovered","status":"True"}]}
```

After:

```txt
2023-09-16T03:57:50.937Z	INFO	condition	condition/controller.go:132	updating conditions
{"conditions": [{"type":"Paused","status":"False"},{"type":"Selected","status":"False"},{"type":"AllInjected","status":"False"},{"type":"AllRecovered","status":"False"}]}
```

Instead of arbitrarily setting the initial values of `allInjected` and `allRecovered`, the value is assigned based on the current selected state.

## Related changes

- [x] release-2.6
- [ ] release-2.5

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```


controllers/common/condition/controller.go
```diff
@@ -139,5 +102,68 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			Status: corev1.ConditionTrue,
+		}
+	} else {
+		newConditionMap[v1alpha1.ConditionSelected] = StatusAndReason{
+			Status: corev1.ConditionFalse,
+		}
+	}
+
+	// If records is `nil`, we don't need to check the `allInjected` and `allRecovered` conditions.
+	var allInjected corev1.ConditionStatus
```
cwen0, line 124:
```
allInjected := corev1.ConditionFalse
if records != nil && every(records, func(record *v1alpha1.Record) bool {
	return record.Phase == v1alpha1.Injected
}) {
	allInjected = corev1.ConditionTrue
}
```
I prefer this style instead of `if xxx else xxx`

controllers/common/condition/controller.go
```diff
@@ -139,5 +102,68 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+	} else {
+		newConditionMap[v1alpha1.ConditionPaused] = StatusAndReason{
+			Status: corev1.ConditionFalse,
+		}
+	}
+
+	return
+}
+
+func every[T any](arr []T, condition func(T) bool) bool {
```
cwen0, line 162: Can you rename this function? Let us know roughly what this function does from the name

controllers/common/condition/controller.go
cwen0: Go style refer this doc: https://github.com/uber-go/guide/blob/master/style.md#unnecessary-else

g1eny0ung: Fixed in [8681aea](https://github.com/chaos-mesh/chaos-mesh/pull/4199/commits/8681aea547c6161d6e72a60184e87ebfe4680499).

controllers/common/condition/controller.go
g1eny0ung: This function is generic and usually tests whether all elements in the slice pass the condition function, and it needs to be put into context to know what it is used for:

```go
every(records, func(record *v1alpha1.Record) bool {
        return record.Phase == v1alpha1.Injected
})
```

I don't think there's a better name for it, so I'll go ahead and add a comment to explain what it's for.

g1eny0ung: [7c91f92](https://github.com/chaos-mesh/chaos-mesh/pull/4199/commits/7c91f928ba5f84de84140e5a8ad4b9bad5c18ddc)

cwen0 (approved):
LGTM

PR #4154, opened 2023-08-16
Update k8s dependencies to 1.28.0

fasmat:
## What problem does this PR solve?

Updates k8s dependencies to 1.28.1

## What's changed and how it works?

Updated dependencies and fixed issues that popped up when doing so.

## Related changes

- [ ] This change also requires further updates to the [website](https://github.com/chaos-mesh/website) (e.g. docs)
- [ ] This change also requires further updates to the `UI interface`

## Cherry-pick to release branches (optional)

> This PR should be cherry-picked to the following release branches:

- [ ] release-2.6
- [ ] release-2.5

### CHANGELOG

> Must include at least one of them.

- [x] I have updated the `CHANGELOG.md`
- [ ] I have labeled this PR with "no-need-update-changelog"

### Tests

> Must include at least one of them.

- [x] Unit test
- [x] E2E test
- [ ] Manual test

### Side effects

- [ ] **Breaking backward compatibility**

## DCO

If you find the DCO check fails, please run commands like below (Depends on the actual situations. For example, if the failed commit isn't the most recent) to fix it:

```shell
git commit --amend --signoff
git push --force
```

PR #4200, opened 2023-09-17
refactor: deprecate local up script

g1eny0ung:

According to https://github.com/chaos-mesh/chaos-mesh/pull/4154#issuecomment-1705875424, the script 'local-up-chaos-mesh.sh' is rarely used in the current repository. Personally, I have never used it for development, as I prefer to use Helm to start development clusters. Introducing this script would only add to the complexity compared to Helm.

Delete `local-up-chaos-mesh.sh` and update the CI of integration tests.

## Related changes

- [x] release-2.6
- [ ] release-2.5

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```


test/integration_test/README.md
```diff
@@ -1,36 +1,36 @@
...
-You can install Chaos Mesh by commands:
+   You can install Chaos Mesh by commands:
 
-```bash
-hack/local-up-chaos-mesh.sh
-kubectl set env deployment/chaos-dashboard SECURITY_MODE=true -n chaos-mesh
-kubectl port-forward -n chaos-mesh svc/chaos-dashboard 2333:2333 &
-```
+   ```bash
+   hack/local-up-chaos-mesh.sh
```
STRRL, line 10: I think this line of doc could be replaced by

```bash
make image
minikube start
minikube image load ghcr.io/chaos-mesh/chaos-daemon:latest
minikube image load ghcr.io/chaos-mesh/chaos-mesh:latest
minikube image load ghcr.io/chaos-mesh/chaos-dashboard:latest
helm upgrade --install chaos-mesh ./helm/chaos-mesh/ -n chaos-mesh --create-namespace --set dashboard.securityMode=false
```

I am not sure that's a drop in replacement 🤔 maybe it requires some tunning work.

g1eny0ung: I still use the kind for bootstrapping the cluster as before in CI. Yes, this document needs some "tunning work" to fit the current development environment.

g1eny0ung: Prepare to replace with https://chaos-mesh.org/docs/configure-development-environment/.

g1eny0ung:
Now I found the integration test will fail with memStress. When testing `memStress`, it's child process won't be killed after recovering by chaosd. I will attempt to address the issue with `memStress` first.

PR #4229, opened 2023-10-15
fix: goroutine leak

0xff-dev:
Based on my local tests, it has been discovered that in this particular scenario, the goroutine may fail to exit when an unbuffered channel is used.

```go
package main

import (
	"fmt"
	"runtime"
	"time"
)

func try() {
	ch := make(chan error)
	go func() {
		time.Sleep(5 * time.Second)
		fmt.Println("write error")
		ch <- fmt.Errorf("leak")
	}()
	select {
	case <-time.After(3 * time.Second):
		break
	case <-ch:
		fmt.Println("get chan error")
	}
   fmt.Println("wait a moment")
	time.Sleep(3 * time.Second)
}
func main() {
	before := runtime.NumGoroutine()
	try()
	after := runtime.NumGoroutine()
	fmt.Println(after == before)
}
```

Using buffered channel.

## Related changes

- [x] release-2.6
- [x] release-2.5

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [x] Manual test

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```


STRRL:
LGTM. But only for cap 1 buffered channel is one of the error-prone practices.

/lgtm

cwen0 (approved):
/lgtm Good catch!

cwen0 (approved):

g1eny0ung:
@cwen0 Please refer to https://prow.chaos-mesh.dev/command-help?repo=chaos-mesh%2Fchaos-mesh to get the new commands about merging PRs and cherry-picks.

PR #4247, opened 2023-11-03
fix: disable `ctrl` server by default

grahambrereton-form3:
Disable GQL server for chaosctl by default, and only add `pods/exec` permissions if GQL server is specifically enabled.

The longterm solution could be to make the RPC calls required for the GQL API to function without using `pods/exec` calls.

This [RFC](https://github.com/chaos-mesh/rfcs/blob/main/text/2021-07-23-extensible-chaosctl.md) seems to highlight this solution and the risks associated with the current implementation.

I think there would be a lot of value in only allowing the controller manager to perform a restricted set of actions as part of the GQL API, at which point it should be reasonable to make this endpoint available by default again.

This reduces the risk associated with a default install. With the current state of things, the `chaos-controller-manager` can execute anything on any pod in the installation namespace, including the daemon, which could lead to privilege escalation.

## Related changes

- [ ] release-2.6
- [ ] release-2.5

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [x] Manual test

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```


STRRL:
LGTM! Thx!

STRRL:
Hi @grahambrereton-form3 , could you update this PR with the base branch?

you could merge the master branch into your branch or rebase your commits.

Then we could merge this PR.

grahambrereton-form3:
@STRRL just updated against `master`

STRRL:
sorry for this PR opened for a looong time with more and more conflicts, also as we have #4702 done, so I am going to close this PR;

thanks @grahambrereton-form3 sooooo much for the contribution ❤️

PR #4249, opened 2023-11-07
Allow logging configuration

grahambrereton-form3:

Allow configuration of key logging parameters through environment, and expose these environment variables through the helm chart values.

Allow configuration for truncating log fields over a certain limit. This prevents log aggregation failure in some cases where very large objects are logged.

Preserves similar behaviour to present if no configuration is provided.

New configuration environment variables added to setup zap logger in a configurable way.

These environment variables are then exposed in the helm chart through a new top-level key `logging` so that the logging config can be provided in one place and applied to all deployed applications.

## Related changes

- [ ] release-2.6
- [ ] release-2.5

### CHANGELOG

### Tests

- [x] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/pkg/log/zap.go
+++ b/pkg/log/zap.go
@@ -16,7 +16,12 @@
 package log
 
 import (
+	"encoding/json"
+	"fmt"
 	"io"
+	"os"
+	"strconv"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"github.com/go-logr/zapr"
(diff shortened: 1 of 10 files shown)


grahambrereton-form3:
I tried running the failing unit tests locally, and they passed. Not sure what caused the failure in CI.

pkg/log/zap.go
```diff
@@ -16,26 +16,140 @@
...
 // logger of your application, and provide it to your components, by fx or manually.
 func NewDefaultZapLogger() (logr.Logger, error) {
-	// change the configuration in the future if needed.
-	zapLogger, err := zap.NewDevelopment()
+	logLevel := zap.InfoLevel
+	if envLevel := os.Getenv(LogLevel); envLevel != "" {
+		var err error
+		logLevel, err = zapcore.ParseLevel(envLevel)
+		if err != nil {
+			return logr.Discard(), err
```
STRRL, line 63: when we fail to parse the loglevel, please use the default log level, and just use fmt.Println to print a message about this error and notice user we fallback to the default log level.

pkg/log/zap.go
```diff
@@ -16,26 +16,140 @@
...
+		}
+		if v := os.Getenv(LogKeyTimestamp); v != "" {
+			encoderConfig.TimeKey = v
+		}
+
+		if v := os.Getenv(LogTimestampFormat); v != "" {
+			var timeEncoder zapcore.TimeEncoder
+
+			if err := timeEncoder.UnmarshalText([]byte(v)); err != nil {
+				return logr.Discard(), err
```
STRRL, line 90: ditto, please fallback to a default timestamp format if we failed to parse that from ENV

STRRL (changes requested):

STRRL:
Hi @grahambrereton-form3, thanks for your contribution!

because the logger is a kind of application basic infrastructure so we'd better make it more robust.

grahambrereton-form3:
@STRRL I pushed a new commit, which should address your concerns.

pkg/log/zap.go
grahambrereton-form3: Interesting detail here is that `timeEncoder.UnmarshalText` in its current implementation never returns an error and silently defaults to `epoch`. I've added the error handler that just prints a warning, because it seems a bad idea to leave the error completely unhandled, but it won't be called with the current implementation in zap.

Could potentially drop the `else` clause that prints the error, given it's currently unreachable.

miketonks-form3:
Adding my review: LGTM.  

We've been using this in our local build for some time, and works well.  We export logs via fluentd to centralised log store, and the structured logging is a nice improvement.

grahambrereton-form3:
Rebased on master and reset author to me on the commit to make DCO happy :crossed_fingers:

patsevanton:
I would really like to manage the logging level. Can I merge this pull request?

STRRL:
> I would really like to manage the logging level. Can I marge this pull request?

Yes I think we could have that. let me rebase it and do some updates;

patsevanton:
Friendly reminder about this PR.

STRRL:
this branch does not allow me to push it, I am going to work in another branch

PR #4258, opened 2023-11-16
fix(test): docker command not found

g1eny0ung:

In https://github.com/chaos-mesh/chaos-mesh/actions/runs/6851229373/job/18627073546?pr=4246, I found that the `make test` will report `./hack/env-image-tag.sh: line 38: docker: command not found` errors.

image

After investigating the Makefile, the `test` target will recursively call `make failpoint-*` targets, which will call `docker` in the container, this is the reason for errors.

So I created a recipe to encapsulate the `failpoint-ctl` command, although we've done a refactoring before to avoid macros, I don't think this recipe is too complicated, so it's not a big problem.

```makefile
define failpoint-ctl
	find $(ROOT)/* -type d | grep -vE "(\.git|bin|\.cache|ui)" | xargs failpoint-ctl $1
endef
```

## Related changes

- [ ] release-2.6
- [ ] release-2.5

### CHANGELOG

### Tests

- [x] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ endif
 
 endif
 
-BASIC_IMAGE_ENV= IMAGE_DEV_ENV_TAG=$(IMAGE_DEV_ENV_TAG) \
+BASIC_IMAGE_ENV=IMAGE_DEV_ENV_TAG=$(IMAGE_DEV_ENV_TAG) \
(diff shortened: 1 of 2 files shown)


g1eny0ung:
Merged due to no response for a long time.

PR #4271, opened 2023-11-29
K8schaos update fix

wvdschel-f3:
Previous k8schaos update PR had a bug in how it stored the original resource value which caused it to behave incorrectly when multiple k8schaos experiments run in parallel.

This should address that problem.


STRRL:
> Appologies, this PR was meant for our internal chaos-mesh fork :)

That's OK.

If you have some cool ideas from your forked version want share with us. Just create an issue and let us know! Thanks!

PR #4286, opened 2023-12-12
fix: don't spawn new workflow child on abort

grahambrereton-form3:
These changes prevent the serial workflow node reconciler from creating a node for the next step if a status check step
fails and has `AbortWithStatusCheck` set.

This addresses a race condition where the serial reconciler spawns a new child node in the window between the statuscheck node aborting and the abort status being propagated to the parent. Without this change, the next step
in the workflow can start and then be immediately terminated after the abort status is propagated.

This allows use of a statuscheck workflow node to act as a gate to prevent further execution unless some conditions are met, e.g. not running a chaos experiment if the affected application is not in a healthy state.

If status check node aborted and has "abort with statuscheck" flag set, don't spawn a new workflow child node for the next step.

## Related changes

- [ ] release-2.6
- [ ] release-2.5

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```


STRRL:
Hi @grahambrereton-form3 , could you add some test cases in pkg/workflow/controllers/serial_node_reconciler_test.go for this situation? Thanks!

STRRL:
I realize that the tricky question is there will always be possibilities that serial reconciler spawn the new node, because different reconcilers run concurrently, there is no synchronization behavior between statuscheck and serial node(also other types of node).

So there is no guarantee that serial reonciler would see the latest status of other nodes / aborted parent workflow marked as aborted.

As we have used multiple reconcilers instead of one huge reconciler, I just have no idea how to make sure the "abort" behavior would be affected "immediately".

STRRL:
I prefer to back to the "eventual consistency", although workflownode is created, but it could stop its business as soon as it noticed the `abort` marked.

I found that the `AbortNodeReconciler#propagateAbortToChildren` would have possibility to left some new created workflownode because of the same concurrency issue: if a new workflownode created after AbortNodeReconciler call the fetchChildNodes. it would never be marked as aborted. 

I think change the propagation model would help to resolve this issue: a workflownode would be marked as aborted if its owner workflownode marked as aborted.

How do you think about it? cc @grahambrereton-form3 @iguoyr

grahambrereton-form3:
@STRRL 

I don't *think* there's a concurrency issue here, but I may well be missing something. As I understand it, the serial reconciler only spawns a new node when there a no active nodes, and the place that completes the status check workflow node is the same place that marks it as aborted.

https://github.com/chaos-mesh/chaos-mesh/blob/f55576898805fbf7b2d342eab250cbec0802c0fa/pkg/workflow/controllers/statuscheck_reconciler.go#L161-L188

The check I added just ensures we don't spawn the next child node in a situation where we know we will be aborting the workflow anyway. It avoids the messy situation where we spawn a node we know is going to be aborted as soon as the status propagates up to the parent node and back down to its children.

The concern I have with the "eventually consistent" approach is that I want to be able to use a status check node to test for system health before creating any chaos, and if the next child is spawned even temporarily, that could lead to a pod kill (for example) on an already unhealthy system.

miketonks-form3:
We've been using this fix in our branch for some time now, and it seems to work really well.

For serial workflows, if the pre-run status check fails, then the workflow aborts and the next serial step is never executed.

I don't think this change will upset the existing eventually consistent behaviour, which makes complete sense for parallel workflows.

Really keen to see this one merged

PR #4301, opened 2023-12-27
ci: use current branch Chaos Mesh version in e2e tests

g1eny0ung:

@STRRL Should we run e2e tests based on the current branch version of Chaos Mesh? For this I've updated the e2e workflow, and I think it's a reasonable change.

Choose a Chaos Mesh version before installing:

```yaml
- name: Choose Chaos Mesh version
  id: chaos-mesh-version
  run: echo "version=$(yq '.images.tag' helm/chaos-mesh/values.yaml)" >> "$GITHUB_OUTPUT"
```

## Related changes

- [ ] release-2.6
- [ ] release-2.5

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/.github/workflows/e2e_test.yml
+++ b/.github/workflows/e2e_test.yml
@@ -163,19 +163,18 @@ jobs:
           - "[Basic] [DNSChaos]"
           - "[Basic] [StressChaos]"
     steps:
-      - name: checkout codes
-        uses: actions/checkout@v4
-      - name: download saved images
+      - uses: actions/checkout@v4
+      - name: Download saved images
         uses: actions/download-artifact@v2
         with:
           name: saved-images
           path: ./output/saved-images
-      - name: download e2e binary
+      - name: Download e2e binary
         uses: actions/download-artifact@v2
         with:
           name: e2e-binary
           path: ./output/e2e-binary
-      - name: move e2e binary
+      - name: Move e2e binary
         run: |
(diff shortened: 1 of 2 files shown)


STRRL:
well, during the e2e test, we would build the container image **locally** with the changed codes, ref: https://github.com/chaos-mesh/chaos-mesh/blob/f55576898805fbf7b2d342eab250cbec0802c0fa/.github/workflows/e2e_test.yml#L97

and the artifacts of `make image` would always be like `ghcr.io/chaos-mesh/chaos-mesh:latest`, the image tag `latest` is kind of hard-coded. so we could just use the `latest` image, instead of using tagged version.

g1eny0ung:
> well, during the e2e test, we would build the container image **locally** with the changed codes, ref:
> 
> https://github.com/chaos-mesh/chaos-mesh/blob/f55576898805fbf7b2d342eab250cbec0802c0fa/.github/workflows/e2e_test.yml#L97
> 
> and the artifacts of `make image` would always be like `ghcr.io/chaos-mesh/chaos-mesh:latest`, the image tag `latest` is kind of hard-coded. so we could just use the `latest` image, instead of using tagged version.

Got it! Thanks for the explanation and I will close this PR.

PR #4172, opened 2023-08-30
feat: loadBalancerSourceRanges in chaos-dashboard service

jamshidi799:
## What problem does this PR solve?
This PR adds support of loadBalancerSourceRanges to chaos-dashboard service in the helm chart

 Close #3920 

## Related changes

- [ ] This change also requires further updates to the [website](https://github.com/chaos-mesh/website) (e.g. docs)
- [ ] This change also requires further updates to the `UI interface`

## Cherry-pick to release branches (optional)

> This PR should be cherry-picked to the following release branches:

- [ ] release-2.6
- [ ] release-2.5

### CHANGELOG

> Must include at least one of them.

- [x] I have updated the `CHANGELOG.md`
- [ ] I have labeled this PR with "no-need-update-changelog"

### Tests

> Must include at least one of them.

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

- [ ] **Breaking backward compatibility**

## DCO

If you find the DCO check fails, please run commands like below (Depends on the actual situations. For example, if the failed commit isn't the most recent) to fix it:

```shell
git commit --amend --signoff
git push --force
```

PR #4331, opened 2024-01-28
fix: cross build on darwin

g1eny0ung:

If I want to make a cross-build on Darwin (macOS M1), I can't build an amd64 image because the `--platform` only passes an arch param (`amd64`) but does not specify os. Since the `build-env` image based on `alpine:3.17` does not exist as `darwin/amd64`, I think we should force mark the `linux` prefix before the `--platform` param?

Add a `linux` prefix in `--platform`, like `f"linux/{os.getenv('TARGET_PLATFORM')}"`.

## Related changes

- [ ] release-2.6
- [ ] release-2.5

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [x] Manual test

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/build/get_env_shell.py
+++ b/build/get_env_shell.py
@@ -81,16 +81,12 @@ def main():
     cmd += ["--user", f"{os.getuid()}:{os.getgid()}"]
 
     target_platform = utils.get_target_platform()
-    # if the environment variable is not set, don't pass `--platform` argument,
+    # If the environment variable is not set, don't pass `--platform` argument,
     # as it's not supported on some docker build environment.
-    if os.getenv("TARGET_PLATFORM") is not None and os.getenv(
-            "TARGET_PLATFORM") != "":
-        cmd += ["--platform", f"linux/{os.getenv('TARGET_PLATFORM')}"]
+    if target_platform.from_env:
+        cmd += ["--platform", f"linux/{target_platform.platform}"]
     else:
-        cmd += ["--env", f"TARGET_PLATFORM={target_platform}"]
-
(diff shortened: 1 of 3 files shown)


build/get_env_shell.py
```diff
@@ -81,16 +81,12 @@ def main():
...
-    if os.getenv("TARGET_PLATFORM") is not None and os.getenv(
-            "TARGET_PLATFORM") != "":
-        cmd += ["--platform", f"linux/{os.getenv('TARGET_PLATFORM')}"]
+    if target_platform.from_env:
+        cmd += ["--platform", f"linux/{target_platform.platform}"]
     else:
-        cmd += ["--env", f"TARGET_PLATFORM={target_platform}"]
-
-    if target_platform == "arm64":
-        cmd += ["--env", "ETCD_UNSUPPORTED_ARCH=arm64"]
```
g1eny0ung, line 93: @STRRL I found a strange env here, and I removed it because we have any strong associations with etcd.

g1eny0ung:
I merged this PR due to a lack of response for a long time. @STRRL @cwen0 If you guys have any questions, please reply directly here.

build/get_env_shell.py
g1eny0ung: I forgot we must set this env for testing on the arm64 machine. Re-added it in https://github.com/chaos-mesh/chaos-mesh/pull/4172/commits/d3ce18d2764c1f4855e91c1f9f3c55f6f1e304c1.

PR #4338, opened 2024-02-01
fix: TTL configuration from wrong environment variables at dashboard.go

YuSitong1999:

The environment variable TTL_EXPERIMENT is incorrectly used for ScheduleTTL and WorkflowTTL.
Since the default values are accidentally the same, the unit test cannot catch this error.

reference:
https://github.com/chaos-mesh/chaos-mesh/blob/5a08714d88380d38b6f9b968253cbbbf839c2f9d/helm/chaos-mesh/values.yaml#L366

Change the ScheduleTTL source environment variable to TTL_SCHEDULE and the WorkflowTTL source environment variable to TTL_WORKFLOW.

## Related changes

- [x] release-2.6
- [x] release-2.5

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [x] Manual test

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```


g1eny0ung (approved):
Oops! A stupid mistake...Thanks for the fix!

PR #4373, opened 2024-03-18
fix: context cancel leak in portforward

zhanluxianshen:
fix: context cancel leak in portforward.

image
image


zhanluxianshen:
The ci error almost the ```timeout err``` ,  this change is not affectd by that much.

STRRL:
Integration test failed by setting up the environment. I would retry them.

PR #4377, opened 2024-03-19
fix: comments in pkg/chaosdaemon/util.GetChildProcesses

zhanluxianshen:

Fix comments in `pkg/chaosdaemon/util.GetChildProcesses`.

## Related changes

- [ ] release-2.6
- [ ] release-2.5

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/pkg/chaosdaemon/util/util.go
+++ b/pkg/chaosdaemon/util/util.go
@@ -46,11 +46,11 @@ func ReadCommName(pid int) (string, error) {
 }
 
 // GetChildProcesses will return all child processes's pid. Include all generations.
-// only return error when /proc/pid/tasks cannot be read
+// Only return error when `/proc` cannot be read.
 func GetChildProcesses(ppid uint32, logger logr.Logger) ([]uint32, error) {
 	procs, err := os.ReadDir(bpm.DefaultProcPrefix)
 	if err != nil {
-		return nil, errors.Wrapf(err, "read /proc/pid/tasks , ppid : %d", ppid)
+		return nil, errors.Wrapf(err, "read /proc")
 	}
 
 	type processPair struct {


pkg/chaosdaemon/util/util.go
```diff
@@ -46,11 +46,11 @@ func ReadCommName(pid int) (string, error) {
 }
 
 // GetChildProcesses will return all child processes's pid. Include all generations.
-// only return error when /proc/pid/tasks cannot be read
+// only return error when /proc/pid/stat cannot be read
```
g1eny0ung, line 49: It's not accurate here either, it should be `/proc`.

```suggestion
// Only return error when `/proc` cannot be read.
```

pkg/chaosdaemon/util/util.go
```diff
@@ -46,11 +46,11 @@ func ReadCommName(pid int) (string, error) {
...
 func GetChildProcesses(ppid uint32, logger logr.Logger) ([]uint32, error) {
 	procs, err := os.ReadDir(bpm.DefaultProcPrefix)
 	if err != nil {
-		return nil, errors.Wrapf(err, "read /proc/pid/tasks , ppid : %d", ppid)
+		return nil, errors.Wrapf(err, "read /proc/pid/stat , ppid : %d", ppid)
```
g1eny0ung, line 53:
```suggestion
		return nil, errors.Wrapf(err, "read /proc")
```

g1eny0ung (changes requested):

zhanluxianshen:
Thanks for review.
Rebased.

g1eny0ung (approved):

PR #4384, opened 2024-03-26
Add a common.sh to make it compabible with MacOS

see-quick:

This PR solves a problem, that I had during building chaos-mesh locally i.e., 
```java
...
./hack/update_install_script.sh
sed: preserving permissions for './sedeoFShX': Permission denied
sed: preserving permissions for './sedj9R9jW': Permission denied
sed: preserving permissions for './sed0xSYWY': Permission denied
sed: preserving permissions for './sedpu3wN1': Permission denied
sed: preserving permissions for './sedePNZZ0': Permission denied
sed: preserving permissions for './seddW6CR3': Permission denied
sed: preserving permissions for './sedIIbGR6': Permission denied
sed: preserving permissions for './sedvFKiL5': Permission denied
sed: preserving permissions for './sed8mPtg9': Permission denied
sed: preserving permissions for './sedVawu17': Permission denied
sed: preserving permissions for './sedcyJtTb': Permission denied
sed: preserving permissions for './sed9OlwHa': Permission denied
sed: preserving permissions for './sediqTSde': Permission denied
sed: preserving permissions for './sedTzIhvc': Permission denied
sed: preserving permissions for './sed2C6F4f': Permission denied
sed: preserving permissions for './sedZHJPbf': Permission denied
...
```
So I then used `gsed` and it worked without any problem. Also, I have decided to not change all scripts because other scripts worked fine (did not try all of them but if it works I think it's fine and there is no need to change that)

## Related changes

No, does not need to update the UI interface or docs

### CHANGELOG

### DIFF
--- a/hack/common.sh
+++ b/hack/common.sh
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+# Copyright Chaos Mesh Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#
+# Compatibility script.
+#
+
+# Initialize variables with default tool names
+SED=sed
+GREP=grep
+
+# Function to check if a command exists
+command_exists() {
+    command -v "$1" >/dev/null 2>&1
+}
+
+# Check if running on macOS and adjust tool names
+if [[ "$OSTYPE" == "darwin"* ]]; then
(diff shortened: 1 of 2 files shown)


g1eny0ung:
Ref: https://cloud-native.slack.com/archives/C0193VAV272/p1711034978076309.

g1eny0ung:
image

I found that when I was running `make install.sh` locally with docker desktop, I got the same error. Perhaps it's not related to different systems.

image

After I thought about it, and then switched from VirtioFS to gRPC FUSE, the problem was solved.

I also found a discussion with this problem: https://forums.docker.com/t/sed-couldnt-open-temporary-file-xyz-permission-denied-when-using-virtiofs/125473/5.

So this problem may be solved by upgrading the sed version to 4.8. I'll try it.

g1eny0ung:
@see-quick After #4388 is merged, this problem should be resolved.

I've thought about it some more, and I think adding a common.sh script might not be good practice either. Since gsed, and ggrep, for example, don't come with the Mac system, and users who don't have these tools installed should be confused by errors when they run the scripts, I'd recommend running these scripts in docker (by running Make targets).

So I would close this PR. But I still thank you for suggesting a solution!

PR #4395, opened 2024-04-15
fix(helm): should not set webhook caBundle when cert-manager enabled

erikgb:

We use FluxCD to install chaos-mesh, and prefer to use cert-manager for all PKI-related in our clusters. When enabling cert-manager in chaos-mesh Helm chart, the templates for the validating and mutating webhook set a dummy value for the webhook `caBundle` field. This is not correct IMO, and causes Flux and cert-manager to "fight" about managing the field - since both are reconciling the resources periodically.

This PR removes the dummy value set for webhook `caBundle` field when cert-manager enabled, which will allow cert-manager (ca-injector) to fully manage this particular field in the resources.

## Related changes

- [x] release-2.6
- [x] release-2.5

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/helm/chaos-mesh/templates/mutating-admission-webhooks.yaml
+++ b/helm/chaos-mesh/templates/mutating-admission-webhooks.yaml
@@ -38,9 +38,7 @@ metadata:
 webhooks:
   {{- range $crd := .Values.webhook.CRDS }}
   - clientConfig:
-      {{- if $certManagerEnabled }}
-      caBundle: Cg==
-      {{- else }}
+      {{- if not $certManagerEnabled }}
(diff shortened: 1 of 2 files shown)


Gallardot (changes requested):
@erikgb Thank you for your contribution. LGTM, but please make CI happy.

erikgb:
Thanks @Gallardot! The failing integration test seems to be unrelated (a timeout). For the changelog check: do you think this change requires a changelog entry or not?

PR #4404, opened 2024-04-23
chore: added openshift-specific helm values

var-i:

Removes need for manual configuration in Openshift environments

An additional RoleBinding is added to add the chaos-daemon and chaos-dns-server to the "privileged" openshift SCC

## Related changes

- [ ] release-2.6
- [ ] release-2.5

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [x] Manual test

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/helm/chaos-mesh/templates/openshift-scc.yaml
+++ b/helm/chaos-mesh/templates/openshift-scc.yaml
@@ -1,0 +1,17 @@
+{{- if .Values.openshift }}
+kind: RoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ .Release.Name }}-scc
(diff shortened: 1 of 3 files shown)


STRRL:
Hi @var-i , thanks for your contribution, however we do not encouraged that intruducing too many things about platform-specific configurations.

For example, we would also NOT introduce dedicated configuration fields for other kubernetes distribution.

It's welcome to make a section in FAQ to introduce your solution with openshift: https://chaos-mesh.org/docs/faqs/

How do you think about it?

g1eny0ung:
Due to the long delay in getting a reply, I will close this PR.

PR #4459, opened 2024-07-23
fix: chaos controller can't find daemonIP over 1000 nodes using endpoints (#4421)

g1eny0ung:
## Related changes

- [ ] release-2.6
- [ ] release-2.5

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

```shell
git commit --amend --signoff
git push --force
```


g1eny0ung:
@STRRL The [EndpointSlices](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/) API is marked as stable in Kubernetes v1.21, so I guess this is the reason of some e2e tests failed in `v1.20.15`? Should we update the k8s matrix to the same level as the `master`? Since `release-2.5` is also a stable branch.

STRRL:
due to the kubernetes version support policy, chaos mesh 2.6.x would continue support 1.20, so we would not make this PR merged.

PR #4531, opened 2024-10-24
fix: set default of ChaosControllerConfig.DNSServicePort to 53

### LINKED ISSUE #4217: Struct field's type and default value's type mismatch in `pkg/config/controller.go`
https://github.com/chaos-mesh/chaos-mesh/blob/87b6322f9b46548053e03731c520652a5d7e2f4b/pkg/config/controller.go#L96

The empty string can not be the default value for an `int` field. In this particular case, the correct default value seems to be `53` (the default DNS port number)

(found by [revive](https://github.com/mgechev/revive) while testing it on chaos-mesh code base)

chavacava:
Fix #4217 by setting ChaosControllerConfig.DNSServicePort default value to 53

### CHANGELOG

- [X] I have labeled this PR with "no-need-update-changelog"

### Tests

- [ ] Unit test
- [ ] E2E test
- [X] Manual test

### Side effects

### DIFF
--- a/pkg/config/controller.go
+++ b/pkg/config/controller.go
@@ -93,7 +93,7 @@ type ChaosControllerConfig struct {
 
 	// DNSServiceName is the name of DNS service, which is used for DNS chaos
 	DNSServiceName string `envconfig:"CHAOS_DNS_SERVICE_NAME" default:""`
-	DNSServicePort int    `envconfig:"CHAOS_DNS_SERVICE_PORT" default:""`
+	DNSServicePort int    `envconfig:"CHAOS_DNS_SERVICE_PORT" default:"53"`
 
 	// SecurityMode is used for enable authority validation in admission webhook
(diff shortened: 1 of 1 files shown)


STRRL:
Hi @chavacava ,when chaos mesh is running, the CHAOS_DNS_SERVICE_PORT would be set to `9288`

ref: 

https://github.com/chaos-mesh/chaos-mesh/blob/64a7a81821a074e919e0422c17e89ee55d72ef50/helm/chaos-mesh/values.yaml#L457

https://github.com/chaos-mesh/chaos-mesh/blob/64a7a81821a074e919e0422c17e89ee55d72ef50/helm/chaos-mesh/templates/controller-manager-deployment.yaml#L115

STRRL:
so please make the default value as 9288.

STRRL:
> (found by [revive](https://github.com/mgechev/revive) while testing it on chaos-mesh code base)

btw I noticed that maybe your're considering adopt chaos mesh? could you share more information about that?

PR #4546, opened 2024-11-27
fix!: use consistent units for netem delay/jitter

### LINKED ISSUE #4545: [Bug]: NetworkChaos delay times being passed as microseconds
### Chaos Mesh Version

v2.7.0

### Kubernetes Version

v1.28.13

### Describe the bug

The latency and jitter values on Delay experiments appear to be passed with incorrect magnitude to `tc`.
When I apply the NetworkChaos as defined below, I see no effect when I curl from the "frontend" to the "backend" pod (status is marked as injected).
When I increase the latency x1000 (ie `50,000ms`), then I get the expected 50ms increase in latency (measured with `time curl ...`, from 0.025s to 0.075s)

```
kind: NetworkChaos
metadata:
  name: network-delay-50ms
  namespace: chaos-mesh
spec:
  action: delay
  mode: all
  direction: to
  selector:
    namespaces:
      - default
    labelSelectors:
      tier: frontend
  delay:
    latency: "50ms"
    correlation: "0"
  target:
    mode: all
    selector:
      namespaces:
        - default
      labelSelectors:
        tier: backend
  duration: "5m"
```

The generated `tc` command in the chaos-daemon pod log is `tc qdisc add dev eth0 root netem delay 50000`, which doesn't have the desired effect.
If I run either of the below 2 commands manually, then I get a ~50ms delay as expected:
```
tc qdisc add dev eth0 root netem delay 50ms
tc qdisc add dev eth0 root netem delay 50000000
```

### To Reproduce

_No response_

andreyzher:

Fixes #4545 

NetworkChaos latency and jitter values now preserve their format and are passed as strings with units to `netem delay`, instead of micro/nanoseconds.

`time.ParseDuration` is more permissive than `netem delay` in terms of the duration expressions it can parse, so I've applied a pattern-based validation on the corresponding CRD fields to keep inputs compatible.

## Related changes

- [ ] release-2.7
- [ ] release-2.6

### CHANGELOG

### Tests

- [x] Unit test
- [ ] E2E test
- [x] Manual test

### Side effects

## DCO

> [!TIP]

```shell
git commit --amend --signoff
git push --force
```


pkg/chaosdaemon/types.go
```diff
@@ -16,22 +16,34 @@
...
 
 import (
+	"time"
+
 	"github.com/vishvananda/netlink"
 
 	pb "github.com/chaos-mesh/chaos-mesh/pkg/chaosdaemon/pb"
 )
 
 func ToNetlinkNetemAttrs(netem *pb.Netem) netlink.NetemQdiscAttrs {
```
andreyzher, line 26: This method doesn't appear to be referenced anywhere within the Chaos Mesh codebase - all former references seem to have been refactored out.

STRRL: we could safely remove this file so :)

andreyzher: Gone! Wasn't sure if it was retained for some sort of backwards compatibility, or just overlooked.

PR #4559, opened 2024-12-12
chore: Add support for userInfo.extra in validating webhook

miketonks-form3:

We experienced problems in our GKE and AKS stacks, where user authentication does not work correctly creating chaos-mesh CRDs.

We received errors such as:

`Error from server (Forbidden): error when creating "/experiment.yaml": admission webhook "vauth.kb.io" denied the request: user@my.domain is forbidden on namespace test`

On investigation, we identified that the validating webooks are at fault, since they do not pass all user information: `username` and `groups` are passed, but not `extra` field which contains cloud provider specific auth tokens.

This PR adds `extra` field to the request.  Tested in the development stacks and worked as expected.

Test can be reproduced by creating a SAR with the following bash command:

```
kubectl --context my-gke-stack auth whoami -o json | jq '{
  "apiVersion": "authorization.k8s.io/v1",
  "kind": "SubjectAccessReview",
  "spec": {
    "resourceAttributes": {
      "namespace": "nats",
      "verb": "create",
      "group": "chaos-mesh.org",
      "resource": "networkchaos"
    },
    "user": .status.userInfo.username,
    "groups": .status.userInfo.groups,
     "extra": .status.userInfo.extra
  }
}' | kubectl --context  my-gke-stack create --v=9 -f -  
```

`"status":{"allowed":true,"reason":"access granted by IAM permissions."`

For refrence, see kubernetes SubjectAccessReview:
- https://dev-k8sref-io.web.app/docs/authorization/subjectaccessreview-v1/

## Related changes

- [ ] release-2.7
- [ ] release-2.6

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [x] Manual test

### Side effects

## DCO

> [!TIP]

```shell
git commit --amend --signoff
git push --force
```


STRRL (approved):
LGTM! 

BTW, I noticed there is another field called `UID`, maybe we should also set this field?

reference: 

https://github.com/kubernetes/kubernetes/blob/4024390d8c8a19056ab7ced95eef5cce43c8096d/staging/src/k8s.io/apiserver/plugin/pkg/authorizer/webhook/webhook.go#L191-L196

How do you think about it?

STRRL:
and please fix the DCO, follow the instruction

image

https://github.com/chaos-mesh/chaos-mesh/pull/4559/checks?check_run_id=34642866233

It will ask you to execute `git rebase` to update the commit message, append `sign-off` info

miketonks-form3:
Hey, thanks for the review.  Yes, adding UID as well seems like a good idea.  Updated

STRRL:
We're facing some issues with arm64 integration tests; anyway I will over take this pr and merge it.

STRRL:
Thanks @miketonks-form3 for your contribution!

PR #4589, opened 2025-02-08
ci: remove arm64 in integration_test

g1eny0ung:
## What problem does this PR solve?

Ref: #4564

We took much time on the integration test with the arm64 arch, and it's still not working properly 🥲.

In the latest bi-weekly meeting, our maintainers/committers agreed we can temporarily disable the arm64 matrix in the `integration_test.yml` until we have enough energy to solve it deeply.

## What's changed and how it works?

Disable the integration test on arm64.

## Related changes

- [ ] This change also requires further updates to the [website](https://github.com/chaos-mesh/website) (e.g. docs)
- [ ] This change also requires further updates to the `UI interface`

## Cherry-pick to release branches (optional)

> This PR should be cherry-picked to the following release branches:

- [x] release-2.7
- [ ] release-2.6

### CHANGELOG

> Must include at least one of them.

- [ ] I have updated the `CHANGELOG.md`
- [x] I have labeled this PR with "no-need-update-changelog"

### Tests

> Must include at least one of them.

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

- [ ] **Breaking backward compatibility**

## DCO

If you find the DCO check fails, please run commands like below to fix it:

> [!TIP]
> Depends on actual situations, for example, if the failed commit isn't the most recent
> one, you can use `git rebase -i HEAD~n` to re-signoff the commit.

```shell
git commit --amend --signoff
git push --force
```

PR #4590, opened 2025-02-08
fix: syntax of integration test

g1eny0ung:

Referring to #4589, I naively thought that removing the arm64 key-value pair would disable the arm64 arch, but it still requires some additional modifications...

## Related changes

- [x] release-2.7
- [ ] release-2.6

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

> [!TIP]

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/.github/workflows/e2e_test.yml
+++ b/.github/workflows/e2e_test.yml
@@ -28,7 +28,6 @@ jobs:
             .gitignore
             **.md
             .github/ISSUE_TEMPLATE/*
-            .github/pull_request_template.md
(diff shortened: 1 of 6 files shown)


g1eny0ung:
@STRRL Just a reminder. In this PR, I also replaced the old kind-action with setup-minikube. Now our CI workflows will only use minikube, this can prevent potential environmental problems.

PR #4630, opened 2025-02-26
feat: upgrade to gorm v2

### LINKED ISSUE #4825: bump gorm v1 to v2
for more context: https://github.com/chaos-mesh/chaos-mesh/pull/4800#issuecomment-3816253244

g1eny0ung:

Resolve #4825

## Related changes

- [ ] release-2.8
- [ ] release-2.7

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

> [!TIP]

```shell
git commit --amend --signoff
git push --force
```


g1eny0ung:
I also found a problem when I updated this PR: at https://github.com/g1eny0ung/chaos-mesh/blob/8bea6c67641ed80aff8649e7504ab5e55fe55aad/pkg/dashboard/collector/collector.go#L77, when a chaos object was updated, the set operation was triggered at the same time. However, since frequent updates are not needed here, I think we should put the database synchronization logic in the apiserver creation function.

g1eny0ung:
> I also found a problem when I updated this PR: at https://github.com/g1eny0ung/chaos-mesh/blob/8bea6c67641ed80aff8649e7504ab5e55fe55aad/pkg/dashboard/collector/collector.go#L77, when a chaos object was updated, the set operation was triggered at the same time. However, since frequent updates are not needed here, I think we should put the database synchronization logic in the apiserver creation function.

I'm now using a new method to determine whether to update: since the spec of experiments actually is immutable after creation. So we only need to update the `FinishTime` (aka `DeletionTimestamp`, which means this experiment has already been deleted, so we need to archive it). This differs from #4854, which adds a new field to the database table, we can solve this problem without adding new fields.

PR #4649, opened 2025-03-13
Support for deploying extra Kubernetes objects with the helm chart

remyj38:

Allow to deploy extra objects with the chart (ExternalSecret, Traefik IngressRoute, ...)

Added an `extraObjects` value in the charts values and use it during templating

### CHANGELOG

- [X] I have updated the `CHANGELOG.md`

### Tests

- [X] Manual test


STRRL:
I am not sure what exactly object should be deployed with this helm chart, and it basically allows any object to be created when installation...

So I prefer to not to do this with a certain helm chart(eg. chaos mesh), it should be done with your CD workflow, like an extra step after helm install chaos-mesh

but we're open to hear your idea, feel free to leave any comments.

enneitex:
Hi, +1 for this PR. We are deploying chaos mesh with ArgoCD and we have to build a second helm chart for CR like ExternalSecrets. An `extraObjects` would solve that.

This is a pretty common feature:
- [kube-prometheus-stack](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L5301)
- [traefik](https://github.com/traefik/traefik-helm-chart/blob/master/traefik/values.yaml#L924)
- [External Secret Operator](https://github.com/external-secrets/external-secrets/blob/main/deploy/charts/external-secrets/values.yaml#L135)
- [Velero](https://github.com/vmware-tanzu/helm-charts/blob/main/charts/velero/values.yaml#L182)
- [All Bitnami helm charts](https://github.com/bitnami/charts/blob/main/template/README.md?plain=1#L10)

PR #4657, opened 2025-03-30
fix reduce useless recursively expanded variables in Makefile

Andrewmatilde:
In the original Makefile, `RUN_IN_DEV_SHELL` and `RUN_IN_BUILD_SHELL` were defined using = (recursively expanded variables). This notation defers evaluation until the variable is actually used, which can be helpful if the values need to reflect changes in other variables over time. However, for the variables updated in this PR, we do not require lazy (deferred) evaluation. 

Therefore, switching to := (simply expanded variables) is more appropriate, as each variable is expanded once at the time of definition and remains consistent throughout the build. This can improve performance and clarity by avoiding unnecessary re-evaluation.

This change prevents many repeated computations from being executed multiple times—a behavior that can be observed by setting SHELL := bash -x.

## Related changes

- [ ] release-2.7
- [ ] release-2.6

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

> [!TIP]

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,9 +55,9 @@ BASIC_IMAGE_ENV=IMAGE_DEV_ENV_TAG=$(IMAGE_DEV_ENV_TAG) \
 	IMAGE_TAG=$(IMAGE_TAG) TARGET_PLATFORM=$(TARGET_PLATFORM) \
 	GO_BUILD_CACHE=$(GO_BUILD_CACHE)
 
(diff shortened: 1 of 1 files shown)


Andrewmatilde:
PTAL @STRRL @g1eny0ung

Issue #4707, opened 2025-06-18
[Bug]: chaos-dashboard doesn't include chaos mesh labels


juanmancebo:
### Chaos Mesh Version

master

### Kubernetes Version

v1.31.7

### Describe the bug

When configuring Chaos Mesh helm chart, chaos-dashboard service doesn't read labels defined in `chaos-mesh.labels`

juanmancebo:
fixed [there](https://github.com/chaos-mesh/chaos-mesh/pull/4708)

PR #4708, opened 2025-06-18
fix: chaos dashboard labels in helm templates

juanmancebo:

Be able to use labels in chaos-dashboard service.
https://github.com/chaos-mesh/chaos-mesh/issues/4707

Use same labels logic as in other resources.


cwen0 (approved):
LGTM

PR #4702, opened 2025-06-02
chore!: disable chaosctl by default

g1eny0ung:
## What problem does this PR solve?

Close the chaosctl server by default to reduce potentially unsafe queries.

## What's changed and how does it work?

Update `enableCtrlServer` to `false` in values.

## Related changes

- [x] This change also requires further updates to the [website](https://github.com/chaos-mesh/website) (e.g. docs)
- [ ] This change also requires further updates to the `UI interface`

## Cherry-pick to release branches (optional)

> This PR should be cherry-picked to the following release branches:

- [x] release-2.7
- [ ] release-2.6

### CHANGELOG

> Must include at least one of them.

- [x] I have updated the `CHANGELOG.md`
- [ ] I have labeled this PR with "no-need-update-changelog"

### Tests

> Must include at least one of them.

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

- [ ] **Breaking backward compatibility**

## DCO

If you find the DCO check fails, please run commands like below to fix it:

> [!TIP]
> Depends on actual situations, for example, if the failed commit isn't the most recent
> one, you can use `git rebase -i HEAD~n` to re-signoff the commit.

```shell
git commit --amend --signoff
git push --force
```

PR #4724, opened 2025-07-21
chore!: remove chaosctl

g1eny0ung:

Ref: #4702

We haven't updated `chaosctl` for a long time, and we don't get valid feedback from our users on its use (it's rarely used). So our team's final decision is to remove it in the next release, and to consider exploring new forms of it (since we could achieve the same functionality without the additional use of a GraphQL server).

Remove all `chaosctl` code.

## Related changes

- [ ] release-2.7
- [ ] release-2.6

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

- [x] **Breaking backward compatibility**

## DCO

> [!TIP]

```shell
git commit --amend --signoff
git push --force
```


spkane:
Just a note that your current web docs still refer to `chaosctl` in https://chaos-mesh.org/docs/chaosctl-tool/

STRRL:
> Just a note that your current web docs still refer to `chaosctl` in https://chaos-mesh.org/docs/chaosctl-tool/

aha good catch, would you like to help us remove them from https://github.com/chaos-mesh/website ? ❤️

spkane:
> > Just a note that your current web docs still refer to `chaosctl` in https://chaos-mesh.org/docs/chaosctl-tool/
> 
> aha good catch, would you like to help us remove them from https://github.com/chaos-mesh/website ? ❤️

No, I don't think so. I took a look, but there are a few references to `chaosctl` related to creating TLS certs and such things, so it likely needs to be removed by someone who knows where those references should point to nowadays, and that is not me unfortunately.

spencercjh:
Without chaosctl, how can we complete the task described here?: https://chaos-mesh.org/zh/docs/chaosctl-tool/#%E4%B8%BA-chaosd-%E7%94%9F%E6%88%90-tls-%E8%AF%81%E4%B9%A6

g1eny0ung:
@STRRL @spkane @spencercjh

I have removed the relevant content in the latest version of the documentation. The old documentation and executable file will still be retained and can still be downloaded and used via the existing link.

PR #4786, opened 2025-11-04
Enable configuration of experiments that are available in the dashboard UI

kbabiak:

We need to be able to control the list of experiments available to the user in the UI, both when creating a new Experiment and when defining a Workflow.

As part of the chaos-dashboard configuration, you can define the list of experiments visible to the user in the Dashboard UI. The configuration is provided via an environment variable named ENABLED_EXPERIMENTS. Example: ENABLED_EXPERIMENTS=PodChaos,StressChaos,HTTPChaos,TimeChaos. Dashboard UI retrieves the configuration through a request to /api/common/config.

## Related changes

- [ ] release-2.7
- [ ] release-2.6

### CHANGELOG

### Tests

- [ ] Unit test
- [ ] E2E test
- [x] Manual test

### Side effects

## DCO

> [!TIP]

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/ui/app/src/components/NewExperimentNext/Step1.tsx
+++ b/ui/app/src/components/NewExperimentNext/Step1.tsx
@@ -116,15 +116,6 @@ const Step1 = () => {
     },
   })
 
-  // Auto-switch to k8s if in physic mode but no physical experiments are enabled
-  const hasPhysical = hasPhysicalMachineExperimentsEnabled(config?.enabled_experiments)
-  useEffect(() => {
-    if (env === 'physic' && !hasPhysical) {
-      setEnv('k8s')
-      setKindAction(['', ''])
-    }
-  }, [env, hasPhysical, setEnv, setKindAction])
-
   const typesData = env === 'k8s' ? _typesData : dataPhysic
   let typesDataEntries = Object.entries(typesData) as [Kind, Definition][]
   if (!config?.dns_server_create) {


STRRL:
Hi @kbabiak, thanks for your contribution!

I am not sure about we'd merge this PR into codebase, it seems rather than hide this option on frontend, it's better to use RBAC or OPA webhooks to control user's behavior.  

Feel free to share your idea and leave any comments, Thanks!

kbabiak:
Hi @STRRL, thanks for your feedback!

I understand your point about using RBAC or OPA webhooks to control user behavior, and I agree that such mechanisms are appropriate for fine-grained, per-user authorization. However, this proposal focuses on a different layer - global UI configuration.

A similar concept already exists in Chaos Mesh: on the controller-manager side, it's possible to configure the ENABLED_CONTROLLERS environment variable to define which controllers are active (* meaning all). The proposed ENABLED_EXPERIMENTS variable extends this idea to the dashboard level, allowing operators to specify which experiment types should be visible to users in the UI.

Showing users experiments they cannot actually execute can be confusing and lead to unnecessary support questions or failed attempts. By hiding unsupported experiment types in the dashboard, we provide a cleaner and more intuitive user experience. What do you think ?

This change does not exclude or conflict with RBAC or OPA-based authorization. In fact, both mechanisms can complement each other - the dashboard configuration defines a global baseline of available experiments, while RBAC can still enforce per-user or per-namespace access policies.

STRRL:
@copilot is there a way to unify this config for both controller side and dashboard side?

kbabiak:
Hi @STRRL, Happy New Year!

One viable approach to unify this configuration for both the controller side and the dashboard side would be to lift the "enabledControllers" setting in the official Helm chart one level up in values.yaml, similar to how "enableProfiling" is structured.

By placing enabledControllers at a top-level chart value, we could apply the same configuration both to the controller-manager and to the dashboard UI (e.g., feeding it into both the controller environment and the dashboard ENABLED_EXPERIMENTS config).

If this solution is acceptable to the maintainers/reviewers, I can update this PR accordingly to reflect that chart change.

PR #4821, opened 2026-01-28
fix: jvm latency experiment

ajax-koval-d:

Jvm latency experiment was failing for jdks version 19 and higher, as Byteman was not able to find matching method for Thread.sleep.

```
Rule.ensureTypeCheckedCompiled : error type checking rule systems.ajax.gradle.byteman.BytemanContuniousTest-test-latency-1769592601
org.jboss.byteman.rule.exception.TypeException: MethodExpression.typeCheck : invalid method sleep for target class java.lang.Thread file /tmp/rule.btm508946732 line 8
    at org.jboss.byteman.rule.expression.MethodExpression.findMethod(MethodExpression.java:331)
    at org.jboss.byteman.rule.expression.MethodExpression.typeCheck(MethodExpression.java:187)
    at org.jboss.byteman.rule.Action.typeCheck(Action.java:106)
    at org.jboss.byteman.rule.Rule.typeCheck(Rule.java:601)
    at org.jboss.byteman.rule.Rule.ensureTypeCheckedCompiled(Rule.java:529)
    at org.jboss.byteman.rule.Rule.execute(Rule.java:808)
    at org.jboss.byteman.rule.Rule.execute(Rule.java:789)
    at systems.ajax.gradle.byteman.BytemanContuniousTest.test(BytemanContuniousTest.kt)
    at systems.ajax.gradle.byteman.BytemanContuniousTestKt.main(BytemanContuniousTest.kt:15)
    at systems.ajax.gradle.byteman.BytemanContuniousTestKt.main(BytemanContuniousTest.kt)
```

Starting from java 19 new [overload of sleep method](https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/lang/Thread.html#sleep(java.time.Duration)) was introduced which had Duration object as an argument.

Rule that chaos mesh provided was passing Integer value while Thread.sleep expects Long. After new overload was added, Byteman was not able to resolve which method should it use since there were no exact match by arguments.

The fix is to explicitly provide literal of type Long as Thread.sleep argument, allowing Byteman to unambiguously resolve correct overload of sleep method.

## Related changes

- [ ] release-2.7
- [ ] release-2.6

### CHANGELOG

### Tests

- [x] Unit test
- [ ] E2E test
- [x] Manual test

### Side effects

## DCO

> [!TIP]

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -58,6 +58,7 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
 - Fix dashboard subpath serving by adding base path to Vite config [#5016](https://github.com/chaos-mesh/chaos-mesh/pull/5016)
 - Optimize RBAC Token Generator copy buttons in Dashboard UI [#4941](https://github.com/chaos-mesh/chaos-mesh/pull/4941)
 - Fix a TypeError crash in the Dashboard UI when opening the details page of a Workflow-type schedule [#5034](https://github.com/chaos-mesh/chaos-mesh/pull/5034)
+- Fixed jvm latency experiment not working for jdk version 19 and higher [#4821](https://github.com/chaos-mesh/chaos-mesh/pull/4821)
 
 ### Security
 
--- a/controllers/chaosimpl/jvmchaos/impl.go
+++ b/controllers/chaosimpl/jvmchaos/impl.go
@@ -220,7 +220,7 @@ func generateRuleData(spec *v1alpha1.JVMChaosSpec) error {
 
 	switch spec.Action {
 	case v1alpha1.JVMLatencyAction:
-		bytemanTemplateSpec.Do = fmt.Sprintf("Thread.sleep(%d)", spec.LatencyDuration)
+		bytemanTemplateSpec.Do = fmt.Sprintf("Thread.sleep(%dL)", spec.LatencyDuration)
 	case v1alpha1.JVMExceptionAction:
 		bytemanTemplateSpec.Do = fmt.Sprintf("throw new %s", spec.ThrowException)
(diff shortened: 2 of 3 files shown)


STRRL:
Hi @ajax-koval-d thanks for your contributrion! 

Mostly LGTM, only one more thing that I need to take a look about the why the integer parameter would match `Duration` in JDK 19, and then try to introduce some testcase ...

ajax-koval-d:
Hi @STRRL, thanks for your feedback, 

As to why sleep with int parameter is not resolved correctly since JDK 19 - I've debugged byteman a bit and it looks like the problem is in [this](https://github.com/bytemanproject/byteman/blob/9912e2eb1928cddc74dcee3a598af7ea8aa21157/agent/src/main/java/org/jboss/byteman/rule/expression/MethodExpression.java#L285-L303) loop. Previously there was only one candidate, so `pruneCandidates` call was omitted. After second candidate was added (`sleep(Duration)`), `candidateType` becomes UNDEFINED, causing `pruneCandidates` call. And that method removes both candidates - it does `isAssignableFrom` check, and `long.isAssignableFrom(int)` is `false`.

Regarding tests - I believe some e2e test would be good here, but I'll need some help probably - I can't find any instruction on how to properly run or write those

STRRL (approved):
Thanks for the fix and the detailed root cause analysis! The unit tests cover the rule generation change. We will add JVMChaos e2e coverage in a follow-up PR.

STRRL:
Thanks for your contribution! :heart: Rebased onto latest master to resolve the CHANGELOG conflict, auto-merge is enabled.

PR #4836, opened 2026-02-26
feat(remotecluster): decouple Helm lifecycle from multi-cluster contr…

### LINKED ISSUE #4831: Decouple Helm lifecycle management from multi-cluster controller coordination in RemoteCluster
## Problem

The `RemoteCluster` controller currently handles two distinct responsibilities in a single reconciliation loop:

1. **Helm lifecycle management** — installing, upgrading, and uninstalling Chaos Mesh on the remote cluster via Helm
2. **Multi-cluster controller coordination** — spawning a controller manager that watches the remote cluster for chaos experiment resources

These two concerns are tightly coupled in `controllers/multicluster/remotecluster/controller.go`. The reconciler always calls `ensureHelmRelease()` (line 128) before `ensureClusterControllerManager()` (line 134), with no way to use one without the other.

This coupling causes problems in real-world scenarios:

- **Air-gapped environments**: The Helm chart download from `https://charts.chaos-mesh.org` fails, which blocks the entire reconciliation — including the controller coordination that could otherwise work fine with a pre-installed Chaos Mesh.
- **Pre-installed Chaos Mesh**: Users who manually manage Chaos Mesh on the remote cluster (e.g., with custom RBAC, specific versions, or organization-specific Helm values) still get an unwanted Helm upgrade triggered on top of their installation.
- **Principle of least privilege**: Users who manage their own installation need clear documentation on the minimum RBAC permissions required on the remote cluster, which is currently lacking.

## Current Behavior

```
Reconcile()
  ├── ensureHelmRelease()              # Step 1: Helm lifecycle (always runs)
  │     ├── release not found → InstallRelease()
  │     └── release exists   → UpgradeRelease()
  └── ensureClusterControllerManager() # Step 2: controller coordination (always runs)
```

Step 2 depends on Step 1 succeeding. If the Helm operation fails (e.g., no internet), the controller coordination is never attempted.

## Existing Workaround

The `LOCAL_HELM_CHART_PATH` environment variable allows loading the chart from a local path instead of downloading it. This helps with air-gapped environments but does **not** address the case where users want to fully manage the remote installation themselves.

## Proposed Solution

Decouple the two responsibilities using an annotation on the `RemoteCluster` resource:

```yaml
kind: RemoteCluster
metadata:
  name: remote-01
  annotations:
    chaos-mesh.org/managed-helm-lifecycle: "false"
spec:
  namespace: "chaos-mesh"
  version: "2.7.0"
  kubeConfig:
    secretRef:
      name: remote-kubeconfig
      namespace: default
      key: kubeconfig
```

When `chaos-mesh.org/managed-helm-lifecycle` is `"false"`, the controller would skip `ensureHelmRelease()` and only run `ensureClusterControllerManager()`. The reconciliation flow becomes:

```
Reconcile()
  ├── if managed-helm-lifecycle != "false":
  │     └── ensureHelmRelease()
  └── ensureClusterControllerManager()
```

This approach:

- Requires no CRD schema change
- Is fully backwards-compatible (default behavior unchanged when annotation is absent)
- Cleanly separates the two concerns at the operational level
- Gives users explicit, per-resource control

Additionally, it would be valuable to document the minimum RBAC permissions required on the remote cluster, so users who opt out of managed Helm lifecycle can set up their own installation correctly.

OlTrenin:
  
 Closes #4831      

  Introduces an annotation `chaos-mesh.org/managed-helm-lifecycle` on
  `RemoteCluster` resources. When set to `"false"`, the controller skips all Helm
  operations (install, upgrade, uninstall) and only executes controller coordination
   (`ensureClusterControllerManager`).

  **Default behavior is fully preserved** — without the annotation (or with any
  value other than `"false"`), the controller behaves exactly as before.

  ### Changes

  - **`api/v1alpha1/remote_cluster_types.go`**: Added
  `AnnotationManagedHelmLifecycle` constant with godoc documenting minimum RBAC
  requirements for self-managed installations.
  - **`controllers/multicluster/remotecluster/controller.go`**: Added
  `helmLifecycleManaged()` helper; gated `ensureHelmRelease()` and
  `uninstallHelmRelease()` calls behind it in both create/update and deletion paths.
   Removed unreachable dead code block.
  - **`controllers/multicluster/remotecluster/controller_test.go`**: Added 8
  table-driven tests covering all edge cases (nil/empty/absent annotations, explicit
   `"true"`/`"false"`, empty string, arbitrary values, capitalized `"False"`).
  - **`examples/remote-cluster.yaml`**: Added commented-out annotation example with
  RBAC documentation.

  ### Usage

  ```yaml
  kind: RemoteCluster
  metadata:
    name: cluster-xxxxxxx
    annotations:
      chaos-mesh.org/managed-helm-lifecycle: "false"
  spec:
    namespace: "chaos-mesh"
    kubeConfig:
      secretRef:
        name: chaos-mesh-02.kubeconfig
        namespace: default
        key: kubeconfig
```
  Design decisions

  - No CRD schema changes — uses annotation on ObjectMeta, no regeneration needed
  - Strict matching — only exact lowercase "false" disables Helm management (safe
  default)
  - Per-resource granularity — each RemoteCluster can independently opt out

  Related changes

  - This change also requires further updates to the website (e.g. docs)
  - This change also requires further updates to the UI interface

  Cherry-pick to release branches (optional)

  - release-2.7
  - release-2.6

### Tests

- [x] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

> [!TIP]

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/controllers/multicluster/remotecluster/controller.go
+++ b/controllers/multicluster/remotecluster/controller.go
@@ -92,18 +92,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	r.Log.Info("remote cluster", "Generation:", obj.ObjectMeta.Generation, "ObservedGeneration:", obj.Status.ObservedGeneration)
 
-	if obj.ObjectMeta.Generation <= obj.Status.ObservedGeneration {
+	managedHelm := helmLifecycleManaged(&obj)
+
+	// Annotation-only changes do not bump .metadata.generation, so we skip the
+	// generation short-circuit when Helm lifecycle is disabled to ensure the
+	// opt-out takes effect immediately on existing resources.
+	if managedHelm && obj.ObjectMeta.Generation <= obj.Status.ObservedGeneration {
 		r.Log.Info("the target remote cluster has been up to date", "remote cluster", obj.Namespace+"/"+obj.Name)
 		return ctrl.Result{}, nil
 	}
-
-	clientConfig, err := r.getRestConfig(ctx, obj.Spec.KubeConfig.SecretRef)
-	if err != nil {
-		r.Log.Error(err, "fail to get clientConfig from secret")
-		return ctrl.Result{Requeue: true}, nil
-	}
-
-	managedHelm := helmLifecycleManaged(&obj)
 
 	// if the remoteCluster itself is being deleted, we should remove the cluster controller manager
 	if !obj.DeletionTimestamp.IsZero() {
@@ -116,6 +113,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		}
 
 		if managedHelm {
+			clientConfig, err := r.getRestConfig(ctx, obj.Spec.KubeConfig.SecretRef)
+			if err != nil {
+				r.Log.Error(err, "fail to get clientConfig from secret")
+				return ctrl.Result{Requeue: true}, nil
+			}
 			err = r.uninstallHelmRelease(ctx, &obj, clientConfig)
 			if err != nil {
 				r.Log.Error(err, "fail to uninstall helm release")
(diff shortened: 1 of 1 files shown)


OlTrenin:
@STRRL hi, could you check this PR?

PR #4859, opened 2026-03-18
chore(deps): replace go-ethereum jsonrpc with creachadair/jrpc2

### LINKED ISSUE #4857: chore(deps): replace go-ethereum jsonrpc with a permissively-licensed alternative
## Background

`github.com/ethereum/go-ethereum` is currently listed as a dependency in `go.mod`, but the only part actually used in the codebase is its internal JSON-RPC client (`github.com/ethereum/go-ethereum/rpc`).

go-ethereum is licensed under **LGPL v3**, which is incompatible with chaos-mesh's **Apache 2.0** license when the code is vendored/copied. This creates a license concern and also pulls in a large dependency for a very small use case.

See also: https://github.com/chaos-mesh/chaos-mesh/issues/4830#issuecomment-4071674392

## What's used

Only **one file** uses it:

**`pkg/chaosdaemon/iochaos_server.go`**

```go
import jrpc "github.com/ethereum/go-ethereum/rpc"

// dial over stdin/stdout of the toda subprocess
client, err := jrpc.DialIO(ctx, proc.Pipes.Stdout, proc.Pipes.Stdin)

// two RPC calls
client.CallContext(timeOut, &ret, "update", actions)
client.CallContext(timeOut, &ret, "get_status", "ping")
```

## Suggested fix

Replace with a permissively-licensed JSON-RPC library, such as:

- [`creachadair/jrpc2`](https://github.com/creachadair/jrpc2) (BSD-3) — has native `channel.Line(r, w)` support for stdin/stdout connections, minimal API change
- [`sourcegraph/jsonrpc2`](https://github.com/sourcegraph/jsonrpc2) (MIT) — widely used alternative

This is a small, well-scoped change limited to a single file. Great entry point for a first contribution!

Bingtagui404:
Fixes #4857.

## What

Replace `github.com/ethereum/go-ethereum/rpc` (LGPL v3) with `github.com/creachadair/jrpc2` (BSD-3) to resolve license incompatibility with chaos-mesh's Apache 2.0 license.

## Why

go-ethereum is licensed under LGPL v3, which is incompatible with Apache 2.0 when vendored. The only usage is in `pkg/chaosdaemon/iochaos_server.go` for communicating with the toda subprocess over stdin/stdout JSON-RPC.

## Changes

**`pkg/chaosdaemon/iochaos_server.go`** (1 file):
- `jrpc.DialIO(ctx, stdout, stdin)` → `channel.Line(stdout, stdin)` + `jrpc2.NewClient(ch, nil)`
- `client.CallContext(ctx, &ret, method, args)` → `client.CallResult(ctx, method, []interface{}{args}, &ret)`

**`go.mod`**:
- Added: `github.com/creachadair/jrpc2 v0.46.0` (BSD-3)
- Removed: `github.com/ethereum/go-ethereum v1.15.11` (LGPL v3)
- Go version unchanged at `1.24.11`

## Compatibility

- `channel.Line` matches toda's stdio newline-delimited framing protocol
- Parameter wrapping with `[]interface{}{args}` preserves toda's expected `params: [[...]]` / `params: ["..."]` format
- jrpc2 v0.46.0 requires only Go 1.19, compatible with the project's Go 1.24.11

### DIFF
--- a/.github/workflows/codecov_unittest.yaml
+++ b/.github/workflows/codecov_unittest.yaml
@@ -11,6 +11,7 @@ permissions: read-all
 
 jobs:
   unitTestAndCodeCoverage:
+    if: github.repository_owner == 'chaos-mesh'
     runs-on: ubuntu-latest
     name: Unit Test And Code Coverage
     steps:
--- a/.github/workflows/e2e_test.yml
+++ b/.github/workflows/e2e_test.yml
@@ -159,9 +159,9 @@ jobs:
       fail-fast: false
       matrix:
         kubernetes-version:
-          - v1.32.2
-          - v1.31.6
-          - v1.30.10
+          - v1.35.2
+          - v1.34.5
+          - v1.33.9
         focus:
           - "[Graceful-Shutdown] [IOChaos]"
           - "[Graceful-Shutdown] [HTTPChaos]"
@@ -197,7 +197,7 @@ jobs:
       - name: Setup minikube
         uses: medyagh/setup-minikube@latest
         with:
-          minikube-version: 1.35.0
+          minikube-version: 1.38.1
           kubernetes-version: ${{ matrix.kubernetes-version }}
 
       - name: load image into minikube
--- a/.github/workflows/release_helm_chart.yml
+++ b/.github/workflows/release_helm_chart.yml
@@ -9,6 +9,7 @@ permissions: read-all
 
 jobs:
   release-chart:
+    if: github.repository_owner == 'chaos-mesh'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
--- a/.github/workflows/upload_env_image.yml
+++ b/.github/workflows/upload_env_image.yml
@@ -14,6 +14,7 @@ permissions: read-all
 
 jobs:
   build-specific-architecture:
+    if: github.repository_owner == 'chaos-mesh'
     permissions:
       # https://docs.github.com/en/packages/managing-github-packages-using-github-actions-workflows/publishing-and-installing-a-package-with-github-actions#authenticating-to-package-registries-on-github
       packages: write
--- a/.github/workflows/upload_image.yml
+++ b/.github/workflows/upload_image.yml
@@ -11,7 +11,7 @@ permissions: read-all
 
 jobs:
   build-specific-architecture:
-    if: ${{ github.repository == 'chaos-mesh/chaos-mesh' }}
+    if: github.repository_owner == 'chaos-mesh'
     permissions:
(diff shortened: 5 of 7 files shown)


g1eny0ung:
@Bingtagui404 I think we can use the latest version of `jrpc2`. What do you think?

Bingtagui404:
Thanks for the review feedback! Two updates:

**1. Fixed ignored `update` RPC error (37d94f5)**

The `update` call now checks for errors and fails fast — if it fails, toda is killed and the error is returned immediately, instead of silently proceeding to `get_status`.

**2. Re: upgrading to jrpc2 v1 (@g1eny0ung)**

I looked into upgrading to jrpc2 v1 as well.

The blocker is not that `channel.Line` was removed, but that its API changed between v0 and v1. In our current code we use:

```go
channel.Line(proc.Pipes.Stdout, proc.Pipes.Stdin)
```

With v1, `channel.Line` is no longer a constructor with `(io.Reader, io.Writer)` parameters — it's a `Framing` variable (`var Line = Split('\n')`), so this call no longer compiles and the channel setup needs to be rewritten/adapted.

To keep this PR focused on fixing the license issue and the ignored `update` RPC error, I kept the current jrpc2 version here. I can investigate the v1 migration separately in a follow-up PR if you'd like.

STRRL:
> Thanks for the review feedback! Two updates:
> 
> **1. Fixed ignored `update` RPC error ([37d94f5](https://github.com/chaos-mesh/chaos-mesh/commit/37d94f5f15218d19fcc5c7051d9361b512b8ed21))**
> 
> The `update` call now checks for errors and fails fast — if it fails, toda is killed and the error is returned immediately, instead of silently proceeding to `get_status`.
> 
> **2. Re: upgrading to jrpc2 v1 (@g1eny0ung)**
> 
> I looked into upgrading to jrpc2 v1 as well.
> 
> The blocker is not that `channel.Line` was removed, but that its API changed between v0 and v1. In our current code we use:
> 
> ```go
> channel.Line(proc.Pipes.Stdout, proc.Pipes.Stdin)
> ```
> 
> With v1, `channel.Line` is no longer a constructor with `(io.Reader, io.Writer)` parameters — it's a `Framing` variable (`var Line = Split('\n')`), so this call no longer compiles and the channel setup needs to be rewritten/adapted.
> 
> To keep this PR focused on fixing the license issue and the ignored `update` RPC error, I kept the current jrpc2 version here. I can investigate the v1 migration separately in a follow-up PR if you'd like.

Use the latest stable version please.

STRRL:
in other words, it's better to investigate which version to use before migrating, instead of migrating to 0.y.z and then struggling to migrate to 1.y.z.

feel free to push any changes or rework with a new PR :)

PR #4879, opened 2026-03-30
workflow: add EphemeralTask template type for single-run task semantics

drybalka-s:

Issues #4878

Chaos Mesh Workflow currently supports `templateType: Task`, but it does not provide a safe one-shot task mode.

In operational workflows, a task may include not only the test itself, but also shell commands with side effects such as notifications, service operations, zone selection, and alerting toggles. If the task pod disappears, the workflow may recreate the pod and re-run the whole step.

In practice, this can lead to repeated and unplanned execution of the test and all related shell commands. The most dangerous case is when the rerun happens while the original test is still running, which may cause actions to overlap across 2 availability zones.

This PR adds a dedicated `templateType: EphemeralTask` to support one-shot task execution without losing `stdout` / `exitCode` semantics used by `conditionalBranches`.

This PR introduces `templateType: EphemeralTask` with the following behavior:

- the task pod is created only once
- after the pod reaches a terminal state, `stdout` and `exitCode` are collected and persisted into `WorkflowNode`
- `conditionalBranches` are evaluated from the persisted context
- after result collection, the pod is deleted
- if the pod disappears before the result is collected, the controller records a failed outcome and does not recreate the pod
- if a completed pod is deleted before the next reconcile, the controller still does not recreate the pod

Implementation details:

- a dedicated `EphemeralTaskReconciler` is added instead of changing regular `Task` behavior inline
- `Task` behavior remains unchanged for backward compatibility
- `EphemeralTask` state is tracked in `WorkflowNode` annotations:
  - `workflow.chaos-mesh.org/ephemeral-task-pod-created`
  - `workflow.chaos-mesh.org/ephemeral-task-result-collected`
  - `workflow.chaos-mesh.org/ephemeral-task-context`
- API / webhook / bootstrap / abort / deadline handling are extended to recognize `TypeEphemeralTask`
- dashboard backend and frontend are updated so workflows containing `EphemeralTask` render correctly in the UI

## Related changes

- [x] release-2.8
- [ ] release-2.7

### CHANGELOG

### Tests

- [x] Unit test
- [x] E2E test
- [x] Manual test

### Side effects


drybalka-s:
The root cause has been identified: this was not solely a webhook issue.

The problem stemmed from the TypeEphemeralTask being added only to the generated file (api/v1alpha1/zz_generated.workflow.chaosmesh.go) but omitted from the generator source (cmd/chaos-builder/workflow.go).

During the CI process, make check rebuilt and executed chaos-builder, which regenerated zz_generated.workflow.chaosmesh.go without TypeEphemeralTask. Consequently, api/v1alpha1/workflow_webhook.go failed to compile with the error:

undefined: TypeEphemeralTaskThe fix involved correcting the generator source and subsequently regenerating the workflow file. This was verified using the following steps:
```
go build -o /tmp/chaos-builder ./cmd/chaos-builder && /tmp/chaos-builder
cd api && go test -c ./v1alpha1
```

drybalka-s:
The last commit limits the amount of task stdout persisted into `WorkflowNode.status.conditionalBranchesStatus.context`.

Previously, the full stdout was stored in node status. That kept the `stdout` / `exitCode` contract for conditional branches, but it also increased memory pressure and could make large task output expensive to keep in cache and reconciliation state.

Now the reconciler persists:
- `exitCode`
- truncated `stdout`
- `stdoutTruncated=true` when truncation happened
- `stdoutOriginalBytes=256`

This keeps conditional branch evaluation working for the intended short-output use case while reducing the amount of data retained in workflow node status.

Are there any other issues that prevent this pr from being considered?

drybalka-s:
It's been a month since we've fully transitioned to version 2.8.2 with the proposed patches. We're running about 200-300 tests a day, and everything is stable.

STRRL:
fix: https://github.com/chaos-mesh/chaos-mesh/issues/4878

drybalka-s:
ping

PR #4896, opened 2026-04-09
Partial crd support

pims:

This PR adds the ability to configure which CRDs the controllers and dashboard watch. This allows cluster operator to only install a subset of the Chaos Mesh CRDs.

Fixes #4874

[!IMPORTANT]  These changes are fully backward compatible.

The controller and dashboard look at an optional environment variable to know which CRDs they should be watching.

The PR is split into multiple commits to ease reviewing:

1. Small bug fix to the controller
2. Same change for the dashboard
3. Dashboard API returns empty lists if a CRD is not enabled
4. Helm chart changes
5. Changelog updated

## Related changes

## Checklist

Ran everything locally on a KinD cluster and manually verified that both controller and dashboards are no longer erroring when only a subset of CRDs are installed.

### CHANGELOG

### Tests

- [x] Unit test
- [ ] E2E test
- [x] Manual test

I'd be happy to add E2E tests with the proper guidance.

### Side effects

## DCO

> [!TIP]

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
 
 ### Added
 
+- Add `ENABLED_COLLECTORS` env var and `dashboard.enabledCollectors` Helm value to gate dashboard collectors and API handlers for partial CRD installations [#4874](https://github.com/chaos-mesh/chaos-mesh/issues/4874)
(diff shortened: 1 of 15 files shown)


pims:
@STRRL I pushed a fix for the dashboardconfig issue in swagger. PTAL when you have a moment.

PR #4897, opened 2026-04-09
fix: correct abstractTcFilter to use SourcePort instead of EgressPort

archy-rock3t-cloud:

`abstractTcFilter` appends `tc.EgressPort` when `tc.SourcePort` is set, causing incorrect tc filter keys when SourcePort differs from EgressPort.

Fix the error and add unit tests for `abstractTcFilter` and `convertTbfToArgs`.

## Related changes

### CHANGELOG

### Tests

- [x] Unit test

### Side effects

### DIFF
--- a/pkg/chaosdaemon/tc_server.go
+++ b/pkg/chaosdaemon/tc_server.go
@@ -507,7 +507,7 @@ func abstractTcFilter(tc *pb.Tc) string {
 	}
 
 	if len(tc.SourcePort) > 0 {
-		filter += "-" + tc.EgressPort
+		filter += "-" + tc.SourcePort
 	}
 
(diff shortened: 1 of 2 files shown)


STRRL:
Thanks for your contribution and for adding unit tests! :heart:

After looking into this, we found that `source_port` and `egress_port` on the `Tc` message are never set by the Kubernetes controllers, so the buggy branch in `abstractTcFilter` is dead code on the Kubernetes path. The actual port filtering happens through the iptables `Chain` message, which uses the correct fields.

Instead of fixing the copy-paste line, we decided to remove these unused fields entirely in #5056. Closing this one, but thanks again for spotting the bug!

PR #4908, opened 2026-04-27
test: standardize custom zap logger usage in integration tests

### LINKED ISSUE #2892: Refine logging in integration tests, use customized logger with GinkgoWriter
The library `controller-runtime` requires setting a logger (by `log.SetLogger()`) at the first 30s when the application starts, or it would use the default `NullLogSink`. We should also call it in testing codes.

When we test with ginkgo, ginkgo provides a helpful `GinkgoWriter`, which hides the output as default, only prints it when the test failed. We'd better use it to keep our testing output clean.

So here is the proposal:

- Create another constructor in `pkg/log/zap.go` like `NewDefaultZapLogger` but with the custom `io.Writer`.
- replace `log.SetLogger()` with above logger

Here are several files that need changes:

- api/v1alpha1/suite_test.go
- cmd/chaos-controller-manager/provider/suit_test.go
- controllers/common/suit_test.go
- controllers/schedule/suit_test.go
- pkg/webhook/inject/setup_test.go
- pkg/workflow/controllers/suite_test.go

About using customized with zap, you could create a logger by using `zapcore` directly. Or using `zap.Config` with registering custom sink factory with [RegisterSink](https://pkg.go.dev/go.uber.org/zap#RegisterSink). Reference: https://github.com/uber-go/zap/issues/979

Maximus-08:

This PR addresses leftover tasks from #2892 by standardizing the logger initialization in the integration test suites. Instead of instantiating loggers inline via `controller-runtime`'s `zap.New(...)`, this ensures the test suites uniformly use Chaos Mesh's custom `NewZapLoggerWithWriter(GinkgoWriter)` constructor from `pkg/log`.
Resolves #2892

- Updates `api/v1alpha1/suite_test.go` to initialize the logger using `chaoslog.NewZapLoggerWithWriter(GinkgoWriter)`.
- Aliases the `github.com/chaos-mesh/chaos-mesh/pkg/log` import to `chaoslog` in order to safely inject the logger without colliding with the existing package-level `var log` defined in `api/v1alpha1/common_types.go`.

## Related changes

- [ ] release-2.8
- [ ] release-2.7

### CHANGELOG

### Tests

- [x] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

> [!TIP]

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/api/v1alpha1/suite_test.go
+++ b/api/v1alpha1/suite_test.go
@@ -28,7 +28,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
-	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+
+	chaoslog "github.com/chaos-mesh/chaos-mesh/pkg/log"
 )
 
 // These tests use Ginkgo (BDD-style Go testing framework). Refer to
@@ -46,7 +47,7 @@ func TestAPIs(t *testing.T) {
 
 var _ = BeforeSuite(func(ctx SpecContext) {
 
-	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
+	logf.SetLogger(chaoslog.NewZapLoggerWithWriter(GinkgoWriter))
 
 	By("bootstrapping test environment")
 	t := true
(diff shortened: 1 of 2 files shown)


Maximus-08:
cc @g1eny0ung @STRRL

Maximus-08:
@STRRL There is no need to update changelog for this change, I don't have priviledges to label this PR I think, could you please take a look

PR #4922, opened 2026-05-01
test(e2e): add e2e test for external network delay with ExternalTargets

### LINKED ISSUE #1047: Add e2e test for external network delay
There is no e2e test and documents for external network delay and partition now :facepalm: .

The e2e test could be found in `/test/e2e/chaos/basic.go`. Reading other tests in that file could help accomplish this issue. 

```
kind: NetworkChaos
metadata:
  name: network-netem-example
  namespace: chaos-testing
spec:
  action: netem
  mode: one
  selector:
    labelSelectors:
  delay:
    latency: "90ms"
    correlation: "25"
    jitter: "90ms"
  loss:
    loss: "25"
    correlation: "25"
  direction: to
  target:
    selector:
      labelSelectors:
    mode: one
  duration: "10s"
  scheduler:
    cron: "@every 15s"
```

madmecodes:

Fixes #1047

There was no e2e test coverage for `NetworkChaos` with the `ExternalTargets` field. This field allows targeting IPs/domains outside Kubernetes, but the code path had no tests.

## What's changed

Added `TestcaseExternalNetworkDelay` covering two scenarios:

1. **Single external target** - applies delay chaos targeting one IP, verifies only that specific connection is slowed
2. **Multiple external targets** - applies delay chaos targeting two IPs, verifies both connections are slowed and recovery works

Uses peer pod IPs as external targets so the test is self-contained and doesn't depend on external connectivity, while still exercising the `ExternalTargets` code path in the controller.

### Files changed

- `e2e-test/e2e/chaos/networkchaos/network_external_delay.go` - new test function
- `e2e-test/e2e/chaos/networkchaos/misc.go` - new `makeExternalNetworkDelayChaos` helper
- `e2e-test/e2e/chaos/basic.go` - registered test under `[Netem][ExternalTargets]`

## Checklist

- [x] Follows existing patterns from `network_delay.go` and `misc.go`
- [x] Verified with `go vet`

### DIFF
--- a/e2e-test/e2e/chaos/basic.go
+++ b/e2e-test/e2e/chaos/basic.go
@@ -439,6 +439,9 @@ var _ = ginkgo.Describe("[Basic]", func() {
 			ginkgo.It("[PeersCrossoverWithDirectionBoth]", func() {
 				networkchaostestcases.TestcasePeersCrossover(ns, cli, networkPeers, ports, c)
 			})
+			ginkgo.It("[ExternalTargets]", func() {
+				networkchaostestcases.TestcaseExternalNetworkDelay(ns, cli, networkPeers, ports, c)
+			})
 		})
 
 		ginkgo.JustAfterEach(func() {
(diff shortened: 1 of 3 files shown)


STRRL:
could we use a real external IP like well known IPs like 1.1.1.1, 8.8.8.8?

PR #4939, opened 2026-05-05
refactor: replace context.TODO() with context.Background() in chaos-controller-manager metrics collector

Ridhi-03Kumari:

While exploring the codebase, I noticed that three background collector 
functions in `pkg/metrics/chaos-controller-manager.go` were using 
`context.TODO()`. Since these functions run as background goroutines with 
no incoming request context, `context.Background()` is more appropriate here.

Closes #4939

Replaced `context.TODO()` with `context.Background()` in:
- `collectChaosExperiments()` (line 175)
- `collectChaosSchedules()` (line 202)
- `collectChaosWorkflows()` (line 222)

`context.TODO()` is generally a temporary placeholder when the right context 
hasn't been decided yet, whereas `context.Background()` clearly communicates 
that this is a long-running background operation.

## Related changes

### CHANGELOG

### Tests
- [x] Manual test

### Side effects

### DIFF
--- a/pkg/metrics/chaos-controller-manager.go
+++ b/pkg/metrics/chaos-controller-manager.go
@@ -172,7 +172,7 @@ func (collector *ChaosControllerManagerMetricsCollector) collectChaosExperiments
 	for kind, obj := range v1alpha1.AllKinds() {
 		expCache := map[string]map[string]int{}
 		chaosList := obj.SpawnList()
-		if err := collector.store.List(context.TODO(), chaosList); err != nil {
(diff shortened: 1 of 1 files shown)


Ridhi-03Kumari:
Hi @STRRL @g1eny0ung, I've opened this PR to replace `context.TODO()` 
with `context.Background()` in the chaos-controller-manager metrics 
collector functions. 
Similar to #4938, these are background goroutines with no incoming 
request context, so `context.Background()` is more semantically correct.
Would love to get your feedback and review on this. Please let me know 
if any changes are needed. Thank you!

PR #4956, opened 2026-05-11
fix(chaosdaemon): normalize IPv6 address before shell interpolation in SetDNSServer

### LINKED ISSUE #4943: [Bug]: SetDNSServer does not normalize IPv6 addresses before shell interpolation
### Chaos Mesh Version

master

### Kubernetes Version

v1.28+ (any dual-stack cluster)

### Describe the bug

In `pkg/chaosdaemon/dns_server.go`, the `SetDNSServer` function validates the DNS server address using `net.ParseIP()` on line 52, but then discards the parsed result and uses the raw `req.DnsServer` string directly in a `sed` command on line 74.

This means non-canonical IPv6 representations (like `0:0:0:0:0:0:0:1`) pass validation just fine, but get written verbatim into `/etc/resolv.conf` instead of the canonical form (`::1`). On dual-stack Kubernetes clusters where `service.Spec.ClusterIP` could be an IPv6 address, this could lead to unexpected behavior.

The fix is straightforward — capture the `net.ParseIP()` return value and use `ip.String()` for the canonical form before interpolating it into the shell command. This is safe for IPv4 too since `ip.String()` of `"8.6.4.2"` returns `"8.6.4.2"` unchanged.

I have a fix ready and will open a PR shortly.

### To Reproduce

1. Deploy Chaos Mesh on a dual-stack Kubernetes cluster
2. Create a DNSChaos experiment targeting a pod where the chaos DNS service has an IPv6 ClusterIP
3. Alternatively, call the `SetDNSServer` gRPC endpoint directly with a non-canonical IPv6 address like `0:0:0:0:0:0:0:1`
4. Inspect the target pod's `/etc/resolv.conf` — it will contain the non-canonical form instead of the normalized `::1`

**Relevant code:**

https://github.com/chaos-mesh/chaos-mesh/blob/master/pkg/chaosdaemon/dns_server.go#L52-L74

Harishrs2006:
`SetDNSServer` validates the DNS address with `net.ParseIP` but discards
the parsed result, using the raw input string directly in the sed command.
Non-canonical IPv6 addresses like `0:0:0:0:0:0:0:1` pass validation but
get written verbatim into `/etc/resolv.conf` instead of their canonical
form (`::1`). On dual-stack clusters this produces malformed nameserver
entries that break DNS chaos injection.

Fixes: #4943

 What's changed and how it works?

Capture the `net.ParseIP` return value and use `ip.String()` when
constructing the sed command so the written address is always canonical.
Adds `Test_SetDNSServer_Enable_IPv6NonCanonical` covering the
non-canonical IPv6 → canonical normalization path.

Checklist:
Tests
- [x] Unit test: `Test_SetDNSServer_Enable_IPv6NonCanonical` added
- [ ] E2E test: not required (unit test covers the fix)

Code changes
1) [x] Has Go code change: Yes
2) [x] Has CI related scripts change: No
3) [x] Has new or updated documentation: No

Does this PR introduce a breaking change?
 
No. IPv4 and already-canonical IPv6 addresses are unaffected —
`net.IP.String()` is idempotent for canonical forms.

### DIFF
--- a/pkg/chaosdaemon/dns_server.go
+++ b/pkg/chaosdaemon/dns_server.go
@@ -49,7 +49,8 @@ func (s *DaemonServer) SetDNSServer(ctx context.Context,
 	if req.Enable {
 		// set dns server to the chaos dns server's address
 
-		if net.ParseIP(req.DnsServer) == nil {
+		ip := net.ParseIP(req.DnsServer)
+		if ip == nil {
 			return nil, ErrInvalidDNSServer
 		}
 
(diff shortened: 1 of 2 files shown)


STRRL:
thanks for your contribution!

we do not consider this is a bug. Thanks!

fyi: https://github.com/chaos-mesh/chaos-mesh/issues/4943#issuecomment-4426494970

PR #4958, opened 2026-05-11
fix: use continue instead of return in pipeline loops

### LINKED ISSUE #4957: [BUG] : pipeline AddSteps and Bootstrap loops use return instead of continue, silently dropping all subsequent steps
### Chaos Mesh Version

latest (master)

### Kubernetes Version

N/A - code bug, not environment specific

### Describe the bug

In controllers/common/pipeline/pipeline.go, the AddSteps function uses `return` when a step evaluates to nil, instead of `continue`. This exits the entire loop, silently dropping all remaining steps from the pipeline.

The same pattern exists in 3 Bootstrap loops:
- controllers/common/fx.go
- controllers/multicluster/remotechaos/fx.go
- controllers/multicluster/remotechaosmonitor/fx.go

Each step uses a unique per-step name with ShouldSpawnController (e.g. podchaos-initFinalizers, podchaos-desiredphase), implying each step is meant to be independently toggleable. But return exits the whole loop instead of just skipping that one step.

The critical impact: disabling any early step also silently drops records.Step — the step that actually injects and recovers chaos — with no error or log. The experiment appears registered but does nothing.

This is latent under the default config (ENABLED_CONTROLLERS=*) but triggers when a non-wildcard list is set.

This was introduced in #2465 (commit 409238e) which added the pipeline controller. The nil check was added as a nil-pointer guard to prevent panics in Reconcile, but return was used where continue was intended.

I have a fix ready (return → continue in all 4 loop bodies) and will open a PR immediately.

### To Reproduce

1. Set ENABLED_CONTROLLERS to a non-wildcard list that excludes one early step, e.g. podchaos-desiredphase,podchaos-condition,podchaos-records,podchaos-cleanFinalizers (omitting podchaos-initFinalizers)
2. Create a PodChaos experiment
3. Observe that chaos is never injected and no error is logged

bitflicker64:
Fixes #4957 

In AddSteps (pipeline.go) and three Bootstrap loops (fx.go, remotechaos/fx.go, remotechaosmonitor/fx.go), return/return nil was used inside for loops when ShouldSpawnController returns false. This silently drops all remaining steps instead of just skipping the disabled one.

The critical impact: disabling any early step (e.g. initFinalizers) also drops records.Step — the step that actually injects and recovers chaos — with no error or log. The bug is latent under the default config (ENABLED_CONTROLLERS=*) but triggers when a non-wildcard list is set.

Introduced in #2465 (commit 409238e).

### DIFF
--- a/controllers/common/fx.go
+++ b/controllers/common/fx.go
@@ -66,7 +66,7 @@ func Bootstrap(params Params) error {
 	for _, pair := range pairs {
 		name := pair.Name + "-records"
 		if !config.ShouldSpawnController(name) {
-			return nil
+			continue
 		}
 
 		setupLog.Info("setting up controller", "resource-name", pair.Name)
--- a/controllers/common/pipeline/pipeline.go
+++ b/controllers/common/pipeline/pipeline.go
@@ -59,7 +59,7 @@ func (p *Pipeline) AddSteps(steps ...PipelineStep) {
 	for _, step := range steps {
 		reconciler := step(p.ctx)
 		if reconciler == nil {
-			return
+			continue
 		}
 		p.controllers = append(p.controllers, reconciler)
 	}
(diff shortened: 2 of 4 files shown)


STRRL:
but a bug

fyi: https://github.com/chaos-mesh/chaos-mesh/issues/4957#issuecomment-4426430938

STRRL:
not a bug

Issue #4256, opened 2023-11-13
Failed to apply chaos: rpc error: code = Unknown desc = exit status 101


stefan-malin-db:
## Bug Report

**What version of Kubernetes are you using?**
latest stable version

**What version of Chaos Mesh are you using?**
v2.6.2

**What did you do? / Minimal Reproducible Example**
run jvm chaos with stress option
**What did you expect to see?**
delay in reponse time

**What did you see instead?**
Failed to apply chaos: rpc error: code = Unknown desc = exit status 101

**Output of chaosctl**
<img width="565" alt="image" src="https://github.com/chaos-mesh/chaos-mesh/assets/129365594/c3ad82b5-6282-40b6-a42c-676ec55a00d9">

camaeel:
Error message I get:
```
chaos-controller-manager-c4dc9f7f4-k5f5n chaos-mesh 2024-07-30T13:23:13.067Z	ERROR	jvmchaos	jvmchaos/impl.go:143	install jvm rules	{"error": "rpc error: code = Unknown desc = exit status 101"}
chaos-controller-manager-c4dc9f7f4-k5f5n chaos-mesh github.com/chaos-mesh/chaos-mesh/controllers/chaosimpl/jvmchaos.(*Impl).Apply
chaos-controller-manager-c4dc9f7f4-k5f5n chaos-mesh 	/runner/_work/chaos-mesh/chaos-mesh/controllers/chaosimpl/jvmchaos/impl.go:143
chaos-controller-manager-c4dc9f7f4-k5f5n chaos-mesh github.com/chaos-mesh/chaos-mesh/controllers/common/records.(*Reconciler).Reconcile
chaos-controller-manager-c4dc9f7f4-k5f5n chaos-mesh 	/runner/_work/chaos-mesh/chaos-mesh/controllers/common/records/controller.go:150
chaos-controller-manager-c4dc9f7f4-k5f5n chaos-mesh github.com/chaos-mesh/chaos-mesh/controllers/common/pipeline.(*Pipeline).Reconcile
chaos-controller-manager-c4dc9f7f4-k5f5n chaos-mesh 	/runner/_work/chaos-mesh/chaos-mesh/controllers/common/pipeline/pipeline.go:73
chaos-controller-manager-c4dc9f7f4-k5f5n chaos-mesh sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
[...]

PR #4971, opened 2026-05-15
fix(chaosdaemon): make JVMChaos work on distroless target containers

YifanYang6:

Addresses one specific failure mode reported in #4256: `JVMChaos` cannot install the byteman agent into target containers built on distroless / scratch-based Java base images (e.g. `gcr.io/distroless/java*`, jib- or CNB-built images). On those images every retry fails with:

```
rpc error: code = Unknown desc = exit status 101
```

and the chaos never reaches `phase: Injected`.

Root cause: `pkg/chaosdaemon/jvm_server.go` enters the target container's mount namespace and shells out to `mkdir` / `cat`:

```go
bpm.DefaultProcessBuilder("sh", "-c", "mkdir -p ...").SetNS(pid, bpm.MountNS)
bpm.DefaultProcessBuilder("sh", "-c", "cat > ...").SetNS(pid, bpm.MountNS)
```

This requires `/bin/sh` + `/bin/mkdir` + `/bin/cat` to exist in the *target* rootfs. Distroless Java images ship only `/usr/bin/java` — any binary nsexec'd into them fails with exit 101. @STRRL's quick workaround in the issue thread (drop the `sh -c` wrapper, leave bare `mkdir`) helps when `/bin/sh` is missing but `/bin/mkdir` exists; it doesn't help when neither does, which is the common case for production distroless images.

Stop entering the target mount namespace for these two filesystem operations. chaos-daemon already runs privileged in the host PID namespace and can see the container's rootfs at `/proc/<pid>/root`, so the same effect can be achieved from chaos-daemon's own view with plain Go `os.*` APIs:

- New helper `mkdirInContainer(pid, dir)` → `os.MkdirAll(filepath.Join("/proc/<pid>/root", dir), 0o755)`. Used by `InstallJVMRules` for the `$BYTEMAN_HOME/lib/` directory.
- `copyFileAcrossNS` rewritten to `os.OpenFile` + `io.Copy` on `/proc/<pid>/root/<dest>` instead of `sh -c "cat > $dest"` over `SetNS(pid, bpm.MountNS)`.

The shell-based agent-attach and rule-submit paths (`bminstall.sh`, `bmsubmit.sh`) are unchanged — those run in chaos-daemon's own container where `sh` is guaranteed to exist. No new capabilities required; chaos-daemon already traverses `/proc/<pid>/...` elsewhere.

No behavior change for non-distroless target containers — `os.MkdirAll` on `/proc/<pid>/root/<dir>` is idempotent and produces the same directory tree the previous `sh -c "mkdir -p"` did. File mode `0644` for copied jars and `0755` for the lib dir match the effective mode the previous flow produced (default umask 0022 in the privileged chaos-daemon container).

## Related changes

- [x] release-2.8
- [x] release-2.7

(The bug exists on every release that ships `pkg/chaosdaemon/jvm_server.go` with the `SetNS(pid, MountNS)` calls — confirmed reproducible on 2.8.x and present in current `master`.)

### CHANGELOG

- [x] I have updated the CHANGELOG.md

### Tests

- [x] Manual test

Validated against the Oracle Coherence Helidon `carts` service from the [sock-shop](https://github.com/oracle/coherence-helidon-sockshop-sample) sample (distroless, `/usr/bin` contains only `java`):

Before this PR:
```
rpc error: code = Unknown desc = exit status 101
(infinite retry, phase: Not Injected)
```

After this PR:
```yaml
status:
  experiment:
    containerRecords:
    - events:
      - operation: Apply
        type: Succeeded
      phase: Injected
      injectedCount: 1
```

And in the target JVM stderr (byteman):
```
Rule.execute called for com...CartResource-getItems-exception-...
Rule.execute called for com...CartResource-getItems-exception-...
```

i.e. the rule is actually firing on every invocation of the targeted method.

I'm happy to add a unit test for `copyFileAcrossNS` / `mkdirInContainer` (e.g. with a fake `/proc/<pid>/root` rooted at a tmpdir) if reviewers would like — let me know which side of the testability/scope tradeoff is preferred here.

## Other call sites affected by the same root cause

While preparing this PR I audited the other `SetNS(pid, bpm.MountNS)` call sites in `pkg/chaosdaemon/` to see which share this failure mode:

| Call site | Spawned binary | Same bug? |
|---|---|---|
| `dns_server.go` (3 sites, backup/replace/restore of `/etc/resolv.conf`) | `sh -c "ls/cp/sed/cat …"` | **Yes** — same `sh` + coreutils dependency on target rootfs. Should be rewritable along the same `/proc/<pid>/root` lines (with a small Go-side regex replace for the `nameserver` line). |
| `iochaos_server.go` (toda) | `todaBin` (chaos-mesh-shipped binary) | Same symptom, different cause — `toda` is not in the target rootfs, so `/proc/<pid>/root` doesn't help; would need a different approach (e.g. bind-mount or run from chaos-daemon view). |
| `blockchaos_server_linux.go` (`normalizeVolumeName`) | `SetNS(1, …)` targets host, not container — unaffected. |
| `jvm_server.go` (`bminstall.sh` / `bmsubmit.sh`) | gated on `NetNS` only or `EnableLocalMnt()`; `sh` resolves in chaos-daemon's own rootfs — unaffected. |

To keep this PR focused on #4256's reported `JVMChaos` failure and easy to review, I've intentionally **not** touched DNS chaos here. Happy to follow up with a second PR for DNS chaos (and possibly a separate proposal for the toda case) if reviewers prefer that split, or to expand this PR to cover DNS in one go — let me know which you'd like.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
 - Remove caBundle placeholder in webhook templates when cert-manager is enabled to fix server-side apply conflicts [#4828](https://github.com/chaos-mesh/chaos-mesh/pull/4828)
 - Add missing schedule types in dashboard collector [#4840](https://github.com/chaos-mesh/chaos-mesh/pull/4840)
 - Add missing NodeType for workflows api in dashboard [#4834](https://github.com/chaos-mesh/chaos-mesh/pull/4834)
+- Fix JVMChaos injection into distroless target containers by avoiding target-rootfs shell and coreutils dependencies [#4971](https://github.com/chaos-mesh/chaos-mesh/pull/4971)
 
 ### Security
 
--- a/pkg/chaosdaemon/jvm_server.go
+++ b/pkg/chaosdaemon/jvm_server.go
@@ -18,7 +18,9 @@ package chaosdaemon
 import (
 	"context"
 	"fmt"
+	"io"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/golang/protobuf/ptypes/empty"
@@ -28,6 +30,14 @@ import (
 	pb "github.com/chaos-mesh/chaos-mesh/pkg/chaosdaemon/pb"
 	"github.com/chaos-mesh/chaos-mesh/pkg/chaosdaemon/util"
 )
+
+// mkdirInContainer creates dir inside the target container's mount namespace
+// by writing through /proc/<pid>/root from chaos-daemon. The target rootfs may
+// be distroless (no /bin/sh, no /bin/mkdir), so we cannot rely on nsexec'ing
+// any binary into the target mount namespace.
+func mkdirInContainer(pid uint32, dir string) error {
+	return os.MkdirAll(filepath.Join(fmt.Sprintf("/proc/%d/root", pid), dir), 0o755)
+}
 
 const (
 	bmInstallCommand = "bminstall.sh -b -Dorg.jboss.byteman.transform.all -Dorg.jboss.byteman.verbose -Dorg.jboss.byteman.compileToBytecode -p %d %d"
@@ -68,14 +78,13 @@ func (s *DaemonServer) InstallJVMRules(ctx context.Context,
 	}
 
 	// Copy byteman.jar, byteman-helper.jar and chaos-agent.jar into container's namespace.
+	// Distroless target containers (e.g. Google distroless, scratch-based Java
+	// images) do not ship /bin/sh or /bin/mkdir, so nsexec'ing a shell into the
+	// target mount namespace fails with exit 101. Operate on the container
+	// rootfs from chaos-daemon directly through /proc/<pid>/root instead.
 	if req.EnterNS {
-		processBuilder := bpm.DefaultProcessBuilder("sh", "-c", fmt.Sprintf("mkdir -p %s/lib/", bytemanHome)).SetContext(ctx).SetNS(pid, bpm.MountNS)
-		output, err := processBuilder.Build(ctx).CombinedOutput()
-		if err != nil {
-			return nil, err
-		}
-		if len(output) > 0 {
-			log.Info("mkdir", "output", string(output))
+		if err := mkdirInContainer(pid, fmt.Sprintf("%s/lib", bytemanHome)); err != nil {
+			return nil, errors.Wrap(err, "create byteman lib dir in container")
 		}
 
 		jars := []string{"byteman.jar", "byteman-helper.jar", "chaos-agent.jar"}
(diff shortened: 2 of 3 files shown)


STRRL:
I did not take a deep review yet, but basically it could work 🤔

anyhow, we have `nsexec` already integrated with bpm, maybe we could try another path which have `nsOption` enabled for activating `nsexec`

YifanYang6:
@STRRL thanks for the pointer! I looked into routing this through `nsexec` with `nsOption`, but I don't think it can avoid the distroless failure for the file-copy steps — here's the reasoning:

**`nsexec` is already being activated.** `SetNS(pid, bpm.MountNS)` sets `nsOptions`, which is exactly what flips `bpm` onto the `nsexec` path in [`build_linux.go#L36-L45`](https://github.com/chaos-mesh/chaos-mesh/blob/ec1500bb9c7f79a3348bdbf7ba0cdbe692421457/pkg/bpm/build_linux.go#L36-L45). So the original code already runs through `nsexec`; the activation isn't the missing piece.

**The failure is binary resolution, not namespace entry.** `nsexec` does `setns()` then `exec(cmd)`. Once we've entered the *target* mount namespace, `exec("sh")` / `mkdir` / `cat` resolve against the **target container's rootfs** — which on distroless/scratch images has no `/bin/sh`, so it fails with exit 101.

The only `nsexec` flag that keeps the daemon's own coreutils reachable is `-l` (`EnableLocalMnt`), i.e. *not* entering the target mount namespace. But then `cat > /usr/local/byteman/lib/x.jar` writes into the **daemon's** filesystem, not the container's. To land the bytes inside the container while staying in the daemon's mount ns, you still have to address the path as `/proc/<pid>/root/...` — which is precisely what this PR does, just without forking a shell. So the `nsexec` route either fails on distroless (enter target mnt) or collapses back to `/proc/<pid>/root` anyway (local mnt).

That's why the copy/mkdir steps here go straight through `/proc/<pid>/root` in pure Go ([`mkdirInContainer`](https://github.com/chaos-mesh/chaos-mesh/blob/307cccb1767f5e340569f9deab5d9fe9595f20ea/pkg/chaosdaemon/jvm_server.go#L38-L39), [`copyFileAcrossNS`](https://github.com/chaos-mesh/chaos-mesh/blob/307cccb1767f5e340569f9deab5d9fe9595f20ea/pkg/chaosdaemon/jvm_server.go#L228-L235)). Note the byteman *attach* itself is unchanged and still goes through `bpm` with [`EnableLocalMnt()`](https://github.com/chaos-mesh/chaos-mesh/blob/307cccb1767f5e340569f9deab5d9fe9595f20ea/pkg/chaosdaemon/jvm_server.go#L106-L107) — only the file staging moved off the shell.

**This isn't a one-off.** Since #4985 reports the same root cause for `DNSChaos` (`sh`/`cp`/`sed`/`cat`/`rm` missing on distroless), and its accepted suggested direction — plus the fix in #4986 — both land on the same `/proc/<pid>/root/...` + native-Go pattern. Might be worth aligning both PRs on that pattern so the procfs approach is reviewed once.

**Also note the one last possible failure mode is iochaos. It may also fails on distroless images.**

PR #4980, opened 2026-05-20
fix(controllers): match ChaosImpl pair controllers by base name in enabledControllers

### LINKED ISSUE #4753: [Bug]: It is impossible to specify a ChaosImpl Pair Controller within the enabledController configurations and potentially more controllers.
### Chaos Mesh Version
v2.7.2

### Kubernetes Version
v1.31.2

### Describe the bug
I'm trying to deploy the chaos mesh's controller manager with a subset of controllers, as specified within the values under `controllerManager.resources.enabledControllers`.  The default case is '*' which denotes [deploying the manager with all controllers.](https://github.com/chaos-mesh/chaos-mesh/blob/master/helm/chaos-mesh/values.yaml#L126)

However, I noticed a flaw with how the source handles certain controllers, specifically anything maintained by the common bootstrap function. All controllers generated as a `ChaosImplPair` object seem to never start if explicitly listed in the `enabledControllers` list. I can specify things like `pod<X>chaos`, `workflow`, and `schedule` specific controllers just fine, since these aren't under chaosimpl, contain their own bootstrap methods, and are outside of the common bootstrap, but anything common tries to spin up is skipped.

For example, if I want to add any of the `awschaos` controllers, specifying any of them specifically will not work unless I include '*', which will then start up all controllers.

I believe it has to do with this line:
https://github.com/chaos-mesh/chaos-mesh/blame/43653eb24147a02ac833833259780416e2811891/controllers/common/fx.go#L67

Where the ChaoImplPair's name is appended with the `-records` suffix, making the the `ShouldSpawnController` check impossible to be true unless checking the '*' path.

[This helper function was added here](https://github.com/chaos-mesh/chaos-mesh/commit/01979024e71d0595cb100f44fe42dd8ac04a74f1) and other bootstrap methods may also suffer from this issue.

karimhabush:

Closes #4753                                                                                                        

`controllerManager.enabledControllers` cannot selectively enable ChaosImpl-based controllers (e.g. `awschaos`, `networkchaos`, `podchaos`). The only working value is `"*"`. Specifying any controller by name silently starts zero ChaosImpl controllers.

Three bugs in the same code path:

1. **Name mismatch** — `Bootstrap()` in `controllers/common/fx.go` checked `ShouldSpawnController(pair.Name + "-records")`, but users specify the base name (e.g. `"awschaos"`, not `"awschaos-records"`). Fixed to check `pair.Name` directly.

2. **Early return** — The same function used `return nil` instead of `continue` when a pair didn't match, aborting the entire loop. All subsequent pairs were skipped even if they matched. Fixed to `continue`.

3. **Pipeline steps had the same suffix problem** — `finalizers/step.go`, `desiredphase/step.go`, and `condition/step.go` all checked suffixed names like `"awschaos-initFinalizers"`. Fixed to check `ctx.Object.Name` (the base name).

This makes ChaosImpl controllers consistent with non-ChaosImpl controllers (`podhttpchaos`, `workflow`, `schedule-cron`, etc.) which already check their base name.

### Verified manually                                     

- Deployed with `enabledControllers: [awschaos, podchaos]`
- Before fix: 0 controllers registered
- After fix: only `awschaos` and `podchaos` registered with all pipeline steps                                                          

## Related changes                                                                                                                      

- [x] release-2.8                                        
- [ ] release-2.7

### CHANGELOG

### Tests

- [x] Manual test

### Side effects

### DIFF
--- a/controllers/common/condition/step.go
+++ b/controllers/common/condition/step.go
@@ -24,10 +24,10 @@ import (
 
 func Step(ctx *pipeline.PipelineContext) reconcile.Reconciler {
 	setupLog := ctx.Logger.WithName("setup-condition")
-	name := ctx.Object.Name + "-condition"
-	if !config.ShouldSpawnController(name) {
+	if !config.ShouldSpawnController(ctx.Object.Name) {
 		return nil
 	}
+	name := ctx.Object.Name + "-condition"
 
 	setupLog.Info("setting up controller", "name", name)
 
--- a/controllers/common/desiredphase/step.go
+++ b/controllers/common/desiredphase/step.go
@@ -24,10 +24,10 @@ import (
 
 func Step(ctx *pipeline.PipelineContext) reconcile.Reconciler {
 	setupLog := ctx.Logger.WithName("setup-desiredphase")
-	name := ctx.Object.Name + "-desiredphase"
-	if !config.ShouldSpawnController(name) {
+	if !config.ShouldSpawnController(ctx.Object.Name) {
 		return nil
 	}
+	name := ctx.Object.Name + "-desiredphase"
 
 	setupLog.Info("setting up controller", "name", name)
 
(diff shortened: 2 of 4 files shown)


STRRL:
thank for your contribution! @karimhabush 

We would take a look later 🤔

PR #4986, opened 2026-05-27
fix: support DNSChaos injection on distroless containers

### LINKED ISSUE #4985: [Bug]: DNSChaos injection fails on minimal/distroless containers due to shell dependency inside target MountNS
### Chaos Mesh Version

master

### Kubernetes Version

v1.35.1

### Describe the bug

`DNSChaos` fails to inject against minimal/distroless containers because the injection path depends on shell utilities existing inside the target container filesystem.

The failure occurs after entering the target container mount namespace via:
`SetNS(pid, bpm.MountNS)`

and then executing shell commands such as:
- sh
- cp
- cat
- sed
- rm

inside the target filesystem.

Minimal containers (including pause/distroless-style images) do not contain these binaries, causing injection failure.

### Suggested direction

Instead of executing shell commands inside the target mount namespace, the daemon could manipulate files directly through procfs paths such as:

`/proc/<pid>/root/etc/resolv.conf`

using native Go filesystem operations.

This would:
- remove dependency on target container userspace
- support distroless/minimal containers
- avoid shell execution inside MountNS
- simplify the injection path

#### 1. Start Minikube
minikube start --container-runtime=docker

#### 2. Install Chaos Mesh
```
helm repo add chaos-mesh https://charts.chaos-mesh.org
helm repo update

kubectl create namespace chaos-mesh

helm install chaos-mesh chaos-mesh/chaos-mesh \
  -n chaos-mesh \
  --version 2.8.2
```
Wait until all Chaos Mesh components are healthy:

`kubectl get pods -n chaos-mesh`

#### 3. Create test namespace
`kubectl create namespace test-dns-bug`

#### 4. Create minimal target pod
```
kubectl run distroless-pod \
  --image=registry.k8s.io/pause:3.9 \
  -n test-dns-bug \
  --labels="app=distroless"
```

Wait for readiness:

```
kubectl wait --for=condition=Ready pod/distroless-pod \
  -n test-dns-bug \
  --timeout=120s
```
 
#### 6. Apply DNSChaos
```
kubectl apply -f - <<EOF
kind: DNSChaos
metadata:
  name: distroless-dns-chaos
  namespace: test-dns-bug
spec:
  action: error
  mode: all
  selector:
    namespaces:
      - test-dns-bug
    labelSelectors:
      app: distroless
  duration: "5m"
EOF
```
#### 7. Observe failure
`kubectl describe dnschaos distroless-dns-chaos -n test-dns-bug`
Observed behavior

Injection repeatedly fails with:
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value:
Os { code: 2, kind: NotFound, message: "No such file or directory" }'

The target image does not contain standard shell/coreutils binaries.

## Expected behavior
DNSChaos should work against minimal/distroless containers without requiring:
- /bin/sh
- cp
- cat
- sed
- rm

inside the target container filesystem.

## Root cause
pkg/chaosdaemon/dns_server.go#L57-L60
pkg/chaosdaemon/dns_server.go#L74-L77
pkg/chaosdaemon/dns_server.go#L90-L93

`pkg/chaosdaemon/dns_server.go` enters the target mount namespace and executes shell commands inside the target filesystem.

Because executable path resolution occurs inside the entered mount namespace, containers lacking standard userspace tooling fail during injection.

vky5:

Closes #4985

DNSChaos injection currently fails on minimal/distroless containers because the existing implementation depends on executing shell/coreutils binaries inside the target container mount namespace.

The current flow:
- enters the target mount namespace using `SetNS(pid, bpm.MountNS)`
- executes shell commands through `bpm`
- relies on binaries such as:

```sh
sh
cp
sed
cat
rm
```

inside the target container filesystem.

Minimal/distroless containers do not contain these binaries, causing DNSChaos injection to fail during command execution.

---

This PR removes the dependency on namespace shell execution for DNSChaos injection.

Previously the implementation:
- used `bpm` + `nsexec`
- entered the target mount namespace
- modified `/etc/resolv.conf` through shell commands

The new implementation:
- resolves the target container PID through `s.crClient`
- accesses the target container filesystem directly through:

```text
/proc/<pid>/root/etc/resolv.conf
```

- performs backup/restore and resolv.conf mutation using native Go filesystem APIs instead of shell utilities

This avoids:
- shell execution inside target MountNS
- dependency on container userspace tooling
- failures on distroless/minimal images

The unit tests were also updated from command-construction validation to actual filesystem behavior validation using temporary resolv.conf files.

This follows the direction discussed in #4985 around manipulating the target filesystem through `/proc/<pid>/root`.

---

## Related changes

- [ ] This change also requires further updates to the website (e.g. docs)

- [x] release-2.8
- [ ] release-2.7

### CHANGELOG

### Tests

- [x] Unit test
- [ ] E2E test
- [x] Manual test

### Side effects

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
 
 ### Fixed
 
+- Fixed DNSChaos injection failure on minimal/distroless containers by manipulating `resolv.conf` through `/proc/<pid>/root` instead of shell commands inside the target mount namespace [#4986](https://github.com/chaos-mesh/chaos-mesh/pull/4986)
 - Fixed NetworkChaos recovery failure when target container is in CrashLoopBackOff by falling back to sandbox (pause) container PID for network namespace operations
 - Fixed helm chart template include for extra objects to use the correct render function [#4780](https://github.com/chaos-mesh/chaos-mesh/pull/4780)
 - Fix `install.sh` exiting when kubectl version prints warnings to stderr [#4796](https://github.com/chaos-mesh/chaos-mesh/pull/4796)
--- a/pkg/chaosdaemon/dns_server.go
+++ b/pkg/chaosdaemon/dns_server.go
@@ -19,16 +19,16 @@ import (
 	"context"
 	"fmt"
 	"net"
+	"os"
+	"strings"
 
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/pkg/errors"
 
-	"github.com/chaos-mesh/chaos-mesh/pkg/bpm"
 	pb "github.com/chaos-mesh/chaos-mesh/pkg/chaosdaemon/pb"
-	"github.com/chaos-mesh/chaos-mesh/pkg/chaosdaemon/util"
 )
 
-const (
+var (
 	// DNSServerConfFile is the default config file for DNS server
 	DNSServerConfFile = "/etc/resolv.conf"
 )
@@ -46,6 +46,12 @@ func (s *DaemonServer) SetDNSServer(ctx context.Context,
 		return nil, err
 	}
 
+	targetResolvPath := DNSServerConfFile
+	if req.EnterNS {
+		targetResolvPath = fmt.Sprintf("/proc/%d/root%s", pid, DNSServerConfFile)
+	}
+	backupPath := targetResolvPath + ".chaos.bak"
+
 	if req.Enable {
 		// set dns server to the chaos dns server's address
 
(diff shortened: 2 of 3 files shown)


vky5:
Hey @STRRL 

I noticed the OSV scanner is now failing due to recently disclosed Go stdlib and dependency vulnerabilities (containerd/spdystream).

These failures appear unrelated to the DNSChaos changes in this PR.

Would you prefer that I submit a separate PR to update the affected dependencies and Go versions, or would you like me to include those updates in this PR?

STRRL:
Hi @vky5 it's ok to ignore OSV Scanner right now. <3 We could do it in other PRs.

PR #4999, opened 2026-06-08
fix(ui): show errors without API responses

bvolpato:

Dashboard's global API interceptor silently drops errors without a response body, such as network or CORS failures. Users receive no visible feedback when those requests fail.

Preserve current token and privilege-error handling. For other errors, show the backend message when available and a generic fallback when Axios has no response body.

## Repro

1. Open the dashboard.
2. Trigger an API request that fails before returning a response body.
3. Dashboard now displays `An unknown error occurred` instead of failing silently.

## Related changes

- [x] release-2.8
- [ ] release-2.7

### CHANGELOG

Maintainer action: apply `no-need-update-changelog` (external contributors cannot set upstream labels).

### Tests

- [ ] Unit test
- [ ] E2E test
- [ ] Manual test

Validation:

- `pnpm -F @ui/app exec eslint src/api/interceptors.ts`
- `pnpm exec prettier --check app/src/api/interceptors.ts`
- `pnpm build`
- `pnpm test` (9 tests passed)
- `git diff --check`

### Side effects

### DIFF
--- a/ui/app/src/api/interceptors.ts
+++ b/ui/app/src/api/interceptors.ts
@@ -49,6 +49,11 @@ export const applyErrorHandling = ({
                 type: 'error',
                 message: 'Please check the validity of the token',
               })
+            } else {
+              openAlert({
+                type: 'error',
+                message: data.message || 'An unknown error occurred',
+              })
             }
 
             break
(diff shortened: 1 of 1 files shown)


g1eny0ung (approved):
LGTM! Thanks!

PR #4093, opened 2023-06-23
refactor!: support for deploying chaos-dashboard under the subpath

g1eny0ung:
## What problem does this PR solve?

Close #2697

## What's changed and how it works?

As the title says, this PR supports deploying the dashboard under the subpath (usually with an ingress).

Details: https://github.com/chaos-mesh/chaos-mesh/issues/2697#issuecomment-1604183735.

## Related changes

- [x] This change also requires further updates to the [website](https://github.com/chaos-mesh/website) (e.g. docs)
- [ ] This change also requires further updates to the `UI interface`

## Cherry-pick to release branches (optional)

> This PR should be cherry-picked to the following release branches:

- [x] release-2.6
- [ ] release-2.5

### CHANGELOG

> Must include at least one of them.

- [x] I have updated the `CHANGELOG.md`
- [ ] I have labeled this PR with "no-need-update-changelog"

### Tests

> Must include at least one of them.

- [ ] Unit test
- [ ] E2E test
- [x] Manual test

### Side effects

- [x] **Breaking backward compatibility**

## DCO

If you find the DCO check fails, please run commands like below (Depends on the actual situations. For example, if the failed commit isn't the most recent) to fix it:

```shell
git commit --amend --signoff
git push --force
```

PR #5016, opened 2026-06-25
fix(dashboard): support serving UI under a configurable subpath

### LINKED ISSUE #4968: [Bug]: Dashboard cannot be served from a subpath — Vite asset base is hardcoded to "/"
### Chaos Mesh Version

v2.8.2

### Kubernetes Version

eks version 1.35

### Describe the bug

  **Chaos Mesh version:** v2.8.2
  **Deployment:** Helm chart `chaos-mesh/chaos-mesh` v2.8.2 on Kubernetes 1.35

  ### What happened

  The chaos-dashboard UI cannot be hosted at a non-root path (e.g. `https://example.com/chaos-mesh/`) behind an ingress that prefix-rewrites `/chaos-mesh → /`.

  The initial HTML loads, but the browser then fails to fetch the JS bundles because the served `index.html` references them at `/assets/...` — absolute from root, with no `/chaos-mesh` prefix.

  ### Reproduction

  1. Install chaos-mesh v2.8.2 via the helm chart.
  2. Expose the `chaos-dashboard` Service behind an ingress that matches `/chaos-mesh` and rewrites the path prefix to `/` before forwarding. Example (Contour `HTTPProxy`):

      ```yaml
      spec:
        routes:
        - pathRewritePolicy:
            replacePrefix:
            - prefix: /chaos-mesh
              replacement: /
          services:
          - name: chaos-dashboard
            port: 2333
      ```

  3. Open `https://<host>/chaos-mesh/`.

  ### Expected

  Dashboard UI loads and works under the subpath.

  ### Actual

  - `GET /chaos-mesh/` → 200, returns `index.html`.
  - `index.html` contains `<script type="module" src="/assets/index-<hash>.js"></script>` (absolute path, no `/chaos-mesh` prefix).
  - Browser fires `GET /assets/index-<hash>.js` — that request doesn't match the chaos-mesh ingress route and fails (404 in a plain setup; 302 to the IdP in our auth-gated setup).

  ### Root cause

  The UI is a Vite build with `base: '/'`, which bakes absolute `/assets/...` paths into `index.html` and lazy-load chunks. Those assets are compiled into the `chaos-dashboard` binary via `embed.FS`, so no runtime env var or helm value can change
   them.

  Verified that none of the existing knobs help:

  | Knob | Effect on asset paths |
  |---|---|
  | `dashboard.rootUrl` / `ROOT_URL` env | None — used only for OAuth/OIDC callback URL construction |
  | `dashboard.env.*` | None — no env var (`UI_SERVE_BASE_PATH`, `PUBLIC_PATH`, etc.) is read by the binary for this purpose |
  | `dashboard.ingress.paths` | None — only affects the chart-generated Ingress rule, not the HTML payload |

  ### Suggested fix

  Make the UI base path configurable. Two options, in order of friendliness to users:

  1. **Runtime rewrite (preferred).** Emit `index.html` with a placeholder (e.g. `<base href="__UI_BASE_PATH__">` plus `base: '__UI_BASE_PATH__'` in the bundle) and have the Go server substitute it on request from an env var like `UI_BASE_PATH`.
  Grafana uses this pattern. The helm chart can then expose `dashboard.uiBasePath` as a first-class value, and users get subpath hosting without rebuilding the image.

  2. **Build-time arg.** Have `vite.config.ts` read `process.env.PUBLIC_PATH` (or similar) and pass it to `base`. Expose via a Dockerfile/Makefile build arg. Downstream users wanting a subpath still need to build their own image, but at least
  it's a supported path.

### To Reproduce

_No response_

victoriacheng15:

Fixes #4968.

The chaos-dashboard UI cannot be hosted at a non-root path (e.g. `https://example.com/chaos-mesh/`) behind an Ingress/Proxy. Absolute `/assets/...` URLs hardcoded in `index.html` resolve against the domain root instead of the subpath, causing 404 resource errors.

1. **Relocatable UI Bundle**:
   - In `ui/app/vite.config.ts`, set production `base` to `./` so Vite emits relative asset paths (`./assets/...`).
   - In `ui/app/index.html`, updated the favicon link to use `%BASE_URL%favicon.ico` for dynamic evaluation.
2. **Go Backend Path Prefix Stripping**:
   - Added `UI_BASE_PATH` environment variable configuration (`UIBasePath`) to `pkg/config/dashboard.go`.
   - Registered a Gin middleware in `pkg/dashboard/apiserver/server.go` to strip the configured `UI_BASE_PATH` prefix from incoming request paths (supporting passthrough ingress configurations).
3. **Helm Charts Wiring**:
   - Added `dashboard.uiBasePath` value to `helm/chaos-mesh/values.yaml`.
   - Configured `UI_BASE_PATH` env var in `helm/chaos-mesh/templates/chaos-dashboard-deployment.yaml`.

## Related changes

- [ ] This change also requires further updates to the website https://github.com/chaos-mesh/website (e.g. docs)

This PR should be cherry-picked to the following release branches:

- [x] release-2.8
- [ ] release-2.7

### CHANGELOG

- [x] I have updated the CHANGELOG.md

### Tests

- [x] Unit test
- [ ] E2E test
- [x] Manual test

### Side effects

### DIFF
--- a/pkg/dashboard/apiserver/server.go
+++ b/pkg/dashboard/apiserver/server.go
@@ -55,24 +55,27 @@ var (
 func register(r *gin.Engine, conf *config.ChaosDashboardConfig) {
 	listenAddr := net.JoinHostPort(conf.ListenHost, fmt.Sprintf("%d", conf.ListenPort))
 
-	go r.Run(listenAddr)
+	basePath := normalizeBasePath(conf.UIBasePath)
+	var handler http.Handler = r
+	if basePath != "" {
+		handler = http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			path := req.URL.Path
+			if path == basePath {
+				http.Redirect(w, req, basePath+"/", http.StatusMovedPermanently)
+				return
+			}
+			if strings.HasPrefix(path, basePath+"/") {
+				req.URL.Path = strings.TrimPrefix(path, basePath)
+			}
+			r.ServeHTTP(w, req)
+		})
+	}
+
+	go http.ListenAndServe(listenAddr, handler)
 }
 
 func newEngine(config *config.ChaosDashboardConfig) *gin.Engine {
 	r := gin.Default()
-
-	basePath := normalizeBasePath(config.UIBasePath)
-	if basePath != "" {
-		r.Use(func(c *gin.Context) {
-			path := c.Request.URL.Path
-			if path == basePath {
-				c.Request.URL.Path = "/"
-			} else if strings.HasPrefix(path, basePath+"/") {
-				c.Request.URL.Path = strings.TrimPrefix(path, basePath)
-			}
-			c.Next()
-		})
-	}
 
 	if config.EnableProfiling {
 		// default is "/debug/pprof"
--- a/pkg/dashboard/apiserver/server.go
+++ b/pkg/dashboard/apiserver/server.go
@@ -19,7 +19,6 @@ import (
 	"fmt"
 	"net"
 	"net/http"
-	"strings"
 
 	"github.com/gin-contrib/pprof"
 	"github.com/gin-gonic/gin"
@@ -55,23 +54,7 @@ var (
 func register(r *gin.Engine, conf *config.ChaosDashboardConfig) {
 	listenAddr := net.JoinHostPort(conf.ListenHost, fmt.Sprintf("%d", conf.ListenPort))
 
-	basePath := normalizeBasePath(conf.UIBasePath)
-	var handler http.Handler = r
-	if basePath != "" {
-		handler = http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
-			path := req.URL.Path
-			if path == basePath {
-				http.Redirect(w, req, basePath+"/", http.StatusMovedPermanently)
-				return
-			}
-			if strings.HasPrefix(path, basePath+"/") {
-				req.URL.Path = strings.TrimPrefix(path, basePath)
-			}
-			r.ServeHTTP(w, req)
-		})
-	}
-
-	go http.ListenAndServe(listenAddr, handler)
+	go r.Run(listenAddr)
 }
 
 func newEngine(config *config.ChaosDashboardConfig) *gin.Engine {
@@ -108,18 +91,6 @@ func newEngine(config *config.ChaosDashboardConfig) *gin.Engine {
 	}
 
 	return r
-}
-
-func normalizeBasePath(basePath string) string {
-	basePath = strings.TrimSpace(basePath)
-	if basePath == "" || basePath == "/" {
-		return ""
-	}
-	if !strings.HasPrefix(basePath, "/") {
-		basePath = "/" + basePath
-	}
-	basePath = strings.TrimSuffix(basePath, "/")
-	return basePath
 }
 
 func newAPIRouter(r *gin.Engine) *gin.RouterGroup {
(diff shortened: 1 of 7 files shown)


pkg/dashboard/apiserver/server.go
```diff
@@ -54,7 +55,23 @@ var (
 func register(r *gin.Engine, conf *config.ChaosDashboardConfig) {
 	listenAddr := net.JoinHostPort(conf.ListenHost, fmt.Sprintf("%d", conf.ListenPort))
 
-	go r.Run(listenAddr)
+	basePath := normalizeBasePath(conf.UIBasePath)
```
g1eny0ung: I don't think we need to add a handler to handle the base path. In #4093, I tried to solve the same problem. At the time, the UI was built using `create-react-app` (which is now deprecated), and I wrote an Ingress example to handle API requests. You can also view the example at https://chaos-mesh.org/docs/expose-dashboard-with-ingress/.

The annotation `nginx.ingress.kubernetes.io/rewrite-target: /$1` will forward the prefix base path to the capture group, so it performs the same task as this part of code.

I will continue testing to see if the previous Ingress example still works correctly.

ui/app/vite.config.ts
```diff
@@ -4,10 +4,11 @@ import { defineConfig, loadEnv } from 'vite'
 import svgr from 'vite-plugin-svgr'
 
 // https://vite.dev/config/
-export default defineConfig(({ mode }) => {
+export default defineConfig(({ mode, command }) => {
   const env = loadEnv(mode, process.cwd(), '')
 
   return {
+    base: command === 'build' ? './' : '/',
```
g1eny0ung: We can use `base: ./` in both the `dev` and `prod` build processes. There is no need to make a distinction here.

victoriacheng15:
@g1eny0ung 

I have updated the PR based on your feedback:

1. **Vite Configuration**: Simplified `vite.config.ts` to use `base: './'` unconditionally for both dev and prod builds.
2. **Backend & Helm Reverted**: Reverted all changes in `pkg/` and `helm/` since the Ingress `rewrite-target` annotation already handles path-prefix stripping at the infrastructure layer.

### Local Verification Steps

I built and verified the changes in a local dev environment:

1. **Build UI and embed assets**:
   ```bash
   cd ui && pnpm install --frozen-lockfile && UI=1 pnpm build && cd ..
   ./hack/embed_ui_assets.sh
   UI=1 make local/chaos-dashboard
   ```

2. **Verify relative asset paths in built HTML**:
   ```bash
   cat ui/app/dist/index.html | grep -oE '(src|href)="[^"]+"'
   ```
   * **Output**:
     ```html
     href="./favicon.ico"
     href="https://fonts.googleapis.com"
     href="https://fonts.gstatic.com"
     href="https://fonts.googleapis.com/css2?family=Roboto:wght@300;400;500;700&display=swap"
     src="./assets/index-C4j8tj7c.js"
     ```
     *(All local resources correctly resolve to `./` relative paths).*

3. **Verify API and UI serving**:
   Started the dashboard binary `./bin/chaos-dashboard` and tested simulated Ingress routing:
   * **API Config Request**:
     ```bash
     curl -i http://localhost:2333/api/common/config
     ```
     Returns `200 OK` (the Go backend serves API routes correctly without requiring prefix-stripping middleware).
   * **UI Request**:
     ```bash
     curl -i http://localhost:2333/
     ```
     Returns `200 OK` and serves the index HTML page containing the relative asset links.

g1eny0ung (approved):
LGTM!

Issue #5014, opened 2026-06-23
Change container base image debian -> alpine


STRRL:

codic-yeeshu:
Hey @STRRL , just needed one clarification, this issue will involve all those Dockerfiles which are having their base images as debian, right?

STRRL:
> Hey [@STRRL](https://github.com/STRRL) , just needed one clarification, this issue will involve all those Dockerfiles which are having their base images as debian, right?

Yes you're right

saiyam1814:
I'd like to take a first slice of this: migrating chaos-dashboard and chaos-dlv to alpine (both are pure-Go / CGO-free, verified working locally). chaos-daemon needs musl builds of the glibc-linked vendor binaries (toda, nsexec, chaos-tproxy, memStress) plus an alpine OpenJDK, so I'd propose handling it in a follow-up. PR incoming.

codic-yeeshu:
> > Hey [@STRRL](https://github.com/STRRL) , just needed one clarification, this issue will involve all those Dockerfiles which are having their base images as debian, right?
> 
> Yes you're right

got it, are there any version requirements or latest stable release will do the job?

STRRL:
latest stable version will be good 👍

PR #4800, opened 2025-12-11
chore(dashboard): remove CGO support for chaos-dashboard

spencercjh:
**Parts of this PR were drafted with assistance from Warp (with `claude 4.5 haiku&sonnet` and `gpt-5(medium reasoning)`) and fully reviewed and edited by me. I take full responsibility for all changes.**

## What problem does this PR solve?

This PR removes the CGO dependency from chaos-dashboard to simplify the build process and improve cross-compilation support.

**Problem**: Currently, chaos-dashboard requires CGO because it uses `github.com/mattn/go-sqlite3`, a CGO-based SQLite driver. This introduces:
- Build complexity (requires C toolchain)
- Cross-compilation challenges
- Longer build times
- Reduced portability

Part of #4799  

### Changes

1. **Replaced SQLite driver**:
   - Removed: `github.com/mattn/go-sqlite3` (CGO-based)
   - Added: `github.com/glebarez/sqlite` (pure Go, gorm v1-compatible)

2. **Updated build configuration**:
   - Modified `cmd/generate-makefile/generate_binary_makefile.go` to set `UseCGO: false` for chaos-dashboard
   - Regenerated `binary.generated.mk` using `go run ./cmd/generate-makefile`
   - Build command now uses `$(GO)` instead of `$(CGO)`

3. **Updated import**:
   - Changed `cmd/chaos-dashboard/main.go` from `github.com/jinzhu/gorm/dialects/sqlite` to `github.com/glebarez/sqlite`

### How it works

The new driver (`glebarez/sqlite`) is a pure Go SQLite implementation based on `modernc.org/sqlite`. It provides a drop-in replacement for `mattn/go-sqlite3` with full gorm v1 compatibility:
- No API changes required
- Existing databases work without migration
- All GORM features remain supported

### Files changed
- `cmd/chaos-dashboard/main.go` (1 line)
- `cmd/generate-makefile/generate_binary_makefile.go` (1 line)
- `binary.generated.mk` (generated, 1 line)
- `go.mod` and `go.sum` (dependency updates)

### Benefits
- ✅ No C toolchain required for dashboard builds
- ✅ Easier cross-compilation (amd64, arm64)
- ✅ Faster build times (no CGO overhead)
- ✅ Better portability (pure Go binary)

## Related changes

- [ ] This change also requires further updates to the [website](https://github.com/chaos-mesh/website) (e.g. docs)
- [ ] This change also requires further updates to the `UI interface`

## Cherry-pick to release branches (optional)

[...]

PR #5027, opened 2026-07-07
build(images): migrate chaos-dashboard and chaos-dlv base images from debian to alpine

saiyam1814:

Part of #5014 (debian → alpine base image migration).

This is an incremental first slice covering the two images that need no binary changes.

- **chaos-dashboard**: the binary is built with `CGO_ENABLED=0` since #4800, so it runs on musl without changes. Replaced `debian:bookworm-slim` + apt with `alpine:3.23` + apk (`tzdata`, `ca-certificates`), matching the existing chaos-controller-manager image.
- **chaos-dlv**: dlv is pure Go; build with `golang:1.25-alpine`, run on `alpine:3.23` with `procps`. Switched `CMD` from `bash` to `sh` since alpine ships no bash.

**Intentionally out of scope** (follow-up needed): chaos-daemon (CGO-linked against glibc and bundles glibc-linked vendor binaries — toda, nsexec, chaos-tproxy, memStress — plus openjdk-17 and iptables-legacy alternatives), build-env, dev-env.

## Related changes

### CHANGELOG

- [x] I have updated the [CHANGELOG.md](cci:7://file:///Users/saiyam/git/chaos-mesh/CHANGELOG.md:0:0-0:0)

### Tests

- [x] Manual test — both images build locally; `chaos-dashboard --version` and `dlv version` run correctly on alpine; `ps axf` works with procps.

### Side effects

- No breaking backward compatibility. Helm charts reference `/usr/local/bin/chaos-dashboard` and the dlv `CMD_NAME` wiring, both unchanged.

### DIFF
--- a/images/chaos-dashboard/Dockerfile
+++ b/images/chaos-dashboard/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:bookworm-slim
+FROM alpine:3.23
 
 ARG HTTPS_PROXY
 ARG HTTP_PROXY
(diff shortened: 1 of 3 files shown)


images/chaos-dashboard/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM debian:bookworm-slim
+FROM alpine:3.23
```
STRRL, line 1: just a quick question, should we use `3` or `3.23`? should we pin the minor verison?

saiyam1814: I pinned `3.23` to match the existing chaos-controller-manager image (`images/chaos-mesh/Dockerfile` already uses `alpine:3.23`), and for reproducible builds — `alpine:3` silently moves to new minor releases. Happy to change both to `3` if you'd prefer tracking the latest stable.

PR #5035, opened 2026-07-17
fix(chaosdaemon): use SourcePort instead of EgressPort in tc filter

magic-peach:
## What this PR does

Fixes a copy-paste bug in `abstractTcFilter` in `pkg/chaosdaemon/tc_server.go`.

## Changes

- **Line 515**: Changed `tc.EgressPort` to `tc.SourcePort` in the SourcePort branch

## Why

In the `abstractTcFilter` function, the SourcePort branch was incorrectly appending `EgressPort` to the filter string instead of `SourcePort`. This caused the wrong port identifier to be used in tc filter rules when `SourcePort` is set, leading to incorrect traffic classification.

```go
// Before (bug):
if len(tc.SourcePort) > 0 {
    filter += "-" + tc.EgressPort  // wrong: uses EgressPort
}

// After (fix):
if len(tc.SourcePort) > 0 {
    filter += "-" + tc.SourcePort  // correct: uses SourcePort
}
```

This is a straightforward copy-paste error — the `SourcePort` branch duplicates the logic of the `EgressPort` branch but should reference `SourcePort`.

### DIFF
--- a/pkg/chaosdaemon/tc_server.go
+++ b/pkg/chaosdaemon/tc_server.go
@@ -512,7 +512,7 @@ func abstractTcFilter(tc *pb.Tc) string {
 	}
 
 	if len(tc.SourcePort) > 0 {
-		filter += "-" + tc.EgressPort
+		filter += "-" + tc.SourcePort
 	}
 
 	return filter


STRRL:
Thanks for your contribution! :heart:

This is a duplicate of #4897, and after looking into it we found that `source_port` and `egress_port` on the `Tc` message are never set by the Kubernetes controllers, so the affected branch in `abstractTcFilter` is dead code on the Kubernetes path. The actual port filtering happens through the iptables `Chain` message, which uses the correct fields.

Instead of fixing the copy-paste line, we decided to remove these unused fields entirely in #5056. Closing this one, but thanks for spotting the bug!

PR #5036, opened 2026-07-17
fix(chaosdaemon): handle ignored errors and resource leaks

magic-peach:
## Summary

Fixes three resource leak and error handling issues in the chaos-daemon:

### 1. Unchecked `ensureRule` error in iptables initialization (`iptables_server.go:184`)

In `initializeEnv()`, the return value of `ensureRule()` was discarded. If adding the jump rule (`-A INPUT -j CHAOS-INPUT`) failed, the error was silently ignored and the function returned success, leaving the iptables environment in a partially initialized state.

### 2. HTTP response body leak (`httpchaos_server.go:155`)

After `transport.RoundTrip(req)`, the `resp.Body` was read but never closed. This leaks the underlying TCP connection on every `ApplyHttpChaos` call. Added `defer resp.Body.Close()` immediately after the error check.

### 3. Ignored `AppendCertsFromPEM` return values (`grpc/utils.go:79,96`)

Both `FileProvider.getCredentialOption()` and `RawProvider.getCredentialOption()` called `AppendCertsFromPEM()` without checking the boolean return. If the PEM data was malformed or empty, the cert pool would be empty, leading to TLS connections that silently fail to verify the server certificate. Both call sites now return an error if no certificates were appended.

### DIFF
--- a/pkg/grpc/utils.go
+++ b/pkg/grpc/utils.go
@@ -76,7 +76,9 @@ func (it *FileProvider) getCredentialOption() (grpc.DialOption, error) {
 		return nil, err
 	}
 	caCertPool := x509.NewCertPool()
-	caCertPool.AppendCertsFromPEM(caCert)
+	if !caCertPool.AppendCertsFromPEM(caCert) {
+		return nil, errors.New("failed to append CA certificate to pool")
+	}
 
 	clientCert, err := tls.LoadX509KeyPair(it.file.Cert, it.file.Key)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/grpc/utils.go
```diff
@@ -93,7 +95,9 @@ func (it *FileProvider) getCredentialOption() (grpc.DialOption, error) {
 
 func (it *RawProvider) getCredentialOption() (grpc.DialOption, error) {
 	caCertPool := x509.NewCertPool()
-	caCertPool.AppendCertsFromPEM(it.raw.CaCert)
+	if !caCertPool.AppendCertsFromPEM(it.raw.CaCert) {
+		return nil, errors.New("failed to append CA certificate to pool")
```
STRRL, line 99: please also attach the ca file path into the error message!

rest LGTM!

Thanks!

PR #5039, opened 2026-07-19
fix: handle non-200 HTTP status in chart download and return error in stresschaos recovery

magic-peach:
## Summary
- Fix silent nil error when Helm chart download returns non-200 HTTP status
- Return error in stresschaos CancelStressors to enable controller retry

## Changes
- `pkg/helm/chart.go`: Use `errors.Errorf` instead of `errors.Wrapf(err, ...)` when status code is not 200, since `err` is nil in this case (the HTTP request succeeded, just returned a bad status)
- `controllers/chaosimpl/stresschaos/impl.go`: Return the error from `CancelStressors` instead of nil, allowing the controller to retry the recovery operation

## Impact
Without the chart fix, a failed download returns `("", nil)` causing confusing downstream errors. Without the stresschaos fix, stress injection can become permanently stuck on a pod if CancelStressors fails.

### DIFF
--- a/controllers/chaosimpl/stresschaos/impl.go
+++ b/controllers/chaosimpl/stresschaos/impl.go
@@ -139,7 +139,7 @@ func (impl *Impl) Recover(ctx context.Context, index int, records []*v1alpha1.Re
 	}
 	if _, err = pbClient.CancelStressors(ctx, req); err != nil {
 		impl.Log.Error(err, "cancel stressors")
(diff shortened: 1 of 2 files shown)


STRRL (commented):
Basically LGTM! Thanks for your contribution! @magic-peach 

and please update CHANGELOG.md;

and it's better to have 2 PRs for different fixes : )

PR #5040, opened 2026-07-19
fix(chaos-builder): close generated files to prevent handle leaks

magic-peach:
## Summary
- Close generated files in chaos-builder to prevent file handle leaks

## Bug
In `cmd/chaos-builder/main.go`, six files are created with `os.Create` and written with `fmt.Fprint`, but none are ever closed. The `file` variable is reassigned each time, losing the previous handle. This leaks file descriptors and may cause data loss if buffers aren't flushed before process termination.

## Fix
Add explicit `file.Close()` after each `fmt.Fprint` call, before reassigning the variable.

### DIFF
--- a/cmd/chaos-builder/main.go
+++ b/cmd/chaos-builder/main.go
@@ -154,6 +154,7 @@ func init() {
 		os.Exit(1)
 	}
 	fmt.Fprint(file, implCode)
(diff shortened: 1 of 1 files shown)


STRRL (approved):
LGTM! Thanks!

PR #5068, opened 2026-08-18
feat(networkchaos): support fail-fast partition rejection

### LINKED ISSUE #5067: Support fail-fast TCP rejection for NetworkChaos partitions
## Feature request

`NetworkChaos` partitions currently implement only blackhole behavior: the controller manager always sends `DROP` as the iptables target. This is useful for simulating packet loss and long network timeouts, but it cannot model fail-fast failures where clients immediately learn that a TCP connection is unavailable.

The Chaos Daemon API already supports the primitives needed for this. Its iptables `Chain` message accepts both a target and protocol, and can therefore install `REJECT --reject-with tcp-reset` for TCP traffic. The controller manager currently hard-codes `DROP` and does not expose a safe way to select the alternative.

## Proposed API

Add an optional partition-specific behavior to `NetworkChaosSpec`:

```yaml
spec:
  action: partition
  partitionBehavior: reject
```

Supported values would be:

- `drop` (default): preserve the current blackhole behavior.
- `reject`: fail TCP connections immediately with a TCP reset.

The controller would propagate the selected behavior through `PodNetworkChaos` and map it to the existing daemon `Chain` fields. This keeps the daemon protocol unchanged and preserves backward compatibility for all existing manifests.

## Motivation

Blackholes and immediate connection failures exercise materially different timeout, retry, failover, and connection-pool paths. Supporting both lets users reproduce these failure modes deterministically without restarting a database, proxy, or other remote service.

anton-oai:

NetworkChaos partitions currently model only a blackhole: the controller manager always sends `DROP` as the iptables target. That behavior is useful for exercising timeouts, but it cannot model fail-fast TCP failures where a client immediately learns that a destination is unavailable.

The Chaos Daemon protocol already exposes the necessary `target` and `protocol` fields. The missing piece is a safe, typed option in the NetworkChaos API and controller-manager propagation.

Fixes #5067.

This PR adds an optional `spec.partitionBehavior` field for `action: partition`:

- `drop` is the default and preserves the existing blackhole behavior.
- `reject` installs `REJECT --reject-with tcp-reset` for TCP traffic so affected connections fail immediately.

The selected behavior is propagated through the generated PodNetworkChaos rules. The pod network controller then maps the typed value to the existing Chaos Daemon `Chain.target` and `Chain.protocol` fields, so no daemon protocol or daemon implementation change is required.

The change includes:

- API defaulting and CRD validation for `drop` and `reject`;
- controller-manager propagation and strict handling of unknown values;
- unit coverage for default, explicit drop, reject, and invalid inputs;
- regenerated CRDs, Helm CRDs, workflow schemas, and dashboard Swagger output;
- a runnable `network-partition-reject` example; and
- an Unreleased changelog entry.

Existing manifests remain backward compatible and continue to use `DROP`.

## Related changes

The API field documentation, generated schemas, and example manifest are included in this repository. A dedicated Dashboard form control can be added separately; YAML-created experiments work with this change as-is.

- [ ] release-2.8
- [ ] release-2.7

### CHANGELOG

### Tests

- [x] Unit test
- [ ] E2E test
- [x] Manual test

Validation performed:

```text
go test ./controllers/podnetworkchaos -run '^TestBuildIptablesChain$'
go test ./api/v1alpha1 -run '^TestAPIs$'  # in the upstream dev container
make generate
git diff --check
```

The fail-fast behavior was also exercised against a live TCP connection using a controller-manager image built from the equivalent downstream patch.

### Side effects

## DCO

The commit is signed off for DCO compliance.


saiyam1814 (approved):
Reviewed the current 32d52030 diff and ran the focused chain and daemon rule tests. Existing manifests default to drop. Reject mode emits a TCP reset rule plus an unconstrained REJECT fallback for non-TCP traffic, and fallback_target is an additive protobuf field at 10. Direction, device, and ipset matching remain intact. The scalar fields are preserved by the generated DeepCopy methods through the initial *out = *in copy. LGTM.

anton-oai:
@saiyam1814 is there anything I need to do to kick off ci tasks?

anton-oai:
@saiyam1814 I've addressed test failures, can you please trigger CI again. Thanks!

STRRL (approved):

PR #5070, opened 2026-08-18
ci(license-checker): filter ignored files from changed-files

saiyam1814:
## Summary
- filter generated and non-source additions from `tj-actions/changed-files` outputs
- keep the action ignore patterns identical to the existing pull request trigger ignores
- quote file paths passed to `grep`

This fixes mixed pull requests where a source change starts the workflow and generated CRDs, Swagger output, or documentation are then incorrectly checked for source-code license headers.

#### Test plan
- [x] `actionlint .github/workflows/license_checker.yml`
- [x] Parse the workflow as YAML
- [x] Verify all 47 trigger ignore patterns match the action ignore patterns
- [x] `git diff --check`

Generated with [Devin](https://devin.ai)

### DIFF
--- a/.github/workflows/license_checker.yml
+++ b/.github/workflows/license_checker.yml
@@ -1,3 +1,8 @@
+# Note: If you change `paths-ignore` in this workflow,
+# please also update the `files_ignore` in the `Get changed files` step accordingly.
+#
+# Because the `paths-ignore` is used to determine whether to trigger the workflow or not,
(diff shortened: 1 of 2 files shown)


g1eny0ung (approved):
Yes, we do need to filter again from the changed-files because paths-ignore will only be triggered when all files are hit. Additionally, I added an extra note to mention that both file lists need to be updated simultaneously.

PR #5074, opened 2026-08-18
fix(chaosdaemon): convert netem delay/jitter as durations instead of strings

### LINKED ISSUE #5073: [Bug]:  NetworkChaos delay is silently dropped for latency values such as `0.5s`
### Chaos Mesh Version

v2.8.3

### Kubernetes Version

1.36.1

### Describe the bug

`convertNetemToArgs` compares the delay and jitter of a Netem against the string `"0ms"`:

https://github.com/chaos-mesh/chaos-mesh/blob/608b61e7/pkg/chaosdaemon/tc_server.go#L426

```go
if netem.Time > "0ms" {
    args = fmt.Sprintf("delay %s", netem.Time)
    if netem.Jitter > "0ms" {
```

These fields hold strings since PR 4546 changed them from integers, so this is a lexicographic comparison, not a duration comparison, and the raw value is then forwarded to `tc` unchanged.

`"0.5s" > "0ms"` is `false`, because `.` sorts before `m`. So for `latency: 0.5s` no `delay` argument is produced at all, even though the value passes webhook validation (`webhook:"Duration"` only calls `time.ParseDuration`). The same holds for `0.25s`, `.5s` and `0m30s`.

What happens next depends on whether the experiment sets any other netem parameter.

**1. Delay only: the experiment can never be applied.** With an empty argument list, `addNetem` builds a command line ending in a space, and `strings.Split(args, " ")` turns that trailing space into an empty argv element:

```go
args := fmt.Sprintf("qdisc add dev %s %s %s netem %s", device, parent, handle, netemArgs)
processBuilder := bpm.DefaultProcessBuilder("tc", strings.Split(args, " ")...)
// argv = ["qdisc","add","dev","eth0","root","handle","1:","netem",""]
```

`tc` rejects the empty argument:

```
$ tc qdisc add dev lo root handle 1: netem ""
What is ""?
Usage: ... netem [ limit PACKETS ]
                 [ delay TIME [ JITTER [ CORRELATION ] ] ]
...
$ echo $?
1
```

which surfaces as an endlessly retried

```
failed to apply for pod default/net-client: unable to set tcs for pod net-client
```

The same command without the trailing empty argument is accepted, so the empty argv element is what turns a missing `delay` into a hard failure.

**2. Delay combined with loss/corrupt/duplicate/rate: the delay is silently dropped.** The argument list is not empty, so no error is raised. The experiment is reported as injected and the other faults do take effect, but there is no latency at all.

Two more consequences of the same comparison:

**3. Zero latencies are not dropped.** `"0s"`, `"0us"` and `"0ns"` sort *after* `"0ms"` and generate a pointless `delay 0s`.

**4. Units valid for the API are rejected by tc.** The API accepts anything `time.ParseDuration` accepts (`ns`, `us`, `ms`, `s`, `m`, `h`, and compound values), while `tc` only knows seconds, milliseconds and microseconds. `latency: 1m`, `2h`, `1500ns` or `0m30s` reach `tc` verbatim and fail with `Illegal "latency"`.

```bash
for t in 1s 1sec 1secs 1S 1Sec 100ms 100msec 100msecs 100MS 500us 500usec 500usecs 500US 1000 0.5s 1.5ms 1m 1h 1ns 1min 1000ms.  ; do
    out=$(tc qdisc add dev lo root netem delay "$t" 2>&1 | head -1)
    case "$out" in
      RTNETLINK*) echo "  ACCEPT   $t" ;;
      *)          echo "  REJECT   $t   ($out)" ;;
    esac
  done
  ACCEPT   1s
  ACCEPT   1sec
  ACCEPT   1secs
  ACCEPT   1S
  ACCEPT   1Sec
  ACCEPT   100ms
  ACCEPT   100msec
  ACCEPT   100msecs
  ACCEPT   100MS
  ACCEPT   500us
  ACCEPT   500usec
  ACCEPT   500usecs
  ACCEPT   500US
  ACCEPT   1000
  ACCEPT   0.5s
  ACCEPT   1.5ms
  REJECT   1m   (Illegal "latency")
  REJECT   1h   (Illegal "latency")
  REJECT   1ns   (Illegal "latency")
  REJECT   1min   (Illegal "latency")
  REJECT   1000ms.   (Illegal "latency")
```
The existing e2e specs only use `200ms`, which happens to compare correctly, so this was not caught.

### To Reproduce

1.  Create 4 pods. Using the nginx:stable image is ok.

```bash
NAME           READY   STATUS    RESTARTS      AGE
net-client     1/1     Running   1 (7h ago)    16h
net-client-b   1/1     Running   6 (18m ago)   16h
net-server     1/1     Running   1 (7h ago)    16h
net-target     1/1     Running   6 (18m ago)   16h
```

2. Enter the `net-client` and use `curl` to access the net-server pod

```bash
root@net-client:/# SERVER_IP=10.244.0.14
root@net-client:/# curl -s -o /dev/null -w "HTTP Code: %{http_code} | Total Time: %{time_total}s\n" http://$SERVER_IP
HTTP Code: 200 | Total Time: 0.002894s
```
3. Apply the following YAML

```yaml
kind: NetworkChaos
metadata:
  name: test-http-delay
  namespace: chaos-mesh
spec:
  action: delay
  mode: one
  selector:
    namespaces:
      - default
    labelSelectors:
      run: 'net-server'
  target:
    mode: one
    selector:
      namespaces:
        - default
      labelSelectors:
        run: 'net-client'
  delay:
    latency: '1s'
  direction: from
```

4. curl again

```bash
root@net-client:/# curl -s -o /dev/null -w "HTTP Code: %{http_code} | Total Time: %{time_total}s\n" http://$SERVER_IP
HTTP Code: 200 | Total Time: 1.995370s
```

5. Change the delay to 0.5s

```bash
kubectl -nchaos-mesh delete networkchaos test-http-delay
# change delay to 0.5s and apply again
```

6. crul again

```bash
root@net-client:/# curl -s -o /dev/null -w "HTTP Code: %{http_code} | Total Time: %{time_total}s\n" http://$SERVER_IP
HTTP Code: 200 | Total Time: 0.000455s
```

7. The following example also has the same problem.

100ms + 20ms is ok
0.1s + 0.02 is failed

```yaml
kind: NetworkChaos
metadata:
  name: test-ping-delay
  namespace: chaos-mesh
spec:
  action: delay
  mode: one
  selector:
    namespaces:
      - default
    labelSelectors:
      'app': 'net-target'
  delay:
    latency: '0.1s' # 100ms
    jitter: '0.02s' # 20ms 
  direction: to
```

```bash
huang@ubuntu:~$ kubectl exec -it net-client-b -- bash
root@net-client-b:/# ping -h
bash: ping: command not found
root@net-client-b:/# SERVER_IP=10.244.0.7
root@net-client-b:/# ping -c 5 $TARGET_IP
ping: usage error: Destination address required
root@net-client-b:/# ping -c 5 $SERVER_IP
PING 10.244.0.7 (10.244.0.7) 56(84) bytes of data.
64 bytes from 10.244.0.7: icmp_seq=1 ttl=63 time=0.054 ms
64 bytes from 10.244.0.7: icmp_seq=2 ttl=63 time=0.046 ms
64 bytes from 10.244.0.7: icmp_seq=3 ttl=63 time=0.044 ms
64 bytes from 10.244.0.7: icmp_seq=4 ttl=63 time=0.048 ms
64 bytes from 10.244.0.7: icmp_seq=5 ttl=63 time=0.047 ms

--- 10.244.0.7 ping statistics ---
5 packets transmitted, 5 received, 0% packet loss, time 4088ms
rtt min/avg/max/mdev = 0.044/0.047/0.054/0.003 ms
root@net-client-b:/# ping -c 5 $SERVER_IP
PING 10.244.0.7 (10.244.0.7) 56(84) bytes of data.
64 bytes from 10.244.0.7: icmp_seq=1 ttl=63 time=103 ms
64 bytes from 10.244.0.7: icmp_seq=2 ttl=63 time=107 ms
64 bytes from 10.244.0.7: icmp_seq=3 ttl=63 time=82.8 ms
64 bytes from 10.244.0.7: icmp_seq=4 ttl=63 time=97.6 ms
64 bytes from 10.244.0.7: icmp_seq=5 ttl=63 time=90.7 ms

--- 10.244.0.7 ping statistics ---
5 packets transmitted, 5 received, 0% packet loss, time 3996ms
rtt min/avg/max/mdev = 82.814/96.194/106.653/8.603 ms
root@net-client-b:/# ping -c 5 $SERVER_IP
PING 10.244.0.7 (10.244.0.7) 56(84) bytes of data.
64 bytes from 10.244.0.7: icmp_seq=1 ttl=63 time=0.048 ms
64 bytes from 10.244.0.7: icmp_seq=2 ttl=63 time=0.067 ms
64 bytes from 10.244.0.7: icmp_seq=3 ttl=63 time=0.078 ms
64 bytes from 10.244.0.7: icmp_seq=4 ttl=63 time=0.058 ms
64 bytes from 10.244.0.7: icmp_seq=5 ttl=63 time=0.035 ms

--- 10.244.0.7 ping statistics ---
5 packets transmitted, 5 received, 0% packet loss, time 4117ms
rtt min/avg/max/mdev = 0.035/0.057/0.078/0.014 ms
````

### PR DESCRIPTION
**Pull request by 0xff-dev**
Fix https://github.com/chaos-mesh/chaos-mesh/issues/5073

https://github.com/chaos-mesh/chaos-mesh/issues/5073

## Related changes

- [x] release-2.8
- [x] release-2.7

### CHANGELOG

### Tests

- [x] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

> [!TIP]

```shell
git commit --amend --signoff
git push --force
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,7 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
 - Fix a TypeError crash in the Dashboard UI when opening the details page of a Workflow-type schedule [#5034](https://github.com/chaos-mesh/chaos-mesh/pull/5034)
 - Fixed jvm latency experiment not working for jdk version 19 and higher [#4821](https://github.com/chaos-mesh/chaos-mesh/pull/4821)
 - Fix dashboard UI token validation and infinite auth lockout loop [#5046](https://github.com/chaos-mesh/chaos-mesh/pull/5046)
+- Fix convert netem delay/jitter as durations instead of strings [#5074](https://github.com/chaos-mesh/chaos-mesh/pull/5074)
 
 ### Security
 
--- a/pkg/chaosdaemon/tc_server.go
+++ b/pkg/chaosdaemon/tc_server.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"net"
 	"strings"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/golang/protobuf/ptypes/empty"
@@ -392,7 +393,12 @@ func (c *tcClient) addPrio(device string, parent int, band int) error {
 func (c *tcClient) addNetem(device string, parent string, handle string, netem *pb.Netem) error {
 	c.log.Info("adding netem", "device", device, "parent", parent, "handle", handle)
 
-	args := fmt.Sprintf("qdisc add dev %s %s %s netem %s", device, parent, handle, convertNetemToArgs(netem))
+	netemArgs, err := convertNetemToArgs(netem)
+	if err != nil {
+		return err
+	}
+
+	args := fmt.Sprintf("qdisc add dev %s %s %s netem %s", device, parent, handle, netemArgs)
 	processBuilder := bpm.DefaultProcessBuilder("tc", strings.Split(args, " ")...).SetContext(c.ctx)
 	if c.enterNS {
 		processBuilder = processBuilder.SetNS(c.pid, bpm.NetNS)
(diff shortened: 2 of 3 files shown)


STRRL (approved):
Thanks for your contribution! @0xff-dev

PR #5090, opened 2026-09-03
fix(networkchaos): isolate generated rule names

bvolpato:

NetworkChaos resources with the same name in different Kubernetes namespaces can generate identical IP-set and iptables-chain names when they select the same workload pod. PodNetworkChaos tracks ownership by namespace/name, but the daemon receives colliding names, allowing one experiment to overwrite another experiment's destination sets or partition rules.

Generate kernel rule names from the full namespace/name identity and IP-set role. Use the available name budget for the hash so long role suffixes do not weaken collision resistance. Preserve the existing 27-character limits, the daemon's temporary IP-set suffix space, and INPUT/OUTPUT chain prefixes.

Names follow namespace/name ownership, matching RawRuleSource and existing recovery transactions. Persisted PodNetworkChaos names are forwarded unchanged; this does not migrate active rules or change the CRD format.

New regressions failed on the base revision and pass with the fix. They apply same-name experiments from different namespaces through the real partition and filtered-delay implementations, verify separate destination sets and daemon requests, retry application, and recover one experiment while preserving the other. They also exercise mixed persisted legacy/new names and name-length boundaries.

Passed on Linux arm64 with Go 1.25.12:

```sh
go run github.com/pingcap/failpoint/failpoint-ctl enable ./pkg/mock
go test -mod=readonly -race ./controllers/podnetworkchaos/... ./controllers/chaosimpl/networkchaos/... -count=1
go run github.com/pingcap/failpoint/failpoint-ctl disable ./pkg/mock
go vet ./controllers/podnetworkchaos/... ./controllers/chaosimpl/networkchaos/...
```

Focused goimports and git diff --check passed. Tests use fake Kubernetes clients and record daemon requests; no live network changes or Kubernetes experiments were run.

## Downsides

Generated names no longer contain a readable experiment-name fragment. Use the persisted PodNetworkChaos Source fields to map kernel names to experiments. The fixed kernel name limits require truncated hashes (108 bits for IP sets, 84/80 bits for input/output chain suffixes).

## Risk and rollback

This changes newly generated NetworkChaos rule names. Existing stored names remain readable by both old and new controllers/daemons, and recovery continues to clear rules by Source. Already-colliding active experiments must be recovered and reapplied to receive distinct names; this change does not repair their stored names automatically.

Watch for unexpected shared names, missing destination filters, or failed recovery events. Revert and redeploy the controller to roll back; persisted rules remain consumable, but newly generated rules after downgrade regain the original cross-namespace collision risk. No CRD or data migration is required.

## Related changes

- [ ] This change requires UI or website updates

## Checklist

- [x] Updated CHANGELOG.md
- [x] Unit tests
- [ ] E2E tests

## DCO

Commits are signed off.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
 
 ### Fixed
 
+- Prevent NetworkChaos experiments with the same name in different namespaces from sharing IP sets and iptables chains on a pod. Existing stored rule names remain valid until the experiment is recovered or reapplied.
 - Fixed helm chart template include for extra objects to use the correct render function [#4780](https://github.com/chaos-mesh/chaos-mesh/pull/4780)
 - Fix `install.sh` exiting when kubectl version prints warnings to stderr [#4796](https://github.com/chaos-mesh/chaos-mesh/pull/4796)
 - Remove caBundle placeholder in webhook templates when cert-manager is enabled to fix server-side apply conflicts [#4828](https://github.com/chaos-mesh/chaos-mesh/pull/4828)
--- a/controllers/podnetworkchaos/ipset/ipset.go
+++ b/controllers/podnetworkchaos/ipset/ipset.go
@@ -95,7 +95,7 @@ func BuildSetIPSet(sets []v1alpha1.RawIPSet, networkchaos *v1alpha1.NetworkChaos
 
 // GenerateIPSetName generates name for ipset
 func GenerateIPSetName(networkchaos *v1alpha1.NetworkChaos, namePostFix string) string {
-	return netutils.CompressName(networkchaos.Name, 27, namePostFix)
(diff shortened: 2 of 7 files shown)


bvolpato:
The integration job passed the network test, then failed to create PhysicalMachineChaos after the OIDC test restarted the API server and upgraded the chart. The admission error was `x509: certificate signed by unknown authority`. The same failure appeared on #5091.

I checked the chart and logs: the upgrade regenerates webhook certificates and trust bundles, but the test only checks dashboard readiness before continuing. That points to an existing upgrade-readiness issue; the logs do not show which endpoint or certificate was stale.

Could a maintainer rerun the [failed integration job](https://github.com/chaos-mesh/chaos-mesh/actions/runs/33785249398/job/100748328298)? The other checks passed, but I couldn't start the rerun because GitHub requires repository admin rights. Thanks.

PR #5091, opened 2026-09-03
fix(httpchaos): retry missing TLS Secret data

bvolpato:

If a TLS Secret exists but lacks the configured certificate, private-key, or CA entry, PodHTTPChaos calls `errors.Wrapf` with a nil error and then panics on `err.Error()`. Its deferred status update still records the generation with an empty failure message. The parent can consequently report `Injected` without any `ApplyHttpChaos` call, and later child reconciliation skips that generation.

Create a non-nil error for each missing entry and preserve it in `FailedMessage`. Request a rate-limited retry, following the controller's existing error convention, because Secret updates are not watched. The parent stays pending with an error until a subsequent attempt actually succeeds.

Regression tests cover missing certificate, private key, and optional CA data; repeated failed reconciliation; zero injection RPCs during failure; repairing only the Secret; the resulting TLS payload and parent `Injected` transition; and successful reconciliation idempotence.

Related: #4994 fixes Secret lookup and read errors. This PR covers the separate missing-data branches and can be applied independently.

The regression test fails on the base commit with a nil-pointer panic in all three cases, and passes with this change. It uses fake Kubernetes and daemon clients.

```sh
go run github.com/pingcap/failpoint/failpoint-ctl enable ./pkg/mock
trap 'go run github.com/pingcap/failpoint/failpoint-ctl disable ./pkg/mock' EXIT
go test -mod=readonly ./controllers/podhttpchaos ./controllers/chaosimpl/httpchaos/... -count=1
go vet ./controllers/podhttpchaos ./controllers/chaosimpl/httpchaos/...
```

Focused tests and vet passed with Go 1.25.12. Files were formatted with goimports. Revive completed with warnings for existing package/exported comments and the mock's required `ApplyHttpChaos` RPC spelling. Full containerized checks and live Kubernetes e2e tests were not run.

## Downsides

Missing Secret entries now cause rate-limited retries and warning events until repaired. Recovery also continues to report failure while required TLS data is missing.

## Risk and rollback

The change is limited to missing TLS data in the controller; no API or daemon changes are required. Previously misreported experiments still need reapplication after repairing their Secret so that the child gets a fresh generation. Revert this commit and redeploy the controller to roll back; this restores the previous false-success behavior.

## Related changes

- [ ] This change also requires further updates to the website

### CHANGELOG

### Tests

- [x] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

The commit is signed off.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
 
 ### Fixed
 
+- Prevent HTTPChaos from reporting successful injection when the TLS Secret is missing a configured certificate, private key, or CA entry
 - Fixed helm chart template include for extra objects to use the correct render function [#4780](https://github.com/chaos-mesh/chaos-mesh/pull/4780)
 - Fix `install.sh` exiting when kubectl version prints warnings to stderr [#4796](https://github.com/chaos-mesh/chaos-mesh/pull/4796)
 - Remove caBundle placeholder in webhook templates when cert-manager is enabled to fix server-side apply conflicts [#4828](https://github.com/chaos-mesh/chaos-mesh/pull/4828)
--- a/controllers/podhttpchaos/controller.go
+++ b/controllers/podhttpchaos/controller.go
@@ -167,25 +167,25 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 		cert, ok := secret.Data[tlsKeys.CertName]
 		if !ok {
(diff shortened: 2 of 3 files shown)


bvolpato:
The integration failure here matches #5090: PhysicalMachineChaos admission rejects the webhook certificate just after the OIDC test upgrades the chart. It fails before physical-machine injection, outside the HTTPChaos code changed here.

The Go verification, build, and test jobs passed on both architectures. Could a maintainer rerun the [failed integration job](https://github.com/chaos-mesh/chaos-mesh/actions/runs/33798974196/job/100793434730)? The personal fork account cannot rerun this workflow. Thanks.

Issue #3493, opened 2022-07-27
Create NetworkChaos with same selector in . and target will cause error


YangKeao:
## Bug Report

```
spec:
  selector:
    namespaces:
    .....
  direction: both
  target:
    selector:
      ....
```

YangKeao:
I'll enrich the description of this issue later :cry: .

PR #5092, opened 2026-09-03
fix(networkchaos): preserve overlapping selector roles

bvolpato:

A bidirectional NetworkChaos experiment can lose one requested direction when a pod matches both selectors. For example, source pods `{A, O}` and target pods `{O, B}` produce two records for `O`. Each record clears and rebuilds the same experiment's per-pod rules, so the second record removes the first role's filter. All records can then report `Injected` while either `O -> A` or `O -> B` has no requested delay.

Related: #3493 reports problems with overlapping source and target selectors.

For `direction: both`, either record for an overlapping pod rebuilds both roles in one transaction before committing its generation. Keep the existing source/target IP-set names, filters, interface selection, and experiment ownership. Retries reproduce the same complete configuration, and recovery continues to clear both roles using the existing experiment owner.

Regression tests cover both processing orders, matching and differing interface settings, non-overlapping `to`/`from`/`both` experiments, repeated application, acknowledgement, and recovery while another experiment remains on the pod.

Resolve the destination Pod IP before starting the timed network integration test. Keep the five-second duration, ten packets, and requirement for three to six packets with at least 10ms latency, while moving hostname resolution out of the measurement window.

This change is independent of the namespace identity fix in #5090. Experiments using different interfaces also need the separate classifier-chain fix in #5089; these tests verify the intended per-interface controller configuration.

```sh
go test -mod=readonly ./controllers/chaosimpl/networkchaos/... -count=1
go vet ./controllers/chaosimpl/networkchaos/...
```

All four overlapping-selector cases fail against the base implementation and pass with this change. The NetworkChaos package tests and vet passed with Go 1.25.12. Tests use fake Kubernetes clients and model API-server generations and daemon acknowledgements.

A separate probe passed both processing orders through actual PodNetworkChaos IP-set, iptables, and traffic-control RPC conversion and netem merging, verifying both filter graphs and the requested 200ms delay with a fake daemon. Files were formatted with goimports; revive reports only existing comment warnings. The integration shell script passes `bash -n`. Full containerized checks and live Kubernetes e2e tests were not run locally.

## Downsides

An overlapping pod now retains both requested role filters instead of just the last one. This adds one traffic-control entry and three IP-set entries for that pod. Each overlapping record rebuilds both roles, with additional target-pod reads during initial application or retries.

## Risk and rollback

The change affects NetworkChaos traffic-control actions with overlapping selectors and `direction: both`; partition handling and public APIs are unchanged. The intended effect is that previously missed destinations now receive the configured fault. Reapply existing experiments to rebuild their child configuration. Revert this commit and redeploy the controller to roll back, then reapply experiments if the previous rule layout is needed; a controller rollback alone does not rewrite already acknowledged child resources.

## Related changes

- [ ] This change also requires further updates to the website

### CHANGELOG

### Tests

- [x] Unit test
- [ ] E2E test
- [ ] Manual test

### Side effects

## DCO

The commit is signed off.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@ For more information and how-to, see [RFC: Keep A Changelog](https://github.com/
 
 ### Fixed
 
+- Preserve both traffic directions when a NetworkChaos pod matches overlapping source and target selectors with `direction: both`.
 - Fixed helm chart template include for extra objects to use the correct render function [#4780](https://github.com/chaos-mesh/chaos-mesh/pull/4780)
 - Fix `install.sh` exiting when kubectl version prints warnings to stderr [#4796](https://github.com/chaos-mesh/chaos-mesh/pull/4796)
 - Remove caBundle placeholder in webhook templates when cert-manager is enabled to fix server-side apply conflicts [#4828](https://github.com/chaos-mesh/chaos-mesh/pull/4828)
--- a/controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
+++ b/controllers/chaosimpl/networkchaos/trafficcontrol/impl.go
@@ -110,66 +110,58 @@ func (impl *Impl) Apply(ctx context.Context, index int, records []*v1alpha1.Reco
 		return v1alpha1.NotInjected, err
 	}
 
+	inSource := record.SelectorKey == "."
+	inTarget := record.SelectorKey == ".Target"
+	if !inSource && !inTarget {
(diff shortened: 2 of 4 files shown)


STRRL:
Hi @bvolpato , thanks for your contribution!

I noticed you're highly active rencently. For now chaos mesh maintainers may have only limited bandwidth on reviewing them. 🙏

Does this sounds a problem? We're open to chat!

STRRL:
We're avaiable in CNCF slack channel #project-chaos-mesh, and DM is also welcome.

